PR #145, opened 2026-03-19
fix: preserve auto_link config on save, fix 9 integration test failures

lancekrogers:
## Summary

- **Production bug fix**: `AutoLinkConfig.IsZero()` implements `yaml.IsZeroer` so `omitempty` only drops the `auto_link` section when it was never explicitly present in source YAML. Previously, setting `enabled: false` caused the entire section to be silently dropped on re-save (e.g. during `fest status set active`), reverting to defaults on next load.
- **Integration test fixes**: Disable auto-link validation in test fest.yaml configs (Docker containers have no real project directories). Fix gate frontmatter to use `fest_gate_id` instead of `fest_gate_type`.

## Test plan

- [x] New unit test `TestAutoLinkConfig_RoundTripPreservesDisabled` verifies load→save→load preserves `enabled: false`
- [x] All 9 previously failing integration tests now pass
- [x] Full integration suite (44 tests) passes
- [x] All unit tests pass

### DIFF
--- a/internal/config/festival_config.go
+++ b/internal/config/festival_config.go
@@ -95,6 +95,14 @@ type AutoLinkConfig struct {
 	validatePathExistsSet bool `yaml:"-"`
 }
 
+// IsZero implements the yaml.IsZeroer interface so that omitempty only drops
+// the auto_link section when it was never explicitly present in the source YAML.
+// Without this, setting enabled: false (a zero value) causes the entire section
+// to be silently dropped on re-save, reverting to defaults on next load.
+func (c AutoLinkConfig) IsZero() bool {
+	return !c.present
+}
+
 // DefaultAutoLinkConfig returns the default auto-link configuration.
 // Auto-link is enabled by default, requiring fest_working_dir on implementation phases.
 func DefaultAutoLinkConfig() AutoLinkConfig {
--- a/internal/config/festival_config_test.go
+++ b/internal/config/festival_config_test.go
@@ -690,6 +690,37 @@ func TestLoadFestivalConfig_AutoLinkExplicitDisabledPreserved(t *testing.T) {
 	}
 }
 
+func TestAutoLinkConfig_RoundTripPreservesDisabled(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	// Write config with auto_link disabled
+	data := []byte("version: \"1.0\"\nauto_link:\n  enabled: false\n")
+	if err := os.WriteFile(filepath.Join(tmpDir, FestivalConfigFileName), data, 0644); err != nil {
+		t.Fatalf("write initial config: %v", err)
+	}
+
+	// Load → save → load round-trip (simulates fest status set active)
+	cfg, err := LoadFestivalConfig(tmpDir, "")
+	if err != nil {
+		t.Fatalf("first load: %v", err)
+	}
+	if cfg.AutoLink.Enabled {
+		t.Fatal("after first load: AutoLink.Enabled = true, want false")
+	}
+
+	if err := SaveFestivalConfig(tmpDir, "", cfg); err != nil {
+		t.Fatalf("save: %v", err)
+	}
+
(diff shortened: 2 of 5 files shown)


internal/config/festival_config.go
```diff
@@ -95,6 +95,14 @@ type AutoLinkConfig struct {
 	validatePathExistsSet bool `yaml:"-"`
 }
 
+// IsZero implements the yaml.IsZeroer interface so that omitempty only drops
+// the auto_link section when it was never explicitly present in the source YAML.
+// Without this, setting enabled: false (a zero value) causes the entire section
```
obey-agent, line 100: **Textbook `yaml.IsZeroer` implementation.** The root cause is subtle and well-documented: `enabled: false` is the Go zero value for bool, so without a custom `IsZero()`, the YAML marshaler's `omitempty` drops the entire `auto_link` section on save. Next load uses defaults (`enabled: true`), silently reverting the user's explicit config.

The fix uses the existing `present` field (set during unmarshal, tagged `yaml:"-"`) to distinguish "user never wrote auto_link" from "user wrote auto_link with zero-valued fields." This is the correct Go YAML pattern.

The godoc comment explains the WHY, not just the WHAT — a junior engineer encountering this method would understand why it exists and why removing it would be a regression.

internal/config/festival_config_test.go
```diff
@@ -690,6 +690,37 @@ func TestLoadFestivalConfig_AutoLinkExplicitDisabledPreserved(t *testing.T) {
 	}
 }
 
+func TestAutoLinkConfig_RoundTripPreservesDisabled(t *testing.T) {
```
obey-agent, line 693: **Good round-trip reproduction test.** This is the exact production scenario: load config with `enabled: false` → save (e.g. during `fest status set active`) → load again → verify `enabled` is still false. Without the `IsZero()` fix, the second load returns `enabled: true` because the section was dropped on save.

The test name `RoundTripPreservesDisabled` clearly communicates the invariant being tested.

obey-agent (commented):
## Staff Engineer Review: `fix: preserve auto_link config on save, fix 9 integration test failures`

### Verdict: **Approve** — clean bugfix with correct root cause analysis and thorough testing.

---

### The Bug

`AutoLinkConfig` with `enabled: false` has all exported fields at Go zero values. YAML's `omitempty` calls `IsZero()` — default Go behavior checks if all fields are zero → returns true → drops the entire `auto_link` section on save. Next load applies defaults (`enabled: true`), silently reverting the user's explicit `enabled: false`.

This is a production data loss bug triggered by any operation that round-trips the config (e.g., `fest status set active`).

### The Fix

`AutoLinkConfig.IsZero()` implements `yaml.IsZeroer`, returning `!c.present`. The `present` field is set during unmarshal (tagged `yaml:"-"`) — it tracks whether the user's source YAML included `auto_link:` at all. If present, `IsZero()` returns false regardless of field values, and `omitempty` keeps the section.

### What's Done Well

1. **Root cause is correctly identified and surgically fixed.** One 8-line method on the right struct. No workarounds, no hacks.

2. **Round-trip test reproduces the exact production scenario.** Load → save → load verifies `enabled: false` survives. The test name communicates the invariant.

3. **Integration test fixes are correct.** Docker containers have no real campaign project directories, so `auto_link: enabled: false` disables validation that requires `fest_working_dir` paths to exist. Gate frontmatter updated to use `fest_gate_id` (follow-through from the gate identity rework in #144).

4. **`disableAutoLinkInFestivalConfig` is a proper helper.** Parses YAML, sets the field, writes back. Used across multiple test files consistently.

### Staff Standard

Would I stake my career on this? Yes. This is exactly the kind of bug a staff engineer finds and fixes cleanly: subtle serialization behavior, correct use of a Go interface contract (`yaml.IsZeroer`), reproduction test, and no collateral damage. +84/-12 across 5 files. Perfectly scoped.

PR #151, opened 2026-04-11
fix: fgo + fest list support dungeon date buckets

lancekrogers:
## Summary

- **`fgo <festival-name>`** now resolves festivals nested inside dungeon date buckets (`dungeon/<substatus>/YYYY-MM-DD/<name>/`). Previously `resolveFestivalByName` only stat'd the direct child of each status dir, making every dungeon festival invisible to bare-name navigation. Fuzzy search continues to exclude the dungeon by design.
- **`fest list completed|someday|archived`** now shows the `moved YYYY-MM-DD` bucket date inline and orders entries newest-bucket-first. The bucket name is carried on a new `FestivalInfo.StatusDate` field populated in every discovery path (`ListFestivalsByStatus`, `ListFestivalsByStatusLight`, `FindFestivalByName`, `parseFestivalInfo`). Non-dungeon listings are untouched.
- **`fest list --json`** now exposes `status_date` on dungeon entries (`omitempty` elsewhere).
- **`sortByStatusDate`** is the new default order for dungeon listings. Explicit `--sort` / `--alpha` still takes precedence.
- `looksLikeDateDir` is exported as `show.LooksLikeDateDir` so the navigation package can reuse one regex source of truth.

Design package: `workflow/design/fest-bug-4-11-2026/` in obey-campaign.

## Before / After

**Before** (dungeon festivals invisible to `fgo`, no bucket date shown):
\`\`\`
$ fgo feb-9th-gathered-improvements-FG0001
error: target not found

$ fest list completed | head -3
DUNGEON/COMPLETED Festivals (90)
────────────────────────────────────────
  feb-9th-gathered-improvements-FG0001
\`\`\`

**After**:
\`\`\`
$ fgo feb-9th-gathered-improvements-FG0001 --print
/Users/.../festivals/dungeon/completed/2026-02-10/feb-9th-gathered-improvements-FG0001

$ fest list completed | head -3
DUNGEON/COMPLETED Festivals (90)
────────────────────────────────────────
  camp-workitems-manual-priority-CW0002  moved 2026-04-07 [100%]
\`\`\`

## Test plan

- [x] \`just test all\` — 1993 unit + 46 integration tests pass
- [x] \`golangci-lint run ./internal/commands/{show,navigation,list}/...\` — 0 issues
- [x] Manual E2E against the obey-campaign workspace dungeon (90 completed + 6 someday + 9 archived entries)
- [x] \`fest list completed\` shows \`moved YYYY-MM-DD\` and sorts newest bucket first
- [x] \`fest go <dungeoned-name> --print\` resolves the nested path
- [x] \`fest list --json completed\` exposes \`status_date\` on every entry
- [x] \`fest list completed --sort name\` still overrides (explicit flag wins)
- [x] \`fest list active\` / \`fest list\` / \`fest go 002\` — byte-identical to pre-fix behavior
- [x] Fuzzy partial matches on dungeon names still fail (deliberate — preserves original design)

## Acceptance

- Additive data-model change (new \`StatusDate string\` field, \`omitempty\`).
- No on-disk layout change, no migrations.
- Only net-new code path: `resolveFestivalByName` gains a date-bucket descent for dungeon statuses.
- Rendering changes are gated on `strings.HasPrefix(status, "dungeon/")` so non-dungeon output is untouched.

### DIFF
--- a/internal/commands/navigation/go.go
+++ b/internal/commands/navigation/go.go
@@ -359,6 +359,8 @@ func resolveFestivalByName(name, festivalsDir string) string {
 		}
 
 		// Dungeon substatuses: descend one level into date buckets.
+		// Pick the newest bucket when the same name exists in multiple
+		// buckets, matching the "newest first" semantics in sortByStatusDate.
 		if !strings.HasPrefix(status, "dungeon/") {
 			continue
 		}
@@ -366,14 +368,20 @@ func resolveFestivalByName(name, festivalsDir string) string {
 		if err != nil {
 			continue
 		}
+		var best, bestBucket string
 		for _, entry := range entries {
 			if !entry.IsDir() || !show.LooksLikeDateDir(entry.Name()) {
 				continue
 			}
 			datedPath := filepath.Join(statusDir, entry.Name(), name)
 			if info, err := os.Stat(datedPath); err == nil && info.IsDir() {
-				return datedPath
+				if entry.Name() > bestBucket {
+					best, bestBucket = datedPath, entry.Name()
+				}
 			}
+		}
+		if best != "" {
+			return best
 		}
 	}
 	return ""
--- a/internal/commands/navigation/go.go
+++ b/internal/commands/navigation/go.go
@@ -342,12 +342,38 @@ func resolveFuzzy(pattern, festivalsDir string) (string, error) {
 	return result, err
 }
 
-// resolveFestivalByName searches for a festival by name in status directories
+// resolveFestivalByName searches for a festival by name in status directories.
+// For dungeon substatuses it also descends one level into YYYY-MM-DD date
+// buckets, matching the on-disk layout used when festivals are moved to the
+// dungeon. Fuzzy search intentionally excludes the dungeon (see
+// internal/navigation/fuzzy.go); exact-name lookup does not.
 func resolveFestivalByName(name, festivalsDir string) string {
 	for _, status := range id.StatusDirectories {
-		festPath := filepath.Join(festivalsDir, status, name)
+		statusDir := filepath.Join(festivalsDir, status)
+
+		// Direct child: active/ready/planning/ritual, or any flat dungeon
+		// entry that was moved without a date bucket.
+		festPath := filepath.Join(statusDir, name)
 		if info, err := os.Stat(festPath); err == nil && info.IsDir() {
 			return festPath
+		}
+
+		// Dungeon substatuses: descend one level into date buckets.
+		if !strings.HasPrefix(status, "dungeon/") {
+			continue
+		}
+		entries, err := os.ReadDir(statusDir)
+		if err != nil {
+			continue
+		}
+		for _, entry := range entries {
+			if !entry.IsDir() || !show.LooksLikeDateDir(entry.Name()) {
+				continue
+			}
+			datedPath := filepath.Join(statusDir, entry.Name(), name)
+			if info, err := os.Stat(datedPath); err == nil && info.IsDir() {
+				return datedPath
+			}
 		}
 	}
 	return ""
--- a/internal/commands/show/stats.go
+++ b/internal/commands/show/stats.go
@@ -21,10 +21,15 @@ type FestivalInfo struct {
 	Name         string    `json:"name"`                    // Directory name (used for linking)
 	MetadataName string    `json:"metadata_name,omitempty"` // Name from fest.yaml metadata (clean name without ID)
 	Status       string    `json:"status"`
-	Priority     string    `json:"priority,omitempty"`
-	Path         string    `json:"path"`
-	ProjectPath  string    `json:"project_path,omitempty"` // Linked project directory from fest.yaml
-	ModTime      time.Time `json:"mod_time"`               // Directory modification time
+	// StatusDate is the dated bucket a festival lives under when it is in a
+	// dungeon substatus. Format is the raw on-disk directory name
+	// (YYYY-MM-DD for new entries, or YYYY-MM for legacy entries).
+	// Empty for festivals that are not stored inside a dated bucket.
+	StatusDate  string    `json:"status_date,omitempty"`
+	Priority    string    `json:"priority,omitempty"`
+	Path        string    `json:"path"`
+	ProjectPath string    `json:"project_path,omitempty"` // Linked project directory from fest.yaml
+	ModTime     time.Time `json:"mod_time"`               // Directory modification time
 	// CreatedAt is when the festival was first created (from fest_created frontmatter or directory mod time).
 	CreatedAt time.Time `json:"created_at,omitempty"`
 	// UpdatedAt is the last modification time (from fest_updated frontmatter or directory mod time).
--- a/internal/commands/show/stats.go
+++ b/internal/commands/show/stats.go
@@ -25,6 +25,11 @@ type FestivalInfo struct {
 	// dungeon substatus. Format is the raw on-disk directory name
 	// (YYYY-MM-DD for new entries, or YYYY-MM for legacy entries).
 	// Empty for festivals that are not stored inside a dated bucket.
+	//
+	// Canonical population: parseFestivalInfo detects the date-dir parent
+	// via LooksLikeDateDir and sets StatusDate automatically.
+	// ListFestivalsByStatusLight uses lightFestivalInfo (no parseFestivalInfo),
+	// so it sets StatusDate manually in its date-dir recursion block.
 	StatusDate  string    `json:"status_date,omitempty"`
 	Priority    string    `json:"priority,omitempty"`
 	Path        string    `json:"path"`
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.25.6
 
 require (
 	github.com/Masterminds/sprig/v3 v3.3.0
-	github.com/Obedience-Corp/camp v0.2.0-dev.1
+	github.com/Obedience-Corp/camp v0.2.2
 	github.com/Obedience-Corp/obey-shared v0.2.0
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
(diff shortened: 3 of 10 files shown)


internal/commands/navigation/go.go
```diff
@@ -342,13 +342,39 @@ func resolveFuzzy(pattern, festivalsDir string) (string, error) {
...
+			continue
+		}
+		for _, entry := range entries {
+			if !entry.IsDir() || !show.LooksLikeDateDir(entry.Name()) {
+				continue
+			}
+			datedPath := filepath.Join(statusDir, entry.Name(), name)
+			if info, err := os.Stat(datedPath); err == nil && info.IsDir() {
+				return datedPath
+			}
```
obey-agent, line 381: **First-match-wins silently hides duplicate names across dungeon buckets**

```go
for _, entry := range entries {
    if !entry.IsDir() || !show.LooksLikeDateDir(entry.Name()) {
        continue
    }
    datedPath := filepath.Join(statusDir, entry.Name(), name)
    if info, err := os.Stat(datedPath); err == nil && info.IsDir() {
        return datedPath
    }
}
```

`os.ReadDir` returns entries in lexical order on macOS/Linux, so a name that happens to exist in multiple date buckets will return the **oldest** bucket first and silently ignore the newer one. That's backwards from the semantics the rest of the PR establishes — `fest list` orders newest-bucket-first with `sortByStatusDate`, but `fgo <name>` jumps to the oldest match.

This is unlikely in practice (most festival names carry an ID suffix like `-FG0001`), but it's not impossible — a user who renamed or cloned a festival could end up with collisions, and the silent-pick behavior means they'd land on the wrong directory with no warning.

Two options, pick whichever fits:

1. **Match "newest bucket wins"** — collect all matches, return the one with the highest bucket name (or the most recent `ModTime` if you want to be paranoid about YYYY-MM vs YYYY-MM-DD mixed format):

```go
var best, bestBucket string
for _, entry := range entries {
    if !entry.IsDir() || !show.LooksLikeDateDir(entry.Name()) {
        continue
    }
    datedPath := filepath.Join(statusDir, entry.Name(), name)
    if info, err := os.Stat(datedPath); err == nil && info.IsDir() {
        if entry.Name() > bestBucket {
            best, bestBucket = datedPath, entry.Name()
        }
    }
}
if best != "" {
    return best
}
```

2. **Error on ambiguous match** — if more than one bucket has a dir with this name, return an error like `multiple festivals named %q in dungeon/%s (buckets: ...)` and let the user disambiguate by ID. More defensive, less ergonomic.

I'd lean toward option 1 since the "newest wins" rule matches the rest of the PR's design, and ambiguous names are rare enough that "just pick the recent one" is the correct default.

internal/commands/show/stats.go
```diff
@@ -21,10 +21,15 @@ type FestivalInfo struct {
...
 	Status       string    `json:"status"`
-	Priority     string    `json:"priority,omitempty"`
-	Path         string    `json:"path"`
-	ProjectPath  string    `json:"project_path,omitempty"` // Linked project directory from fest.yaml
-	ModTime      time.Time `json:"mod_time"`               // Directory modification time
+	// StatusDate is the dated bucket a festival lives under when it is in a
+	// dungeon substatus. Format is the raw on-disk directory name
+	// (YYYY-MM-DD for new entries, or YYYY-MM for legacy entries).
+	// Empty for festivals that are not stored inside a dated bucket.
+	StatusDate  string    `json:"status_date,omitempty"`
```
obey-agent, line 33: **`StatusDate` is populated manually at three separate call sites — centralize to avoid future drift**

`parseFestivalInfo` already has the right parsing logic: it walks up from a festival path, detects the date bucket via `LooksLikeDateDir`, confirms the grandparent is `dungeon`, and sets `info.Status = "dungeon/<name>"` + `info.StatusDate = bucket`. That's the canonical detection path.

But `ListFestivalsByStatus`, `ListFestivalsByStatusLight`, and `FindFestivalByName` each **also** manually set `info.StatusDate = bucket` after their own per-path traversal. Three copies of the same invariant, each of which has to remember to populate the field when a new discovery path lands.

`ListFestivalsByStatusLight` is the most fragile one — it uses `lightFestivalInfo` (which doesn't go through `parseFestivalInfo`), so it *has* to do manual population. The other two could potentially delegate to `parseFestivalInfo` since they already have the full festival dir on disk, but there's a performance reason to keep `Light` light.

Two ways to keep this tidy:

1. **Single helper** — extract a `resolveDungeonBucket(festivalDir string) (status, statusDate string)` that reads the path structure once and is called from every info-construction site. Each call site becomes a one-liner.
2. **Accept as-is** — the current pattern works. Just add a code comment near `FestivalInfo.StatusDate` that says "Populate this at every dungeon-aware discovery site; see ListFestivalsByStatus/Light/FindFestivalByName/parseFestivalInfo for the pattern" so the next contributor knows where to look.

Option 1 is the right long-term fix; option 2 is a 5-minute mitigation if you'd rather ship this PR without restructuring. Neither is blocking — the logic as written is correct at this revision — but the current setup is exactly how "we added a dungeon status and StatusDate started returning empty for the new status" bugs get introduced six months from now.

obey-agent (commented):
## Staff-Level Review — Approve (with two non-blocking notes)

**Verdict**: Approve

**Overview**: This is a well-scoped bug fix. The problem is real — `fgo <name>` was silently failing on any dungeon festival because `resolveFestivalByName` only stat'd direct children of the status directory — and the solution is appropriately minimal: descend one level into date buckets when `strings.HasPrefix(status, "dungeon/")`. The companion `fest list` enhancements (bucket date display, newest-first sort, JSON field) follow the same additive shape, and the test coverage across `list_test.go`, `go_test.go`, and `show_listing_test.go` (+282 lines total) exercises all the new paths. The PR description's before/after example and the explicit "fuzzy search continues to exclude the dungeon by design" note both show careful thought about what *not* to change.

### Key Findings

1. **Non-unique bare-name lookup silently picks the oldest bucket.** `os.ReadDir` returns entries in lexical order on macOS/Linux, and `resolveFestivalByName` returns the first matching `stat`. If the same festival name exists in multiple dungeon buckets (e.g., after a rename or clone), `fgo` jumps to the oldest one — which contradicts the rest of the PR's "newest bucket wins" posture in `sortByStatusDate`. See inline on `internal/commands/navigation/go.go:376`.

2. **`StatusDate` is populated manually at three separate discovery call sites.** `parseFestivalInfo` already has canonical parsing logic for "dungeon + date bucket", but `ListFestivalsByStatus`, `ListFestivalsByStatusLight`, and `FindFestivalByName` each re-implement the bucket-assignment inline. This works today but is exactly how "new discovery path forgets to populate StatusDate" bugs creep in. Options: extract a `resolveDungeonBucket(festivalDir)` helper, or at minimum add a code-comment pointer near the field definition. See inline on `internal/commands/show/stats.go:28`.

### Secondary Observations (no inline — acknowledge or defer)

- **`sortByStatusDate` uses `sort.Slice`, not `sort.SliceStable`.** Same-bucket-same-ModTime festivals are in non-deterministic order. Harmless for display but trivially fixable with `SliceStable` + a name tiebreaker.
- **`LooksLikeDateDir` accepts both `YYYY-MM-DD` and `YYYY-MM`.** Lexical comparison means `"2026-04" < "2026-04-01"`, so legacy YYYY-MM buckets sort *before* any day in that month. Edge case — only matters with mixed-format buckets. A date-aware parser would be cleaner, but regex + string-compare is honest about what the data shape guarantees.
- **Three dungeon substatuses × ~30 buckets each × one ReadDir + one Stat per call.** Well within acceptable CLI overhead but will get noticeable for pathological dungeons (thousands of festivals). Not a fix that belongs in this PR.

### What's Done Well

- **The PR scope is tight.** It fixes the reported bug, adds the bucket-date display that was explicitly missing, and stops there. No drive-by refactors, no speculative generalization.
- **The design package reference** (`workflow/design/fest-bug-4-11-2026/` in obey-campaign) means the fix is traceable back to a written design decision, not "I just patched this."
- **`LooksLikeDateDir` export** is the right call — a single regex source of truth for date-bucket detection across `navigation`, `show`, and `list` packages beats three copies.
- **Fuzzy exclusion of the dungeon is preserved.** The PR explicitly calls out that fuzzy search still skips the dungeon by design, and the fix targets bare-name lookup only. Correct preservation of intent.
- **Test plan is comprehensive and manually validated.** 1993 unit + 46 integration tests pass, linted clean across the three touched packages, and the author ran `fest list completed`, `fgo <dungeoned-name> --print`, and `fest list --json completed` against a real 90-completed/6-someday/9-archived dungeon. That's the kind of E2E verification that catches the things unit tests miss.
- **JSON contract extension is additive** (`omitempty`), so existing consumers of `fest list --json` don't break.
- **`sortByStatusDate`'s comment explicitly notes "ISO YYYY-MM-DD values compare lexicographically"** — the implementation's correctness assumption is documented inline, which is exactly what makes the YYYY-MM edge case above easy to spot on review.

### Staff Standard

Approve. The two inline items are follow-ups rather than blockers — the "wrong festival silently picked" scenario is unlikely in practice and the manual-populate pattern is correct at this revision — but they're both worth addressing in the next iteration or a small follow-up PR. The fix itself is solid, the tests are thorough, and the scoping is disciplined. Good work.

internal/commands/navigation/go.go
lancekrogers: Fixed in 39902b3. Went with option 1 — collect all matches, return the highest bucket name. Test added: `TestResolveFestivalByName_PicksNewestBucketOnCollision`.

Also switched `sortByStatusDate` to `sort.SliceStable` with a name tiebreaker for fully deterministic ordering (secondary observation 3a from the review).

internal/commands/show/stats.go
lancekrogers: Went with option 2 — added a doc comment on the field (39902b3) pointing at `parseFestivalInfo` as the canonical path and calling out `ListFestivalsByStatusLight` as the only non-canonical populator.

`parseFestivalInfo` already handles the detection correctly via its `LooksLikeDateDir(parentName)` switch case, so the manual re-assignments in `ListFestivalsByStatus` and `FindFestivalByName` are redundant in the success path (only load-bearing in the error-fallback branch where a minimal `FestivalInfo` is constructed). Not worth extracting a helper for three call sites today, but the comment makes the pattern visible for the next contributor.

obey-agent (commented):
## Staff-Level Re-Review — Still Approved

**Verdict**: Approve

Both inline items addressed at `39902b3`:

1. **Non-unique name lookup** → `resolveFestivalByName` now collects all date-bucket matches and returns the one with the highest lexical bucket name (`entry.Name() > bestBucket`), matching `sortByStatusDate`'s newest-first semantics. No more silent oldest-wins on `os.ReadDir` order.

2. **Non-stable sort** → `sort.Slice` → `sort.SliceStable` with a `Name` tiebreaker when `StatusDate` and `ModTime` are both equal. Fully deterministic now.

3. **StatusDate centralization** → Author chose option 2 from my suggestion: added a doc-comment on `FestivalInfo.StatusDate` that explicitly names the canonical population path (`parseFestivalInfo`) and the manual-population path (`ListFestivalsByStatusLight`). Clean and sufficient — the next contributor knows exactly where to look.

Ship it.

PR #153, opened 2026-04-16
Include ritual/ in fuzzy navigation and active-workflow status checks

lancekrogers:
## Summary

`fgo` fuzzy navigation silently skipped festivals under `festivals/ritual/`. The root cause: `PrimaryStatusDirs` in `internal/id/generator.go` was `{active, ready, planning}` — missing `ritual`. `CollectNavigationTargets` (`internal/navigation/fuzzy.go`), `fest list` defaults, `go_link`, shell completion, and phase-shortcut resolution all consume this list, so ritual festivals were invisible to fuzzy search across the board.

Ritual festivals are non-terminal workflow items (already reflected in `WorkingStatusDirectories`). They should be first-class navigation targets alongside `active/`, `ready/`, and `planning/`.

## Root cause

```go
// Before
var PrimaryStatusDirs = []string{"active", "ready", "planning"}

// After
var PrimaryStatusDirs = []string{"active", "ready", "planning", "ritual"}
```

Investigation turned up three further sites hardcoding the same stale triple that were latent bugs for ritual festivals (e.g. `show.DetectCurrentLocation` would leave `Status` empty for a ritual festival). Those are fixed in the same PR for consistency.

## Changes

| File | Change |
|---|---|
| `internal/id/generator.go` | Add `"ritual"` to `PrimaryStatusDirs`; document sync with `WorkingStatusDirectories` |
| `internal/navigation/fuzzy_test.go` | **New test** — `TestCollectNavigationTargets_IncludesRitual` |
| `internal/commands/show/detect.go` | Recognise `ritual/` parent when deriving festival status |
| `internal/commands/markers/markers.go` | Recognise `ritual/` as an active-festival location |
| `internal/commands/list/list.go` | Update help/examples to include ritual and ready in defaults (previously stale) |
| `internal/commands/config/shell.go` | Add `ritual` to bash/zsh/fish status-dir case arms so `fgo <status> <name>` combining and tab completion work for rituals |

## Test plan

- [x] **RED then GREEN** for the new test — `TestCollectNavigationTargets_IncludesRitual` fails on current `main`, passes after `PrimaryStatusDirs` change
- [x] `just test unit` — 1995/1995 tests pass
- [x] `just build quick` — clean binary
- [x] Smoke against real campaign:
  - `fest go weekly --print` (ritual festival) → resolves to `festivals/ritual/weekly-review-RI-WR0001` ✓
  - `fest go ritual --print` → resolves to `festivals/ritual` (status dir as nav target) ✓
  - `fest go corp --print` (active festival regression guard) → resolves to `festivals/active/corp-site-build-CS0003` ✓

## Notes

- Existing lint errors in `internal/commands/system/sync.go` and `tools/release-notes/internal/notes/notes.go` are pre-existing on `main` and out of scope for this PR.
- `fls` completion helpers at `shell.go:144` and `330` have pre-existing gaps (missing `ready`, `someday`, `archived`) unrelated to ritual; left for a follow-up to keep this PR focused.

### DIFF
--- a/internal/commands/config/shell.go
+++ b/internal/commands/config/shell.go
@@ -96,7 +96,7 @@ _fgo_completions() {
         completions=$(command fest go completions 2>/dev/null)
     elif [[ ${COMP_CWORD} -eq 2 ]]; then
         case "${COMP_WORDS[1]}" in
-            active|planning|ready|completed|someday|archived|dungeon)
+            active|planning|ready|ritual|completed|someday|archived|dungeon)
                 completions=$(command fest go completions --status "${COMP_WORDS[1]}" 2>/dev/null)
                 ;;
         esac
@@ -119,7 +119,7 @@ if [[ -n "$ZSH_VERSION" ]]; then
         elif (( CURRENT == 3 )); then
             # Second arg: if first arg is a status dir, show its festivals
             case "${words[2]}" in
-                active|planning|ready|completed|someday|archived|dungeon)
+                active|planning|ready|ritual|completed|someday|archived|dungeon)
                     cmd_args="--color --status ${words[2]}"
                     ;;
                 *) return ;;
@@ -237,7 +237,7 @@ fgo() {
             # Normal navigation (festival/phase/status directories)
             # Note: Don't use 2>&1 - stderr must flow to terminal for TUI picker to render
             local dest
-            if [[ -n "$2" && "$1" =~ ^(active|planning|ready|completed|someday|archived|dungeon)$ ]]; then
+            if [[ -n "$2" && "$1" =~ ^(active|planning|ready|ritual|completed|someday|archived|dungeon)$ ]]; then
                 # Status dir + festival name: combine (e.g., active my-fest → active/my-fest)
                 dest=$(command fest go "$1/$2" --print)
             else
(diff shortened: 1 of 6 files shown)


obey-agent (approved):
Verdict: Approve

Overview:
I reviewed PR #153 at head `e3d47a6268828d174c8e059bece9067cb5ff6e79`. The change is appropriately scoped: it fixes a real status-enum drift where `ritual/` had fallen out of the primary navigation surface, and it updates the adjacent call sites that were relying on the same stale assumption.

Key Findings:
- No blocking findings.

What’s done well:
- The fix is applied at the shared source of truth (`id.PrimaryStatusDirs`) instead of papering over individual callers.
- The PR also cleans up the nearby hardcoded ritual omissions in `show`, `markers`, `list`, and shell navigation helpers, which is the right staff-level move for this type of enum drift.
- The new fuzzy-navigation regression test is concrete and checks both the ritual festival entry and the `ritual/` status directory target.
- Validation is strong: I re-ran the directly affected packages and `go test ./...` passed on the PR head.

Validation:
- `go test ./internal/navigation ./internal/commands/show ./internal/commands/markers ./internal/commands/list ./internal/commands/config ./internal/commands/navigation`
- `go build ./cmd/fest`
- `go test ./...`

Staff Standard:
Yes. I would be comfortable submitting this as my own work.

PR #154, opened 2026-04-17
Fix ritual navigation and auto-commit runs

lancekrogers:
## Summary
- ignore generated `dist/` output so release-note markdown does not appear as an untracked dev artifact
- include active ritual runs in `fest go` / `fgo` target discovery and completion, and bias fuzzy matches toward active work over ritual templates
- auto-commit `fest ritual run` filesystem changes with path-scoped staging so the copied run and source ritual metadata update are persisted immediately

## Why
Two separate issues showed up while investigating ritual navigation:
1. `projects/fest` was leaving generated `dist/release-notes.md` unignored even though release tooling writes it as local output.
2. The installed `fest` binary on PATH in `JobSearch` was still `v0.2.2`, which predates the ritual fuzzy-navigation fix. Current `main` already fixes ritual-template fuzzy navigation, but active ritual runs with a `-0001` suffix still fell out of target discovery because navigation only accepted names ending directly in the base festival ID.

The current `JobSearch` workspace no longer has an active daily-job run; the last run now lives in `festivals/dungeon/completed/2026-04-16/daily-job-search-RI-DJ0001-0001`. After rebuilding from current source, `fest go daily-job` resolves the ritual template instead of erroring, and this PR ensures active runs are discoverable and preferred when they do exist.

## Validation
- `go test ./...`
- built a fresh local `fest` binary from this branch and verified `fest go daily-job` against `JobSearch`
- verified the new `ritual run` auto-commit path with a temp git repo test


obey-agent (approved):
## Staff-Level Review — Approve

**Verdict**: Approve

**Overview**: Three distinct fixes bundled cleanly: (1) `.gitignore` for generated `dist/` output, (2) ritual run navigation discovery with active-ritual preference, (3) auto-commit for `fest ritual run` filesystem changes. 7 files changed, +327/-16.

### Issue 1: `.gitignore` for `dist/`

Straightforward — generated release notes were showing as untracked. One-line fix, correct.

### Issue 2: Ritual Run Navigation

**The bug**: `CollectNavigationTargets` used `id.ExtractIDFromDirName` to filter festival directories, but active ritual runs have a `-NNNN` counter suffix appended after the festival ID (e.g., `daily-job-search-RI-DJ0001-0001`). The ID extractor doesn't accept that suffix, so active ritual runs were silently filtered out of navigation.

**The fix**:
- `isNavigableFestivalDir` now falls back to checking for festival marker files (`FESTIVAL_GOAL.md`, `FESTIVAL_OVERVIEW.md`, `fest.yaml`) when ID extraction fails.
- New `statusNavigationPriority` gives active=40, ready=20, planning=10, default=0 bonuses to fuzzy match scores.
- Active ritual runs outrank their ritual template in fuzzy matches for partial names (e.g., `daily-job` prefers the active run).

The priority system is the right design — it makes "current work" discoverable first without excluding the template entirely. The boost magnitude (40) is large enough that active runs reliably outrank templates without being so large that it breaks ranking within the same status.

### Issue 3: Ritual Run Auto-Commit

**`autoCommitRitualRun`** scopes staging to `ritualPath` and `destPath` only, so unrelated working-tree changes stay uncommitted. The test `TestRunRitual_AutoCommitsScopedFilesystemChanges` verifies exactly this — the uncommitted README change survives in the worktree while the ritual files get committed.

Commit message format:
```
chore(fest): ritual run: {name} ({ID}) -> {run-dir}
```

This follows the established "festival ops auto-commit" convention (camp `p commit`, `fest commit`). Users opting into `fest ritual run` are implicitly opting into the commit — consistent with the existing UX.

### Observations (not blocking)

1. **Test coverage is asymmetric on workspace types.** `TestRunRitual_AutoCommitsScopedFilesystemChanges` uses `WorkspaceTypeStandalone`, which exercises the non-campaign branch. The `WorkspaceTypeCampaign` branch (with `LoadCampaignID` + `PrependCampaignTag`) has no test. The existing `commitkit.PrependCampaignTag` tests probably cover the tag formatting, but an end-to-end campaign-type ritual commit test would round this out.

2. **`ShortHash` error fallback returns `"committed"` as the hash** when the short-hash lookup fails post-commit. Harmless — display code treats it as a string — but an empty string or "unknown" might be more obviously a fallback vs. a real hash.

3. **The PR description explains that `JobSearch` doesn't currently have an active daily-job run** (last run is in `dungeon/completed/`). So the fuzzy-match fix is preventative — it covers the case when an active run exists but isn't discoverable. Good to be explicit that this isn't actively broken for the user today, but will be broken the next time a ritual run lands in active.

### What's Done Well

- The fuzzy-match priority system is a clean abstraction — future status-priority tuning (e.g., deprioritizing `ritual/` templates further) is a config change, not a refactor.
- `hasFestivalMarkers` fallback handles arbitrary festival-like directories (including future ritual variants) without needing ID-parser changes.
- The auto-commit is scoped (`StageFiles(..., ritualPath, destPath)`) rather than `git add -A`, which is the right default — users' unrelated work is protected.
- Test for auto-commit verifies both positive (ritual files committed) and negative (README change NOT committed) assertions — strong coverage.
- `ritualID` extraction and inclusion in the commit message makes history queryable.

### Staff Standard

Yes. Three cleanly-scoped fixes, each with targeted tests. The auto-commit is the most behavior-changing piece but follows established convention and is scoped to avoid collateral damage. Ready to merge.

obey-agent (commented):
## Re-Review After Commit `db310ecc`

The new commit replaces the marker-file fallback approach (which I approved) with a cleaner solution: proper ID parsing via `ExtractLogicalIDFromDirName` that natively recognizes all three festival directory formats.

### Why This Is Better

My original review praised the `hasFestivalMarkers` fallback as pragmatic, but the author correctly identified it as a hack. The proper fix is to teach the ID parser the three supported naming formats:

| Format | Example | Extracted ID |
|---|---|---|
| Standard festival | `my-feature-MF0001` | `MF0001` |
| Ritual template | `weekly-review-RI-WR0001` | `RI-WR0001` |
| Ritual run | `daily-job-RI-DJ0001-0001` | `RI-DJ0001` |

Two new regex patterns in `internal/id/generator.go`:
- `ritualIDPattern`: `-RI-([A-Z]{2})(\d{4,})$`
- `ritualRunIDPattern`: `-RI-([A-Z]{2})(\d{4,})-([0-9A-Fa-f]{4})$`

The ritual-run pattern correctly accepts hex run counters (`[0-9A-Fa-f]{4}`) per the ritual naming spec.

### Bug the New Approach Catches

The test update is the key evidence that this is strictly better. The new `TestCollectNavigationTargets_IncludesActiveRitualRun` adds a `notes-scratchpad` directory with a `fest.yaml` but no valid festival ID format, then asserts it is **NOT** included in navigation targets.

With the old marker-file fallback, `notes-scratchpad` would have been picked up (fest.yaml present → considered navigable) — a false positive. With the new parser, only directories matching the three ID formats are navigable. This is the right invariant.

### Test Coverage

New `TestExtractLogicalIDFromDirName` covers:
- All three formats
- Run counter with hex digits (`-00FF`)
- Invalid cases (empty, no ID, ritual-run without RI prefix)

### What's Done Well

- Replaces a heuristic (marker files) with a spec-defined contract (naming patterns).
- Kills two helper functions (`isNavigableFestivalDir` 2-arg variant, `hasFestivalMarkers`) that were only needed by the fallback.
- Test now guards against marker-file false positives that the previous code accepted.

Prior approval stands with an upgraded implementation.

internal/commands/ritual/run_test.go
```diff
@@ -0,0 +1,155 @@
...
+		campaignID, err := commitkit.LoadCampaignID(context.Background(), tmpDir)
+		if err != nil {
+			t.Fatalf("LoadCampaignID returned error: %v", err)
+		}
+		if campaignID != "12345678-1234-1234-1234-123456789abc" {
+			t.Fatalf("LoadCampaignID = %q, want %q", campaignID, "12345678-1234-1234-1234-123456789abc")
+		}
+	}
+
+	runGit(t, tmpDir, "init")
```
obey-agent: **Blocking — filesystem-safety violation (CLAUDE.md §11).**

This test (and `TestRunRitual_AutoCommitsScopedFilesystemChanges_CampaignWorkspace` below) exercises real filesystem mutation on the host: `os.MkdirAll`, `os.WriteFile`, `git init`, `git add`, `git commit`, plus a full directory copy via the production `runRitual` code path. Campaign policy is explicit:

> Tests that mutate the filesystem must run in containerized isolation, not directly on the host via ad hoc temp directories. Do not use host `TempDir`-style tests for workflows that create, copy, move, delete, git-init, chmod, or otherwise exercise real filesystem mutation behavior.

This belongs in `tests/integration/` under the `//go:build integration` tag using the existing testcontainers harness (`NewTestContainer` in `tests/integration/helpers.go`), not as a unit test under `internal/`. Running real `git` against a host-side `t.TempDir()` can pick up ambient git config, GPG signing hooks, user-level gitignore, and environment state — exactly the developer-machine surprises the policy is designed to prevent.

Action: move both `TestRunRitual_AutoCommitsScopedFilesystemChanges*` tests to `tests/integration/ritual_run_test.go` and drive them through the mounted `fest` binary inside the Alpine container. The unit-level coverage left behind should be pure logic (message formatting, scope gating) that doesn't shell out or touch the FS.

obey-agent (commented):
## Follow-Up Review — Request Changes (blocking)

Re-reviewing after closer inspection. My prior approval was too generous — two of the "observations (not blocking)" I called out earlier were in fact blocking issues, and I missed a third, more serious one entirely. The author has already addressed the first two in follow-up commits. The third is still outstanding and is the reason I'm requesting changes.

### Blocking — Filesystem-safety policy violation

`internal/commands/ritual/run_test.go` runs real `git init` / `git add` / `git commit` and performs real directory copies against a host `t.TempDir()`. CLAUDE.md §11 is explicit that tests which mutate the filesystem (create, copy, move, delete, git-init, chmod) must run in containerized isolation via the `tests/integration/` harness, not as unit tests under `internal/`.

This isn't a pedantic rule — it's designed to prevent exactly the kind of developer-machine fallout where ambient git config, GPG signing hooks, commit templates, or `user.signingkey` requirements make the test behave differently across machines and CI environments. Both `TestRunRitual_AutoCommitsScopedFilesystemChanges` and `TestRunRitual_AutoCommitsScopedFilesystemChanges_CampaignWorkspace` need to move to `tests/integration/` under `//go:build integration`, driving the mounted `fest` binary inside Alpine (see `tests/integration/helpers.go#NewTestContainer`). See inline comment for detail.

### Prior "observations" that should have been marked blocking

- **Asymmetric workspace coverage.** The `WorkspaceTypeCampaign` branch in `autoCommitRitualRun` calls `LoadCampaignID` + `PrependCampaignTag` — exactly the path every campaign user hits. Shipping a commit-generating branch with zero test coverage isn't an "observation." Resolved by `76857bf` (`TestRunRitual_AutoCommitsScopedFilesystemChanges_CampaignWorkspace`). Correct fix, wrong initial severity on my end.
- **`ShortHash` returning `"committed"` on error.** Returning a non-hash literal where callers expect a hash meant JSON consumers and display code couldn't distinguish a real commit from a failed lookup. Resolved by `76857bf` (now returns `""`). Again, should have been blocking, not informational.

### Pre-existing (not this PR, tracked separately)

Quick scan shows the following unit tests predate #154 but shell out to real `git`, which is the same class of policy violation: `internal/buildutil/tasks/build_test.go`, `internal/commands/status/handlers_festival_test.go`, `internal/commands/navigation/go_link_test.go`. Worth a dedicated cleanup pass; not a gate on this PR.

### What's Still Good

The structural fixes remain solid: `.gitignore` for `dist/`, the `ExtractLogicalIDFromDirName` parser upgrade (cleaner than my originally-praised marker-file fallback), and the scoped `StageFiles(..., ritualPath, destPath)` auto-commit that protects unrelated worktree changes. The test matrix after `76857bf` is the right shape — it just needs to run in the right place.

### Staff Standard

Not yet. The implementation is good; the test placement isn't. Move the two ritual-run auto-commit tests into the container harness and this is ready.

obey-agent (approved):
## Re-Review Catching Up on Two Commits — Approve

**Verdict**: Approve

I missed commit `76857bf6` in my earlier re-review (it was between `db310ecc` which I did cover and my review post). This catches both `76857bf6` and the new `146a479b`.

### Commit `76857bf6` — "Harden ritual run auto-commit follow-up"

This directly addresses the two observations I raised in my initial review:

**Observation #1 closed**: Added `TestRunRitual_AutoCommitsScopedFilesystemChanges_CampaignWorkspace` — exercises the `WorkspaceTypeCampaign` branch that calls `LoadCampaignID` + `PrependCampaignTag`. The original test only covered `WorkspaceTypeStandalone`. Now both branches have coverage. The test uses `writeCampaignConfig` to drop a minimal `.campaign/campaign.yaml` with a known UUID, then asserts the commit subject starts with `[OBEY-CAMPAIGN-12345678]`.

**Observation #2 closed**: `ShortHash` fallback changed from `return "committed", nil` to `return "", nil`. The display code has `if commitHash != ""` guards, so empty string now cleanly means "no hash to show" — much clearer than the `"committed"` magic string which was confusingly displayable.

Both changes address the observations exactly as suggested. Clean follow-up.

### Commit `146a479b` — "Move ritual run auto-commit tests to container harness"

This is a convention-compliance move: unit tests were doing `git init`, `git commit`, and filesystem mutation on `t.TempDir()`, which CLAUDE.md §11 forbids because ambient git config, GPG signing, and commit templates cause divergent behavior across developer machines and CI. Same pattern as camp #232.

**Migration shape**:
- Deletes `internal/commands/ritual/run_test.go` (155 lines)
- Adds `tests/integration/ritual_run_test.go` (226 lines, `//go:build integration`)
- Preserves both table-driven cases (Standalone + Campaign) via subtests

**Isolation verified**: I traced the `GetSharedContainer(t)` pattern in `helpers.go:661`. It calls `Reset()` on every invocation, and `Reset()` runs `rm -rf /test /output /festivals /workspace ...` to clear state. Since the new test calls `GetSharedContainer(t)` inside each subtest's `t.Run(...)` closure, each case starts from a clean container state. No cross-subtest contamination.

**Helpers added**:
- `ensureGit` — idempotently installs git into the alpine container via `apk add --no-cache git` (with `command -v git` guard)
- `setupRitualWorkspace` — creates `festivals/` tree with ritual template
- `writeCampaignConfig` — drops minimal campaign.yaml for the campaign-type case
- `setupGitRepo` — init + config + seed commit (now includes `commit.gpgsign=false` to avoid ambient GPG drift — this is exactly the concern the migration was meant to address)
- `modifyReadmeUnstaged` — unrelated worktree change that must survive
- `runGit` — demuxes Docker exec via `stdcopy.StdCopy`
- `writeContainerFile` — heredoc-safe file write with shell-quoted path
- `parentDir`, `shellQuote` — utility helpers

**Verification**:
- `go build ./...` — clean
- `go vet ./...` — clean
- `go vet -tags integration ./tests/integration/` — clean
- `go test ./internal/commands/ritual/...` — "[no test files]" — confirms unit tests removed
- `go test ./internal/id/... ./internal/navigation/...` — pass (unrelated packages still green)
- Integration tests compile (runtime failure in my sandbox is just the missing pre-built Linux fest binary, not a code issue)

### Observations (not blocking)

1. **`apk add git` requires network access in the container.** If the Docker network is slow or Alpine repositories are unavailable, `ensureGit` will fail. For CI robustness, consider either baking git into a pre-built test image or handling the install failure more clearly with a test-skip. Not blocking since the existing setup also uses `apk add` at container build time.

2. **`setupGitRepo` explicitly sets `commit.gpgsign=false`**. This is exactly the ambient-config concern the migration was intended to eliminate, and the author correctly guarded against it. Good hardening — it's the sort of thing a reviewer might have had to call out, nice to see it preemptively addressed.

3. **`writeContainerFile` heredoc marker `FEST_WRITE_EOF`** — content containing that literal on its own line would break the heredoc. Not realistic for this test data, and the marker is distinctive enough to be practically collision-free.

### What's Done Well

- Two distinct improvements bundled correctly: observation responses in `76857bf6`, convention-compliance move in `146a479b`.
- The convention-compliance move is the right time for the ambient-git-config hardening (`commit.gpgsign=false`).
- Table-driven subtests with shared helpers — clean test structure.
- Reset semantics verified — each subtest gets a clean container.
- `stdcopy` demuxing for git output capture — correct handling of Docker's interleaved stream.
- Helpers are narrow and named for their role (`ensureGit`, `modifyReadmeUnstaged`, etc.).

### Staff Standard

Yes. Both observations closed, tests migrated cleanly, ambient-config hardening preemptively applied. Ready to merge.

PR #160, opened 2026-04-20
Add build profile command surface listing

lancekrogers:
## What changed

This adds a `just build profile-commands` recipe to make the stable and dev command surfaces easy to inspect and compare.

It is backed by a hidden `fest __commands` helper that prints the visible Cobra command paths for the current build profile, so the output comes from the real registered command tree instead of a hand-maintained list.

## Why

`fest` has distinct stable and dev build profiles, and we needed an easy way to verify exactly which commands are present in each profile.

## Impact

Users can now run:

- `just build profile-commands`
- `just build profile-commands stable`
- `just build profile-commands dev`

The default output prints both profiles and the dev-only delta.

## Validation

- `go test ./internal/commands ./internal/commands/release`
- `go test -tags dev ./internal/commands ./internal/commands/release`
- `just build profile-commands dev`

### DIFF
--- a/.justfiles/build.just
+++ b/.justfiles/build.just
@@ -60,6 +60,50 @@ quick-stable:
 quick-dev:
     BUILD_TAGS=dev just build quick
 
+# Show the command surface for stable/dev build profiles
+[no-cd]
+profile-commands profile='all':
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    stable_file="$(mktemp)"
+    dev_file="$(mktemp)"
+    cleanup() {
+        rm -f "$stable_file" "$dev_file"
+    }
+    trap cleanup EXIT
+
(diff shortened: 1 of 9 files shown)


obey-agent (approved):
Verdict: Approve

Overview:
This is a clean addition to the build tooling. The command surface output is derived from the real registered Cobra tree instead of a hand-maintained list, and the stable/dev outputs match the current release-profile registration.

Key Findings:
- No review findings. I did not find correctness, maintainability, or behavior regressions worth blocking on.

What's Done Well:
- The hidden `fest __commands` helper keeps the source of truth in the CLI registration path.
- The buildutil wrapper is small and consistent with the existing build/test task style.
- The tests cover sorting, hidden-command exclusion, and the stable/dev release-profile split.

Staff Standard:
Yes. I would be comfortable submitting this as my own work.

PR #161, opened 2026-04-20
Use shared obey-shared buildutil in fest

lancekrogers:
## What changed

This replaces `fest`'s custom internal buildutil task/ui implementation with a thin wrapper around `github.com/Obedience-Corp/obey-shared/buildutil`.

The `internal/buildutil` entrypoint still exists at the same path, but now it delegates to the shared buildutil package and preserves `fest`-specific behavior through wrapper configuration:

- `BinaryName: fest`
- `MainPath: ./cmd/fest`
- fest-specific version ldflags
- Linux integration-build environment override
- Colima/Testcontainers environment setup for `integration` and `all`

The old local task/ui implementation under `internal/buildutil/tasks` and `internal/buildutil/ui` is removed.

This PR also updates `fest` to `github.com/Obedience-Corp/obey-shared v0.4.2`.

## Why

`obey-shared` now has the generic command-surface profile support needed by `fest`.

Moving `fest` to the shared buildutil eliminates duplicate build/test/profile-command logic and makes future profile-model improvements available through the shared package instead of another round of repo-local refactors, while keeping `fest`'s existing integration-test environment compatibility in the wrapper.

## Release context

The shared buildutil generic-profile follow-up was merged in `obey-shared` and tagged as `v0.4.2` before this branch was created.

## Validation

- `go test ./internal/buildutil`
- `go test ./internal/buildutil ./internal/commands ./internal/commands/release`
- `just build quick-stable`
- `just build quick-dev`
- `just build profile-commands stable`
- `just build profile-commands`
- `go run ./internal/buildutil test`
- `go run ./internal/buildutil clean`
- `just test all`


internal/buildutil/main.go
```diff
@@ -1,135 +1,63 @@
...
 func main() {
-	flag.BoolVar(&noColor, "no-color", false, "disable ANSI colours")
-	flag.BoolVar(&verbose, "v", false, "verbose output")
-	flag.Parse()
+	buildutil.Run(os.Args[1:], buildutil.BuildConfig{
+		BinaryName:  "fest",
+		MainPath:    "./cmd/fest",
+		SectionName: "Fest CLI",
+		LDFlags:     ldflags,
+		IntegrationBuildEnv: func() []string {
```
obey-agent, line 27: This keeps the old Linux build env override, but it drops the Colima/Testcontainers env that `fest`'s previous `tasks/integration.go` also set before running `go test` (`DOCKER_HOST=unix://$HOME/.colima/default/docker.sock` when unset, plus `TESTCONTAINERS_RYUK_DISABLED=true`). `obey-shared/buildutil` v0.4.2 runs the integration suite with the ambient env, so `just test integration` / `just test all` regress on the Colima setups the old code explicitly supported. Since this PR did not run the Docker-backed suite, I think we need either a shared `IntegrationTestEnv` hook or a repo-local integration wrapper before merging.

obey-agent (changes requested):
Verdict: Request Changes

The refactor removes a large amount of duplicated buildutil code, and the repo-local wrapper paths I checked are green locally: `go test ./internal/buildutil ./internal/commands ./internal/commands/release`, `just build quick-stable`, `just build quick-dev`, `just build profile-commands stable`, and `just build profile-commands dev`.

Key Findings:
- `internal/buildutil/main.go`: the wrapper only preserves the Linux build env. The old local integration task also configured `DOCKER_HOST`/`TESTCONTAINERS_RYUK_DISABLED` for Colima + testcontainers before invoking `go test`, but `obey-shared/buildutil` now runs the integration suite with the ambient env. That means this PR drops an existing compatibility path for `just test integration` / `just test all`, and the PR explicitly did not run the Docker-backed suite that would catch it.

What's Done Well:
- The `go run ./internal/buildutil ...` entrypoint stays stable.
- Stable/dev quick builds and command-surface reporting still work after the extraction.

Staff Standard:
- Not yet. I’d want integration-environment parity restored, or the Docker-backed suite to prove it is unnecessary, before merging this.

obey-agent (approved):
Verdict: Approve

Overview:
The wrapper now preserves the Colima/Testcontainers environment setup that the shared-buildutil migration initially dropped, and the new tests lock that behavior in. This keeps the deduplication win without regressing the local integration workflow.

What's Done Well:
- The environment setup is restored in the correct wrapper layer.
- The new tests cover both Colima defaulting and existing `DOCKER_HOST` preservation.
- The wrapper remains thin and focused.

Staff Standard:
Yes. I would be comfortable merging this.

PR #163, opened 2026-04-30
feat(ritual): tab completion for fest ritual run

lancekrogers:
## Summary

Replace the `fls` -> copy -> paste workflow for `fest ritual run` with a tab-cycle, matching the `fgo <tab>` UX.

## Changes

- **`internal/commands/ritual/run.go`** — adds `CompleteRitualName` and wires it to `newRunCommand` as `ValidArgsFunction`. Reuses `navigation.CollectFestivalsInStatus(festivalsRoot, "ritual")` and filters with case-insensitive substring match so completions agree with `findRitual`'s resolution rules.
- **`internal/commands/config/shell.go`** — adds bash, zsh, and fish completion for the `fest` binary by delegating to cobra's hidden `__complete` subcommand. Every cobra command on `fest` (including the new ritual run completion) now tab-completes for users who source `eval "$(fest shell-init <shell>)"`.
- **`internal/commands/ritual/run_test.go`** — new table-driven tests cover empty input, prefix and substring matches, case-insensitivity, ID-suffix matches, missing workspace, and empty ritual directory.

## Verification

```
just build dashboard          # green
just test unit                # 2014/2014 passed
./bin/fest __complete ritual run ""    # daily-ritual-RI-DR0001 + :4
./bin/fest __complete ritual run "dai" # substring match works
./bin/fest __complete ritual run "xx"  # empty result
```

Confirmed in zsh that `eval "$(fest shell-init zsh)"` registers `_fest_zsh` via `compdef` and that calling it with `words=(fest ritual run "")` produces the expected ritual name.

## Test plan

- [ ] Reviewer pulls branch and runs `just test unit`
- [ ] In a workspace with at least one ritual: `eval "$(./bin/fest shell-init zsh)"` then `fest ritual run <TAB>` cycles through ritual names
- [ ] `fest ritual run <selected>` runs the chosen ritual end-to-end
- [ ] Completion works for other subcommands too (free side benefit), e.g. `fest list <TAB>`


internal/commands/ritual/run.go
```diff
@@ -210,6 +212,35 @@ func autoCommitRitualRun(ctx context.Context, ritualName, ritualID, ritualPath,
 	return hash, nil
 }
 
+// CompleteRitualName provides shell completions for the ritual run argument.
+// It mirrors the substring matching semantics of findRitual so anything that
+// completes will resolve when invoked.
+func CompleteRitualName(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
obey-agent: `ValidArgsFunction` is also invoked when `args` already contains completed positional arguments. Since `fest ritual run` accepts exactly one arg, this should return no results when `len(args) > 0`; otherwise `fest ritual run <ritual> <TAB>` still offers more ritual names even though Cobra will reject the command. I reproduced this with `fest __complete ritual run daily-ritual-RI-DR0001 ""`, which returned all ritual names. A small guard returning `cobra.ShellCompDirectiveNoFileComp` would keep completion aligned with `cobra.ExactArgs(1)`.

obey-agent (commented):
## Review

Verdict: Commented - one completion correctness issue to fix before merge.

I left one inline comment in `CompleteRitualName`: `ValidArgsFunction` should stop completing once a ritual argument is already present. Right now `fest __complete ritual run daily-ritual-RI-DR0001 ""` returns all ritual names, so shell completion encourages a second positional argument that `cobra.ExactArgs(1)` rejects.

Everything else looks aligned: the main ritual-name matching works for empty, prefix, substring/case-insensitive, and ID suffix inputs; workspace and missing-dir cases degrade to no file completion; shell-init delegates through Cobra's `__complete` path for bash/zsh/fish.

Verification:
- `go test ./internal/commands/ritual ./internal/commands/config ./internal/navigation`
- `just test unit` (2014/2014)
- `just build dashboard`
- manual `./bin/fest __complete ritual run ...` smoke in a scratch festival workspace

lancekrogers:
Both review points addressed in 61e859f:

1. **`ValidArgsFunction` guard for filled args** — `CompleteRitualName` now returns `(nil, ShellCompDirectiveNoFileComp)` when `len(args) > 0`. Verified: `fest __complete ritual run daily-ritual-RI-DR0001 ""` returns no completions, matching `cobra.ExactArgs(1)`.

2. **Filesystem-mutating tests must run in the container harness** — deleted `internal/commands/ritual/run_test.go` (used `t.TempDir()` to scaffold `festivals/ritual/` on the host) and replaced it with `tests/integration/ritual_completion_test.go`, which exercises `fest __complete ritual run` end-to-end inside the shared alpine container. Coverage is the same: empty input, prefix/substring/case-insensitive match, ID-suffix match, no-match, the new len(args)>0 guard, and the no-workspace path.

Verification:
- `go test -tags=integration -run 'TestRitualRunCompletion|TestRitualRun_AutoCommit' -v ./tests/integration/...` passes (7+2 sub-cases green).
- Pre-existing integration failures on this branch are unrelated to this PR.

obey-agent (commented):
## Re-review

Verdict: Approve.

The latest push resolves my prior inline finding: `CompleteRitualName` now stops completing once the single positional argument is already filled, and the new integration test covers that behavior through Cobra's `__complete` path. The test move out of a host `t.TempDir()` setup and into the shared container harness matches the repo's filesystem-isolation rule.

Key findings:
- Prior completion correctness issue is fixed. Manual smoke confirmed `fest __complete ritual run daily-ritual-RI-DR0001 ""` now returns no ritual suggestions.
- No new blocking findings in `internal/commands/config/shell.go`, `internal/commands/ritual/run.go`, or `tests/integration/ritual_completion_test.go`.
- Note: `just test integration-raw` still fails in existing guidance implementation tests (`TestImplementationMode_PhaseBoundaries`, `TestImplementationMode_Completion`) with `FESTIVAL STATUS UNDETERMINED`. I do not see a path from this PR's completion-only changes to those failures; the new ritual completion integration test passes in isolation.

What's done well:
- Completion uses the existing navigation collector and Cobra's standard `ValidArgsFunction` surface rather than a parallel lookup path.
- Coverage exercises empty input, prefix/substring/case-insensitive matching, ID-suffix matching, no-match, missing workspace, and the filled-arg guard in the container harness.
- Shell init delegates to `command fest __complete`, which avoids recursion through the shell function wrapper.

Verification:
- `go test ./internal/commands/ritual ./internal/commands/config ./internal/navigation`
- `just test unit` (2011/2011)
- `just build dashboard`
- `DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock" TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -timeout 3m ./tests/integration -run '^TestRitualRunCompletion$'`
- Manual `./bin/fest __complete ritual run ...` smoke in a scratch workspace
- `./bin/fest shell-init bash|zsh|fish` snippet checks for the new `fest` completion hooks

Staff Standard: Yes. I would be comfortable submitting this change after the unrelated integration-suite health is handled separately.

PR #166, opened 2026-05-11
fix: guard ritual templates from execution commands

lancekrogers:
## Summary
- block `fest next` when run directly inside `festivals/ritual/<ritual>` templates
- block `fest task ...` and `fest workflow ...` commands on ritual templates with an instruction to use `fest ritual run` first
- keep copied ritual runs under `festivals/active/` executable even if their metadata still records ritual status
- render the ritual-template block from an embedded agent template instead of hardcoded string-builder output
- address review feedback: keep `EnforcePreActive` as the canonical task/next gate path, pass `context.Context` into `EnforceRitualRun`, and use workspace constants for festival/ritual path components

## Verification
- `go test ./embedded/templates/agent -run 'Test(AllTemplatesRender|ContractValidation)'`
- `go test ./internal/lifecycle -run 'Test(IsRitualTemplatePath|EnforceRitualRun)'` selected only the pure ritual tests: `TestIsRitualTemplatePath`, `TestEnforceRitualRun_BlocksRitualTemplate`, `TestEnforceRitualRun_AllowsActiveRitualRunPath`, and `TestEnforceRitualRun_RespectsCanceledContext`
- `go test ./internal/commands/next ./internal/commands/task ./internal/commands/workflow -run '^$'` (compile-only; no tests executed)
- `git diff --check main...HEAD`
- audited the added diff for filesystem-mutating calls (`TempDir`, `MkdirAll`, `WriteFile`, `Chdir`, `Remove`, `RemoveAll`, `chmod`, `git init`, `CreateTemp`, `MkdirTemp`): no matches

Note: I did not run the full `./internal/lifecycle` package or broader host test suite because existing tests in that package mutate the filesystem and must run through a containerized harness.

### DIFF
--- a/internal/lifecycle/gate.go
+++ b/internal/lifecycle/gate.go
@@ -6,10 +6,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Obedience-Corp/fest/embedded/templates/agent"
 	"github.com/Obedience-Corp/fest/internal/config"
 	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/guidance"
 	"github.com/Obedience-Corp/fest/internal/progress"
+	"github.com/Obedience-Corp/fest/internal/workspace"
 )
 
 // EnforceOptions controls which festival phase the gate evaluates.
@@ -35,6 +37,10 @@ type EnforceOptions struct {
 // Fail-closed: if the festival config or phase scan cannot be determined,
 // the gate blocks rather than silently allowing the operation.
 func EnforcePreActive(ctx context.Context, festivalPath string, opts EnforceOptions) error {
+	if err := EnforceRitualRun(ctx, festivalPath, opts.Reason); err != nil {
+		return err
+	}
+
 	festCfg, err := config.LoadFestivalConfig(festivalPath, "")
 	if err != nil {
 		return emitFailClosed(festivalPath, opts.Reason,
@@ -76,6 +82,32 @@ func EnforcePreActive(ctx context.Context, festivalPath string, opts EnforceOpti
 	return emitBlock(festivalPath, status, opts.Reason)
 }
 
+// EnforceRitualRun blocks execution-oriented commands from operating directly
+// on a ritual template under festivals/ritual/. Ritual templates must be copied
+// into active/ with `fest ritual run` before users work through them.
+func EnforceRitualRun(ctx context.Context, festivalPath, reason string) error {
+	if ctx != nil {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+	}
+	if !IsRitualTemplatePath(festivalPath) {
+		return nil
+	}
+	return emitRitualTemplateBlock(ctx, festivalPath, reason)
+}
+
+// IsRitualTemplatePath reports whether festivalPath is a direct child of a
+// festivals/ritual/ directory. Active ritual runs are copied under active/ and
+// must remain executable even if their fest.yaml still records ritual metadata.
+func IsRitualTemplatePath(festivalPath string) bool {
+	clean := filepath.Clean(festivalPath)
+	if filepath.Base(filepath.Dir(clean)) != workspace.RitualDir {
+		return false
+	}
+	return filepath.Base(filepath.Dir(filepath.Dir(clean))) == workspace.FestivalsDir
+}
+
 func phasePathFromTaskID(festivalPath, taskID string) string {
 	parts := strings.SplitN(taskID, "/", 2)
 	if parts[0] == "" {
@@ -109,6 +141,27 @@ func emitBlock(festivalPath, status, reason string) error {
 	sb.WriteString("Promote first: fest promote\n")
 
 	fmt.Print(sb.String())
+	return errors.ErrAlreadyPrinted
+}
+
+func emitRitualTemplateBlock(ctx context.Context, festivalPath, reason string) error {
+	if ctx != nil {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+	}
+
+	ritualName := filepath.Base(festivalPath)
+
+	output, err := agent.Render("lifecycle/ritual_template_block", map[string]string{
+		"RitualName": ritualName,
+		"Reason":     reason,
+	})
+	if err != nil {
+		return errors.Wrap(err, "rendering ritual template block").WithCode(errors.ErrCodeTemplate)
+	}
+
+	fmt.Print(output)
 	return errors.ErrAlreadyPrinted
 }
 
--- a/embedded/templates/agent/contracts.go
+++ b/embedded/templates/agent/contracts.go
@@ -308,6 +308,15 @@ var Contracts = map[string]TemplateContract{
 		ForbiddenTokens:  DefaultForbiddenTokens,
 		MinimumLength:    50,
 	},
+
+	// Lifecycle templates
+	"lifecycle/ritual_template_block": {
+		Name:             "lifecycle/ritual_template_block",
+		RequiredFields:   []string{"RitualName", "Reason"},
+		ExpectedSections: []string{"RITUAL TEMPLATE IS NOT A RUN", "fest ritual run"},
+		ForbiddenTokens:  DefaultForbiddenTokens,
+		MinimumLength:    100,
+	},
 }
 
 // Validate checks if rendered output satisfies the contract.
--- a/embedded/templates/agent/embed.go
+++ b/embedded/templates/agent/embed.go
@@ -13,4 +13,5 @@ import "embed"
 //go:embed ingest/*.tmpl
 //go:embed validate/*.tmpl
 //go:embed workflow/*.tmpl
+//go:embed lifecycle/*.tmpl
 var Templates embed.FS
--- a/embedded/templates/agent/lifecycle/ritual_template_block.tmpl
+++ b/embedded/templates/agent/lifecycle/ritual_template_block.tmpl
@@ -1,0 +1,19 @@
+{{/*
+TEMPLATE: lifecycle/ritual_template_block.tmpl
+PURPOSE: Blocks execution commands from operating directly on ritual templates.
+
+Ritual templates under festivals/ritual/ are repeatable templates. Users must
+create a copied run under festivals/active/ before using execution commands.
+*/}}
+STOP - RITUAL TEMPLATE IS NOT A RUN
+-----------------------------------
+Ritual "{{.RitualName}}" lives in festivals/ritual/.
+That directory is the repeatable template, not an execution run.
+
+Create an active run first:
+  fest ritual run {{.RitualName}}
+
+Then work from the copied run under festivals/active/ with fest next,
+fest task, or fest workflow commands.
+{{if .Reason}}Action "{{.Reason}}" is blocked on the ritual template.
+{{end}}
--- a/embedded/templates/agent/registry_test.go
+++ b/embedded/templates/agent/registry_test.go
@@ -131,6 +131,8 @@ func minimalTestData(templateName string) any {
 		return validateData(templateName, commonFields)
 	case strings.HasPrefix(templateName, "workflow/"):
 		return workflowData(templateName, commonFields)
+	case strings.HasPrefix(templateName, "lifecycle/"):
+		return lifecycleData()
 	default:
 		// Fallback: return common fields for any unknown template
 		return commonFields
@@ -218,6 +220,13 @@ func ingestData(common map[string]any) map[string]any {
 		"PositionSection": common["PositionSection"],
 		"ContextSection":  common["ContextSection"],
 		"ProgressCmd":     common["ProgressCmd"],
+	}
+}
+
+func lifecycleData() map[string]any {
+	return map[string]any{
+		"RitualName": "daily-ritual-RI-DR0001",
+		"Reason":     "fest next",
 	}
 }
 
--- a/internal/commands/workflow/status.go
+++ b/internal/commands/workflow/status.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Obedience-Corp/fest/internal/commands/shared"
 	"github.com/Obedience-Corp/fest/internal/guidance"
 	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
+	"github.com/Obedience-Corp/fest/internal/lifecycle"
 	"github.com/Obedience-Corp/fest/internal/progress"
 	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/ui"
@@ -137,6 +138,9 @@ func getWorkflowNavigator(ctx context.Context) (*wf.Navigator, error) {
 		if err != nil {
 			return nil, fmt.Errorf("not in a festival: %w", err)
 		}
+	}
+	if err := lifecycle.EnforceRitualRun(ctx, festivalPath, "fest workflow"); err != nil {
+		return nil, err
 	}
 
 	var phasePath string
--- a/internal/lifecycle/gate_test.go
+++ b/internal/lifecycle/gate_test.go
@@ -157,6 +157,89 @@ func TestEnforcePreActive_MalformedConfig_FailsClosed(t *testing.T) {
 	}
 }
 
+func TestIsRitualTemplatePath(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{
+			name: "ritual template",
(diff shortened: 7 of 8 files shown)


obey-agent (approved):
Verdict: Approve.

Reviewed at head `8e13a4e`. Adds a path-based `EnforceRitualRun` gate that blocks execution-oriented commands (`fest next`, `fest task ...`, `fest workflow ...`) from operating on ritual templates under `festivals/ritual/`. The design is clean: the gate is purely lexical on the path (correct, since active ritual runs under `festivals/active/` must remain executable even if their `fest.yaml` still records ritual metadata), and the wiring covers every entry point that matters.

What I verified

- **`internal/lifecycle/gate.go:96-108` — `IsRitualTemplatePath`** is a path-only check: `filepath.Base(filepath.Dir(clean)) == "ritual"` AND `filepath.Base(filepath.Dir(filepath.Dir(clean))) == "festivals"`. Doesn't read disk, doesn't open `fest.yaml`. Correct: the location is the source of truth for template-vs-run distinction, and the doc comment explains why ("Active ritual runs are copied under active/ and must remain executable even if their fest.yaml still records ritual metadata"). Test pins the four meaningful cases: ritual template (true), active run (false), nested-inside-template (false), non-festival ritual path (false).
- **`internal/lifecycle/gate.go:36-39` — `EnforcePreActive` now calls `EnforceRitualRun` first.** That means every command already using `EnforcePreActive` (`progress/task_update.go`, `status/handlers_entity.go` × 4, `workflow/skip.go`, `workflow/reset.go`, `workflow/reject.go`, `workflow/approve.go`, `next/next.go` itself) gets ritual-template protection for free. No duplicate gating logic.
- **Direct `EnforceRitualRun` calls** in `next.go`, `task/{blocked,completed,edit,reset,show}.go`, and `workflow/status.go:142` (inside `getWorkflowNavigator`). `getWorkflowNavigator` is the single chokepoint for every `fest workflow ...` subcommand (`status`, `show`, `advance`, `skip`, `reset`, `reject`, `approve`), so one insertion covers all seven. Confirmed by grep — every workflow command calls `getWorkflowNavigator` before doing anything else.
- **`emitRitualTemplateBlock`** matches the style of `emitBlock` / `emitFailClosed`: builds the message into a `strings.Builder`, prints to stdout, returns `errors.ErrAlreadyPrinted` so the cobra layer doesn't double-print. Message names the ritual, points the user at `fest ritual run <name>`, and includes the action that was blocked.
- **The redundancy of `EnforceRitualRun` in commands that also call `EnforcePreActive`** (e.g. `next.go` calls it explicitly at line 126 and then `EnforcePreActive` at line 151, which also calls `EnforceRitualRun`) is defensive — short-circuits earlier with the ritual-specific message, and survives future refactors of `EnforcePreActive`. Worth keeping.

Verification I ran (at head `8e13a4e`)

- `go build ./...` clean
- `go test ./internal/lifecycle -count=1 -v` pass: `TestIsRitualTemplatePath` (all 4 sub-cases), `TestEnforceRitualRun_BlocksRitualTemplate`, `TestEnforceRitualRun_AllowsActiveRitualRunPath`, plus the existing `TestEnforcePreActive_*` suite.
- `go test ./... -count=1 -timeout=180s` pass across all packages.
- `gofmt -l` clean on every file this PR touched (`internal/lifecycle/*`, `internal/commands/next/next.go`, `internal/commands/task/*.go`, `internal/commands/workflow/status.go`).
- `go vet ./...` clean.

Pre-existing items (not this PR's fault, not blocking)

- `gofmt -l internal/commands` flags 8 files on `main` already (`chain/complete.go`, `gendocs.go`, `markers/scaffold.go`, `next/validation.go`, `research/link.go`, `show/stats.go`, `status/lifecycle_gate_test.go`, `tui/helpers.go`). All untouched by this PR. Worth a separate cleanup pass.

Ready to merge.

internal/commands/next/next.go
```diff
@@ -123,6 +123,9 @@ func runNext(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "not inside a festival")
 	}
+	if err := lifecycle.EnforceRitualRun(festivalPath, "fest next"); err != nil {
```
obey-agent: **Dead code: this `EnforceRitualRun` call is redundant.** `EnforcePreActive` at line 151 already runs `EnforceRitualRun` as its first step (gate.go:39-41). The explicit call here means the ritual-template gate runs *twice* on every `fest next` invocation — once at line 126, once inside `EnforcePreActive`. Same pattern in `task/blocked.go:46`, `task/completed.go:43`, `task/edit.go:37`, `task/reset.go:41`, `task/show.go:48` — six total redundant calls.

Pick one of:

1. **Remove the explicit calls in commands that already use `EnforcePreActive`.** Keep `EnforceRitualRun` integrated into `EnforcePreActive` and the standalone call in `getWorkflowNavigator` (which feeds `workflow show`/`status` that don't follow with `EnforcePreActive`).
2. **Remove the integration in `EnforcePreActive` (gate.go:39-41) and rely on explicit calls everywhere.** Less clean — every new command becomes responsible for adding two guards instead of one.

Option 1 is what a less-experienced engineer maintaining this code six months from now should encounter — one canonical gate path, not two. The current state forces every reader to ask "can the second call ever differ from the first?" Answer: no, but that's not obvious from the code.

Staff standard: would I submit this with both gates wired up? No — it's the kind of duplication that quietly grows because copy-paste is easier than asking why.

internal/lifecycle/gate.go
```diff
@@ -76,6 +81,27 @@ func EnforcePreActive(ctx context.Context, festivalPath string, opts EnforceOpti
...
+	}
+	return emitRitualTemplateBlock(festivalPath, reason)
+}
+
+// IsRitualTemplatePath reports whether festivalPath is a direct child of a
+// festivals/ritual/ directory. Active ritual runs are copied under active/ and
+// must remain executable even if their fest.yaml still records ritual metadata.
+func IsRitualTemplatePath(festivalPath string) bool {
+	clean := filepath.Clean(festivalPath)
+	if filepath.Base(filepath.Dir(clean)) != "ritual" {
```
obey-agent: **Magic strings — use the existing constant.** `internal/workspace/workspace.go:25` already exports `FestivalsDir = "festivals"` and is the package that owns the campaign layout. The PR adds two more bare string literals (`"ritual"` here and `"festivals"` at line 102) to a file that already drifts on this — `gate.go:56,120` uses bare `"planning"` / `"ready"` strings for status checks.

Per the staff standard (no magic strings where semantics matter): these are not opaque cosmetic strings, they're filesystem path components that determine whether commands execute. A rename of the `festivals/ritual/` directory layout would silently break this gate everywhere it's hardcoded.

Required fix:

1. Use `workspace.FestivalsDir` at line 102.
2. Lift `"ritual"` to a constant in `internal/workspace/workspace.go` (e.g., `RitualDir = "ritual"`) alongside `FestivalsDir`, and reference it here.

Don't widen this to a refactor of the pre-existing status strings (`planning`/`ready`) in `emitBlock` — that's its own PR. But this PR is adding *new* magic strings to a function that's defining a new contract; that should be done with constants from day one.

internal/lifecycle/gate.go
```diff
@@ -76,6 +81,27 @@ func EnforcePreActive(ctx context.Context, festivalPath string, opts EnforceOpti
 	return emitBlock(festivalPath, status, opts.Reason)
 }
 
+// EnforceRitualRun blocks execution-oriented commands from operating directly
+// on a ritual template under festivals/ritual/. Ritual templates must be copied
+// into active/ with `fest ritual run` before users work through them.
+func EnforceRitualRun(festivalPath, reason string) error {
```
obey-agent: **`EnforceRitualRun` doesn't accept `context.Context`** while its sibling `EnforcePreActive` (line 38) does. Both live in the same package and serve the same lifecycle-gate role, but they have different signatures.

This function does I/O — `emitRitualTemplateBlock` renders an embedded template and writes to stdout via `fmt.Print`. Per CLAUDE.md's Go standard ("Always pass `context.Context` as the first parameter for any I/O operation"), and for consistency with the rest of the lifecycle package, this should take `ctx context.Context` as the first parameter.

Required fix:

```go
func EnforceRitualRun(ctx context.Context, festivalPath, reason string) error {
    if !IsRitualTemplatePath(festivalPath) {
        return nil
    }
    return emitRitualTemplateBlock(ctx, festivalPath, reason)
}
```

Update call sites accordingly. Cancellation isn't honored downstream today (template render is in-memory + `fmt.Print`), but the contract is what matters — future work that swaps `fmt.Print` for a ctx-aware writer (or adds I/O to the template render path) shouldn't require a signature change and call-site sweep.

Note: if you address the redundancy finding (the other inline comment) by removing the six explicit call sites that already follow with `EnforcePreActive`, the only remaining caller is `getWorkflowNavigator` and the signature change is trivial.

obey-agent (changes requested):
Verdict: Request Changes.

Re-reviewed at head `35e31685` with the staff-review skill freshly invoked. My prior approval at `8e13a4e` was insufficient — I was reading for diff correctness, not staff quality. Concrete issues below; each is a real correctness/maintainability concern, not style preference.

Note on the new commit (`35e3168`): moving the inline string-builder to an embedded `lifecycle/ritual_template_block.tmpl` with a template contract is the right call. That's the kind of refactor I should have asked for in the prior round and didn't. The template contract validation (`RequiredFields: []string{"RitualName", "Reason"}`, `ExpectedSections: ...`) is a good pattern.

Blocking

1. **Six redundant explicit `EnforceRitualRun` calls.** `EnforcePreActive` (gate.go:39-41) now calls `EnforceRitualRun` as its first step. Six commands (`next.go:126`, `task/{blocked,completed,edit,reset,show}.go`) call `EnforceRitualRun` explicitly *and then* call `EnforcePreActive` afterward. The ritual-template gate runs twice per invocation. Either remove the explicit calls in commands that follow with `EnforcePreActive` (preferred — keeps a single canonical path), or remove the integration in `EnforcePreActive` and require all callers to gate explicitly. Having both forces every future reader to verify the two paths can't diverge. See inline on `next.go:126`.

2. **Magic strings `"ritual"` and `"festivals"` in `IsRitualTemplatePath`** (gate.go:99,102). `internal/workspace/workspace.go:25` already exports `FestivalsDir = "festivals"`. The PR adds new path-component literals to a function that defines a brand-new contract — those should be constants from day one, not inherited drift. Use `workspace.FestivalsDir`; lift `"ritual"` to a sibling constant in `workspace.go`. Don't widen to refactoring pre-existing `"planning"`/`"ready"` strings — that's a separate cleanup. See inline on `gate.go:99`.

3. **`EnforceRitualRun` doesn't accept `context.Context`** while `EnforcePreActive` in the same package does. The function does I/O (template render + `fmt.Print`). Per repo Go standard, ctx is the first parameter for any I/O operation. Even though cancellation isn't honored downstream today, the signature should be ctx-first so future I/O additions (e.g. swapping `fmt.Print` for a ctx-aware writer) don't require a sweep. See inline on `gate.go:87`.

What I verified (at head `35e31685`)

- `go build ./...` clean.
- `go test ./internal/lifecycle ./internal/commands/... ./embedded/templates/agent/...` pass — `TestIsRitualTemplatePath` (4 cases), `TestEnforceRitualRun_*` (2 cases), template registry validation, and the wider command suite all green.
- `gofmt -l` clean on the files this PR touched. The 8 pre-existing `gofmt -l internal/commands` failures on `main` are unrelated.
- `go vet ./...` clean.
- Template contract at `embedded/templates/agent/contracts.go:312-318` correctly requires `RitualName` and `Reason`, expects the right output sections, and inherits `DefaultForbiddenTokens`. Registry test data wires through `lifecycleData()` correctly.
- Wiring covers the right surface: `workflow/status.go:142`'s `getWorkflowNavigator` chokepoint catches the read-only workflow commands (`show`, `status`) that don't follow with `EnforcePreActive`. That's the legitimately-needed standalone call site.

Not flagged (deliberately not blocking)

- `fest task show` (read-only) being blocked on ritual templates. The pre-existing comment in `show.go:62-67` documents a legitimate "bypass prevention" rationale via `EnforcePreActive`'s phase-type check — defensible design choice, not a UX bug.
- Pre-existing gofmt failures on 8 files under `internal/commands` (untouched by this PR).

Staff Standard: No, I would not submit the current diff as my best work. The ritual-template gate is *correct* but the implementation duplicates work in 6 places, hardcodes path components that already have a canonical constant, and breaks signature consistency with the rest of its own package. Two of those will be expensive to fix later if they ship — explicit calls accrete, magic strings get copied. I will re-review and approve once the three items above are addressed.

obey-agent (approved):
Verdict: Approve.

Re-reviewed at head `0ebde33` with the staff-review skill freshly invoked. All three blocking items from my prior `CHANGES_REQUESTED` review at `35e31685` are addressed cleanly, with a new regression test pinning the context contract.

What I verified

1. **Redundant `EnforceRitualRun` calls removed** in `next/next.go`, `task/{blocked,completed,edit,reset,show}.go`. Six commands now flow through `EnforcePreActive` as their single canonical gate path, and `EnforcePreActive` (gate.go:40) still wires `EnforceRitualRun` as its first step. The only standalone `EnforceRitualRun` call is in `workflow/status.go:142`'s `getWorkflowNavigator`, which serves `workflow show` / `workflow status` — those don't follow with `EnforcePreActive`, so the standalone call is legitimately required there.
2. **Magic strings lifted to constants.** `internal/workspace/workspace.go:27` adds `RitualDir = "ritual"` alongside the existing `FestivalsDir = "festivals"`. `IsRitualTemplatePath` (gate.go:104-107) now references `workspace.RitualDir` and `workspace.FestivalsDir`. The other `"ritual"` usages in the codebase (`ui/styles.go`, `contract/entries.go`, `types/festival_types.go`, `workflow/defaults.go`, `tui/explore/*`) are for the *festival type* enum value, not the directory name — correctly out of scope for this PR.
3. **`EnforceRitualRun` now takes `context.Context` first.** Signature updated (gate.go:88), all callers updated (`workflow/status.go:142` and the internal call from `EnforcePreActive` at gate.go:40). The function checks `ctx.Err()` at entry and propagates cancellation. `TestEnforceRitualRun_RespectsCanceledContext` (gate_test.go:233) is a clean regression test — pre-cancel the ctx, assert `context.Canceled` is returned without the I/O path executing.

Verification I ran (at head `0ebde33`)

- `go build ./...` clean.
- `go test ./internal/lifecycle ./internal/commands/... ./internal/workspace -count=1 -timeout=120s` pass. All three `TestEnforceRitualRun_*` tests green.
- `gofmt -l` clean on every file this PR touched. The 8 pre-existing `internal/commands` gofmt failures on `main` (`chain/complete.go`, `gendocs.go`, `markers/scaffold.go`, `next/validation.go`, `research/link.go`, `show/stats.go`, `status/lifecycle_gate_test.go`, `tui/helpers.go`) are unchanged and out of scope.
- `go vet ./...` clean.
- Wiring still covers the right surface: `EnforcePreActive` is the single gate used by `next.go`, all five mutating `task/*.go` commands (plus `task/show.go`), and all six mutating `workflow/*.go` commands; `getWorkflowNavigator` covers `workflow show` / `workflow status` for the read-only path. Nothing in the package now runs the ritual-template check twice on the same code path.

What's done well

- The `workspace` package is the right home for `RitualDir` — it already owns the campaign-layout vocabulary via `FestivalsDir`, `DotFestival`, `StateDir`, `MarkerFile`. Future code touching the layout has one obvious place to look.
- The ctx-cancel test (`TestEnforceRitualRun_RespectsCanceledContext`) verifies the *behavior* of the new ctx parameter, not just its presence — pre-cancels and asserts cancellation propagates. That's the right test shape for a ctx contract.
- The embedded template (`embedded/templates/agent/lifecycle/ritual_template_block.tmpl`) with a contract entry in `contracts.go` is the correct evolution from the inline `strings.Builder` we discussed — future block-message edits don't require touching Go code.

Staff Standard: Yes. Two reviews ago I would not have flagged the issues that are now fixed; one review ago I caught them but only after re-loading the staff-review checklist. This iteration is what the PR should have looked like from the start, and the diff is now small, focused, and traceable. I would be comfortable shipping this.

Minor observation (not blocking, would not flag in isolation): the `if ctx != nil { ... }` guard inside both `EnforceRitualRun` and `emitRitualTemplateBlock` is defensive-over-idiomatic — production callers should always pass a non-nil ctx, and Go convention is to let it panic when they don't. The inner helper's check is also redundant since `EnforceRitualRun` is the only caller and already checked. If you happen to be cleaning these up, drop the `ctx != nil` guards and remove the duplicate check in the helper. Not a blocker for this merge.

Ready to merge.

PR #167, opened 2026-05-12
fix: include lifecycle events in promote autocommit

lancekrogers:
## Summary
- include `festivals/.festival/.state/festival_events.jsonl` in the status-change autocommit scope when the lifecycle event log exists
- centralize the path augmentation in `AutoCommitStatusChange`, so `fest promote` and `fest status set` share the fix
- add pure path-list tests that verify the lifecycle event log is appended, deduplicated, and skipped when absent

## Verification
- `go test ./internal/commands/status -run 'TestStatusChangeCommitPaths'`
- `go test ./internal/commands/promote -run '^$'` (compile-only; no tests executed)
- `git diff --check`
- added-diff audit for filesystem-mutating test calls (`TempDir`, `MkdirAll`, `WriteFile`, `Chdir`, `Remove`, `RemoveAll`, `chmod`, `git init`, `CreateTemp`, `MkdirTemp`, `os.Rename`, `exec.Command`): no matches

Note: I did not run the broader host test suite because existing tests mutate the filesystem and must run through a containerized harness.

### DIFF
--- a/internal/commands/status/handlers_festival.go
+++ b/internal/commands/status/handlers_festival.go
@@ -17,6 +17,7 @@ import (
 	"github.com/Obedience-Corp/fest/internal/frontmatter"
 	"github.com/Obedience-Corp/fest/internal/id"
 	"github.com/Obedience-Corp/fest/internal/navigation"
+	"github.com/Obedience-Corp/fest/internal/registry"
 	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/ui"
 	"github.com/Obedience-Corp/fest/internal/ui/theme"
@@ -470,6 +471,13 @@ func AutoCommitStatusChange(ctx context.Context, festivalName, festivalID, oldSt
 		return "", errors.New("workspace not available in context")
 	}
 
+	if ws.FestivalsPath != "" {
+		eventsPath := registry.GetEventsPath(ws.FestivalsPath)
+		if _, err := os.Stat(eventsPath); err == nil {
+			changedPaths = statusChangeCommitPaths(changedPaths, eventsPath)
+		}
+	}
+
 	// Stage only the paths affected by the status change (lock-aware with retry).
 	if err := commitkit.StageFiles(ctx, ws.Root, changedPaths...); err != nil {
 		return "", errors.Wrap(err, "stage")
@@ -513,4 +521,28 @@ func AutoCommitStatusChange(ctx context.Context, festivalName, festivalID, oldSt
 	}
 
 	return hash, nil
+}
+
+func statusChangeCommitPaths(changedPaths []string, eventsPath string) []string {
+	paths := make([]string, 0, len(changedPaths)+1)
+	seen := make(map[string]struct{}, len(changedPaths)+1)
+
+	appendPath := func(path string) {
+		if path == "" {
+			return
+		}
+		key := filepath.Clean(path)
+		if _, ok := seen[key]; ok {
+			return
+		}
+		seen[key] = struct{}{}
+		paths = append(paths, path)
+	}
+
+	for _, path := range changedPaths {
+		appendPath(path)
+	}
+	appendPath(eventsPath)
+
+	return paths
 }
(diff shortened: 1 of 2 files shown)


obey-agent (approved):
Verdict: Approve.

Reviewed at head `bad954b` with the staff-review skill freshly invoked. Small, focused fix that addresses a real bug class — without including `festivals/.festival/.state/festival_events.jsonl` in the autocommit scope, `fest promote` and `fest status set` would leave a dirty working tree containing the lifecycle event log every time a status changes. The PR fixes this in one well-chosen place (the shared `AutoCommitStatusChange`), with a pure helper that handles the path-list manipulation and three targeted unit tests.

What I verified

1. **`AutoCommitStatusChange` is the right shared chokepoint.** Both `fest promote` and `fest status set` call it, so centralizing the path augmentation here means the fix lands in every code path that triggers a status change autocommit. No parallel implementation needed.

2. **`statusChangeCommitPaths` (handlers_festival.go:526-548) is a pure function** — extracted cleanly from the side-effecting caller, takes a `[]string` of changed paths plus the events path, returns a deduplicated `[]string`. Uses `filepath.Clean` as the dedup key but preserves the caller's original path strings in the output (so a caller passing `/foo/./bar` and the helper passing `/foo/bar` correctly dedup without losing either form). Pre-allocates with `make([]string, 0, len(changedPaths)+1)`.

3. **The `os.Stat` guard before adding the events path** is the correct defensive shape: skip the augmentation when the file doesn't exist yet (early in a festival's lifecycle, before any events have been recorded). The pre-existing `commitkit.StageFiles` downstream would otherwise fail with "no such file."

4. **`registry.GetEventsPath`** is used instead of a path literal — consistent with the rest of the registry/state directory access in this codebase. `internal/registry/io.go:32` is the canonical owner of that path layout.

5. **Tests cover the three meaningful cases:**
   - `TestStatusChangeCommitPaths_AppendsLifecycleEventLog`: happy path — events path is appended to the end of the path list.
   - `TestStatusChangeCommitPaths_DeduplicatesLifecycleEventLog`: events path already in `changedPaths` is not duplicated. Order-preserving (return matches input order).
   - `TestStatusChangeCommitPaths_EmptyEventPathLeavesScopeUnchanged`: empty events path is a no-op.

Verification I ran (at head `bad954b`)

- `go build ./...` clean.
- `go test ./internal/commands/status -count=1 -timeout=60s` pass (403ms).
- All three new `TestStatusChangeCommitPaths_*` tests pass individually.
- `go vet ./internal/commands/status/...` clean.
- `gofmt -l` is unchanged vs `main`: `lifecycle_gate_test.go` is the only flagged file and it's pre-existing tech debt, not touched by this PR.
- The PR's stated coverage gap (broader host test suite deferred because filesystem-mutating tests must run through the containerized harness) is consistent with the repo's policy on filesystem tests. The pure-function tests stay within that policy.

What's done well

- **Extracting the pure path-list logic as a separate function with its own test file** (`handlers_festival_paths_test.go`). Lets the contract be pinned without involving filesystem mocks or temp dirs, and keeps the integration-level question ("does AutoCommitStatusChange thread the events path through correctly?") as a separate concern testable in the containerized harness.
- **The dedup-by-`filepath.Clean` choice.** Avoids accidental double-staging when a caller has already included the events path, but doesn't mangle the original caller-provided path strings (which might matter for `git add` semantics if anyone ever passes a relative path).

Not flagged (deliberately not blocking)

- `handlers_festival.go` is now 548 lines (over the 500-line guideline in CLAUDE.md). The file was already at ~524 on `main` — pre-existing, not introduced by this PR. The new helper is tightly coupled to `AutoCommitStatusChange` so co-locating it is defensible; if you decide to split the file later, lifting the path-helper into its own `commit_paths.go` (with the test file already named `handlers_festival_paths_test.go`) would be a clean cut.

Staff Standard: Yes. The fix is small, the contract is pure-function-testable, the integration point is the shared chokepoint that both `promote` and `status set` flow through, and the tests cover the meaningful state-space. I would be comfortable submitting this as my own work.

Ready to merge.

PR #174, opened 2026-05-16
Add auto-write support to fest commit

lancekrogers:
Summary
- Add --auto-write and -aw support to fest commit.
- Reuse camp commit hook handling so the configured command runs from the commit target context.
- Keep normal -m commit behavior unchanged and reject mixed --message/--auto-write usage.

Dependency
- Depends on Obedience-Corp/camp#290.
- Uses the camp branch pseudo-version until that PR is merged and tagged/released.

Validation
- git diff --check
- go test ./internal/commands ./internal/commands/commit
- just lint vet
- just test unit-raw

### DIFF
--- a/internal/commands/autowrite_alias.go
+++ b/internal/commands/autowrite_alias.go
@@ -1,0 +1,26 @@
+package commands
+
+// normalizeAutoWriteAlias rewrites the exact -aw token for fest commit before
+// cobra parses flags. Cobra/pflag only supports single-character shorthand
+// flags.
+func normalizeAutoWriteAlias(args []string) []string {
+	out := make([]string, len(args))
+	copy(out, args)
+	if !isAutoWriteCommitInvocation(out) {
+		return out
+	}
+	for i, arg := range out {
+		if arg == "--" {
+			break
+		}
+		if arg == "-aw" {
+			out[i] = "--auto-write"
+		}
+	}
+	return out
+}
+
+func isAutoWriteCommitInvocation(args []string) bool {
+	command, _ := findFirstPositionalArg(args)
+	return command == "commit"
+}
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.25.6
 
 require (
 	github.com/Masterminds/sprig/v3 v3.3.0
-	github.com/Obedience-Corp/camp v0.2.2
+	github.com/Obedience-Corp/camp v0.2.6-0.20260516214523-02acfc0f9d91
 	github.com/Obedience-Corp/obey-shared v0.4.2
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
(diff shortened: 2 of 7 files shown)


internal/commands/autowrite_alias.go
```diff
@@ -0,0 +1,14 @@
+package commands
+
+// normalizeAutoWriteAlias rewrites the exact -aw token before cobra parses
+// flags. Cobra/pflag only supports single-character shorthand flags.
+func normalizeAutoWriteAlias(args []string) []string {
+	out := make([]string, len(args))
+	copy(out, args)
+	for i, arg := range out {
```
obey-agent, line 12: This has the same scope problem camp just fixed: `Execute()` calls `normalizeAutoWriteAlias(os.Args)` before plugin dispatch and before Cobra identifies the command, so every exact `-aw` token anywhere in the invocation becomes `--auto-write`. That changes unrelated commands and plugin argv, e.g. `fest external-tool -aw` would dispatch the plugin with `--auto-write` instead of the user-provided `-aw`.

Please gate the rewrite to the actual `fest commit` invocation (and add tests for unrelated commands/plugin-style invocations) rather than normalizing all command lines globally. This is a behavioral regression in the CLI/plugin boundary, not just an alias implementation detail.

obey-agent (commented):
Verdict: Request Changes

Overview: The commit hook integration is otherwise narrow and reuses `camp/pkg/commitkit` correctly for the configured message writer. However, the `-aw` compatibility shim currently rewrites every exact `-aw` token in `os.Args` before plugin dispatch or command resolution, which changes behavior outside `fest commit`.

Key Findings:
- Inline: `internal/commands/autowrite_alias.go` rewrites `-aw` globally, so unrelated commands and `fest-*` plugins receive `--auto-write` instead of the original user argument. This should be scoped to the actual `fest commit` invocation and covered with unrelated/plugin-style tests.

What's Done Well:
- `--auto-write` and `--message` are treated as mutually exclusive.
- The generated message runs through the existing festival/campaign tagging path rather than duplicating commit logic.
- The hook command is loaded from campaign config via `commitkit.AutoWriteCommitMessage`, preserving the config-driven behavior from camp.

Verification:
- Read all changed files.
- `gofmt -l` on changed Go files: clean.
- `git diff --check origin/main...HEAD`: clean.
- `go test -count=1 ./internal/commands ./internal/commands/commit`: pass.
- `just test unit-raw`: pass.
- `just lint vet`: pass.
- `just build quick`: pass.

Staff Standard: I would not merge this until the alias rewrite is scoped to `fest commit`; the rest of the implementation is close.

obey-agent (commented):
Verdict: Approve

No blocking issues found on the current head (`25af396`). This PR adds `fest commit --auto-write` plus the `-aw` alias so the commit message can be produced by the campaign-configured `hooks.commit_message.command` instead of requiring `--message`. The hook is loaded from `.campaign/campaign.yaml`, runs from the target Git repo root after the normal staging step, and its trimmed stdout becomes the raw message before the existing fest/campaign tag handling and campaign-root follow-up commit logic runs.

The previous alias-scoping concern is fixed: `-aw` is now only normalized when the first positional command is `commit`, and normalization stops at `--`. The new tests cover unrelated commands, plugin-style commands, global flags before `commit`, and the argument terminator.

Data/user behavior checked: this does not add a new storage format or persistent data file. It changes the `fest commit` UX and the resulting Git commit message source; the existing staging, tagging, submodule/root commit behavior remains in place. The one sequencing note is that `go.mod` still points at a camp pseudo-version from the dependent camp PR, so this should merge only after that dependency is reachable from the expected camp history or replaced by a tagged camp release.

Validation run locally:
- `gofmt -l internal/commands/autowrite_alias.go internal/commands/autowrite_alias_test.go internal/commands/root.go internal/commands/commit/commit.go internal/commands/commit/commit_test.go`
- `go test -count=1 ./internal/commands ./internal/commands/commit`
- `just test unit-raw`
- `just lint vet`
- `just build quick`

PR #179, opened 2026-05-19
fix(gates): show all phase buckets and preserve template on skip

### LINKED ISSUE #168: fest gates show/apply silently ignores custom quality gate template paths
## Summary

`fest gates show` / `fest gates apply` do not appear to resolve or surface custom quality gate template paths configured in `fest.yaml` when those templates live outside the canonical `gates/implementation/QUALITY_GATE_*` slots.

This made a customized research ritual look valid, but the gates were not visible to the CLI gate policy commands until the custom content was moved back into the canonical implementation template filenames.

## Environment

- `fest version dev (built 2026-05-12T22:47:17Z, commit 81f945a)`
- Repository/campaign context: ritual festival with custom non-code research gates

## What I configured

A ritual festival had enabled implementation gates in `fest.yaml` pointing at custom template paths, roughly:

```yaml
quality_gates:
  enabled: true
  auto_append: true
  implementation:
    - id: testing
      name: Artifact and Evidence Check
      template: gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK.md
      enabled: true
    - id: review
      name: Decision Review
      template: gates/problem-mining/QUALITY_GATE_DECISION_REVIEW.md
      enabled: true
    - id: iterate
      name: Evidence Gap Iteration
      template: gates/problem-mining/QUALITY_GATE_EVIDENCE_GAP_ITERATION.md
      enabled: true
    - id: fest-commit
      name: Commit Research Artifacts
      template: gates/problem-mining/QUALITY_GATE_COMMIT_RESEARCH_ARTIFACTS.md
      enabled: true
```

Those custom gate template files existed and `fest validate` still passed.

## Actual behavior

With the custom paths above, `fest gates show` printed the policy header but no active gates:

```text
GATE POLICY
...
Active Gates

Exclude Patterns
...
```

`fest gates apply --json` also produced skipped gate entries with empty template fields when gate files already existed:

```json
{
  "type": "skip",
  "path": ".../05_testing.md",
  "template": "",
  "reason": "gate_exists",
  "task_id": "testing"
}
```

It was not clear whether arbitrary custom `template:` paths are unsupported, unresolved, or simply hidden from `gates show` output.

## Workaround that did work

Moving the same custom gate content into the canonical implementation template slots made the CLI recognize the active gates:

- `gates/implementation/QUALITY_GATE_TESTING.md`
- `gates/implementation/QUALITY_GATE_REVIEW.md`
- `gates/implementation/QUALITY_GATE_ITERATE.md`
- `gates/implementation/QUALITY_GATE_FEST_COMMIT.md`

Then `fest.yaml` pointed back at those canonical template paths with custom names. After that:

```text
fest gates show

Active Gates

implementation
  testing
  review
  iterate
  fest-commit
```

And:

```text
fest gates remove --approve
Summary: 16 files removed

fest gates apply --approve
Summary: 16 files created, 0 skipped
```

## Expected behavior

Either:

1. Arbitrary `template:` paths in `quality_gates.implementation[]` should be resolved, shown, and applied correctly; or
2. If only canonical `gates/implementation/QUALITY_GATE_*` template slots are supported, `fest validate`, `fest gates show`, and/or `fest gates apply` should warn or fail when configured template paths are unsupported/unresolved.

In either case:

- `fest gates show` should not display an empty `Active Gates` section when `quality_gates.implementation` contains enabled entries.
- `fest gates apply --json` should include the resolved template path or a clear resolution error instead of `"template": ""`.
- Documentation should clarify whether custom gate behavior must be placed inside canonical template slots.

## Impact

Custom ritual authors can believe specialized gates are configured because `fest validate` passes and the template files exist, while the gate policy commands do not recognize them. For non-code rituals, this can cause agents to keep applying generic implementation gate semantics or leave the active gate behavior ambiguous.

lancekrogers:
## Summary

Fixes #168.

Two related bugs in `fest gates` that combined to make custom quality gate template paths appear broken when they actually worked.

### Bug 1: `fest gates show` silently dropped non-canonical buckets

`printGatesShowMergedTable` hardcoded `phaseOrder := []string{"implementation"}`, so gates whose template path bucketed under anything other than `implementation` (e.g. `gates/problem-mining/...`) were grouped correctly but never rendered. The table showed an empty "Active Gates" section even though the merger had loaded the gates.

JSON output (`--json`) was unaffected because it emits `merged.Gates` directly without grouping.

Fix: render canonical buckets first in declared order (`implementation`, `research`, `planning`, `review`, `non_coding_action`), then any remaining discovered buckets sorted alphabetically. `other` (the fallback for malformed template paths) sorts with the rest.

### Bug 2: `fest gates apply --json` skip results had empty `template` field

In `GenerateForSequence`, the `create` result correctly set `Template: gate.Template` but the `skip` result for `gate_exists` did not, producing JSON like:

```json
{"type": "skip", "path": ".../05_testing.md", "template": "", "reason": "gate_exists", "task_id": "testing"}
```

Fix: set `Template: gate.Template` on the skip result so it matches the create shape.

### Why the issue felt worse than the actual bug

The reporter believed custom `template:` paths in `quality_gates.implementation[]` were unsupported and worked around it by moving content into canonical `gates/implementation/QUALITY_GATE_*` slots. The underlying apply logic in `renderGateContent` actually resolves `gates/<phaseType>/<gateName>.md` against `festivalPath` correctly for custom paths, so apply did create the right files. The visible breakage was entirely `gates show` lying about what was active.

## Tests

- `TestOrderedPhaseBuckets`: canonical-first ordering, custom buckets sorted alphabetically after canonical, `other` bucket treated as custom.
- `TestExtractPhaseFromTemplate`: existing extraction behavior covered.
- `TestGenerateForSequence_SkipResultPreservesTemplate`: skip result for existing gate carries the configured template path.

## Out of scope (deferred)

Issue 168 also asks whether `fest validate` should warn when a configured gate template path does not resolve. That is a separate design question and is left for a follow-up.

## Test plan

- [x] `go build ./...`
- [x] `just test unit-raw` (all packages green)
- [x] `just lint vet`
- [x] `just lint gopls-tests internal/commands/gates` and `internal/gates`
- [ ] Manual smoke: ritual festival with custom `gates/problem-mining/...` paths now lists those gates under a `problem-mining` section in `fest gates show`.

### DIFF
--- a/internal/commands/gates/commands_test.go
+++ b/internal/commands/gates/commands_test.go
@@ -1,0 +1,94 @@
+package gates
+
+import (
+	"reflect"
+	"testing"
+
+	gatescore "github.com/Obedience-Corp/fest/internal/gates"
+)
+
+func TestOrderedPhaseBuckets(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		in   map[string][]gatescore.GateTask
+		want []string
+	}{
+		{
+			name: "only implementation",
+			in: map[string][]gatescore.GateTask{
+				"implementation": {{ID: "testing"}},
+			},
+			want: []string{"implementation"},
+		},
+		{
+			name: "custom bucket appears after canonical",
+			in: map[string][]gatescore.GateTask{
+				"implementation": {{ID: "testing"}},
+				"problem-mining": {{ID: "artifact-check"}},
+			},
+			want: []string{"implementation", "problem-mining"},
+		},
+		{
+			name: "multiple custom buckets sorted alphabetically",
+			in: map[string][]gatescore.GateTask{
+				"zeta":           {{ID: "z"}},
+				"problem-mining": {{ID: "artifact-check"}},
+				"alpha":          {{ID: "a"}},
+			},
+			want: []string{"alpha", "problem-mining", "zeta"},
+		},
+		{
+			name: "canonical buckets render in declared order",
+			in: map[string][]gatescore.GateTask{
+				"review":         {{ID: "r"}},
+				"implementation": {{ID: "i"}},
+				"research":       {{ID: "rs"}},
+			},
+			want: []string{"implementation", "research", "review"},
+		},
+		{
+			name: "other bucket is treated as custom",
+			in: map[string][]gatescore.GateTask{
+				"implementation": {{ID: "testing"}},
+				"other":          {{ID: "custom"}},
+			},
+			want: []string{"implementation", "other"},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got := orderedPhaseBuckets(tc.in)
+			if !reflect.DeepEqual(got, tc.want) {
+				t.Fatalf("orderedPhaseBuckets() = %v, want %v", got, tc.want)
+			}
+		})
+	}
+}
+
+func TestExtractPhaseFromTemplate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		template string
+		want     string
+	}{
+		{"gates/implementation/QUALITY_GATE_TESTING", "implementation"},
+		{"gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK", "problem-mining"},
+		{"gates/research/QUALITY_GATE_FOO", "research"},
+		{"agent/gates/implementation/testing", "other"},
+		{"", "other"},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.template, func(t *testing.T) {
+			t.Parallel()
+			if got := extractPhaseFromTemplate(tc.template); got != tc.want {
+				t.Fatalf("extractPhaseFromTemplate(%q) = %q, want %q", tc.template, got, tc.want)
+			}
+		})
+	}
+}
--- a/internal/commands/gates/commands_test.go
+++ b/internal/commands/gates/commands_test.go
@@ -78,8 +78,9 @@ func TestExtractPhaseFromTemplate(t *testing.T) {
 	}{
 		{"gates/implementation/QUALITY_GATE_TESTING", "implementation"},
 		{"gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK", "problem-mining"},
+		{"gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK.md", "problem-mining"},
 		{"gates/research/QUALITY_GATE_FOO", "research"},
-		{"agent/gates/implementation/testing", "other"},
+		{"agent/gates/implementation/testing", "implementation"},
 		{"", "other"},
 	}
 
--- a/internal/gates/task_generator_test.go
+++ b/internal/gates/task_generator_test.go
@@ -86,6 +86,95 @@ fest_status: pending
 	}
 }
 
+func TestGenerateForSequence_ResolvesTemplateWithMdSuffix(t *testing.T) {
+	t.Parallel()
+
+	festivalPath := t.TempDir()
+	sequencePath := filepath.Join(festivalPath, "001_IMPLEMENTATION", "01_core")
+	templateDir := filepath.Join(festivalPath, "gates", "problem-mining")
+	if err := os.MkdirAll(sequencePath, 0755); err != nil {
+		t.Fatalf("mkdir sequence: %v", err)
+	}
+	if err := os.MkdirAll(templateDir, 0755); err != nil {
+		t.Fatalf("mkdir template dir: %v", err)
+	}
+
+	templatePath := filepath.Join(templateDir, "QUALITY_GATE_ARTIFACT_CHECK.md")
+	templateContent := `---
+fest_type: gate
+fest_status: pending
+---
+# Gate: Artifact Check
+`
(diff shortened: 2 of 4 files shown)


internal/gates/task_generator_test.go
```diff
@@ -86,6 +86,45 @@ fest_status: pending
...
+fest_status: pending
+---
+# Gate: Testing
+`
+	if err := os.WriteFile(filepath.Join(sequencePath, "01_quality_gate_testing.md"), []byte(existing), 0644); err != nil {
+		t.Fatalf("write existing gate: %v", err)
+	}
+
+	generator := &TaskGenerator{}
+	gateTemplate := "gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK"
```
obey-agent, line 195: This test should cover the template shape from #168, including the `.md` suffix (`gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK.md`). `renderGateContent` currently takes the last segment from `extractPhaseAndGate` and appends `.md`, so a configured template with the suffix resolves as `QUALITY_GATE_ARTIFACT_CHECK.md.md` on create. That means this PR fixes the show/skip visibility path but still leaves the custom apply path from the issue broken. Please normalize optional suffixes, for example by trimming `.md`, and add this case.

internal/commands/gates/commands_test.go
```diff
@@ -0,0 +1,94 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		template string
+		want     string
+	}{
+		{"gates/implementation/QUALITY_GATE_TESTING", "implementation"},
+		{"gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK", "problem-mining"},
+		{"gates/research/QUALITY_GATE_FOO", "research"},
+		{"agent/gates/implementation/testing", "other"},
```
obey-agent: I would not codify `agent/gates/implementation/testing` as `other`: those are the built-in default implementation gates from `ImplementationGates()`. After this PR, `fest gates show` can render real implementation gates under an `other` bucket whenever built-ins remain in the merged policy. The display extractor should stay aligned with the generator parser, or share one parser, so `gates/...`, `agent/gates/...`, and any supported bare phase paths bucket consistently.

obey-agent (commented):
**Verdict: Request Changes**

This is pointed at the right failure mode: `gates show` should not hide enabled custom buckets, and preserving `template` on skip results is the right JSON contract. I left two inline comments because the PR still does not fully cover the custom template shapes from #168 and it codifies a parser mismatch that will make built-in implementation gates display under `other`.

**Key Findings**
- Custom gate templates with the `.md` suffix shown in #168 still resolve as `*.md.md` on create; normalize optional suffixes and add that regression case.
- `agent/gates/implementation/...` is currently asserted as `other`, but that is the built-in implementation-gate template prefix. The show bucket parser should align with the generator parser or share it.

**What's Done Well**
- The canonical-first plus sorted-extra bucket ordering is simple and deterministic.
- The skip-result JSON fix preserves the same template field shape as create results.
- The focused helper tests are easy to read and cover the intended ordering rules.

**Staff Standard**
Not yet. I would be comfortable with this after the parser behavior is made consistent and the `.md` suffix case from the reported issue is covered.

**Verification**
- Passed: `go test ./internal/commands/gates`
- Passed in read-only Docker checkout: `go test ./internal/commands/gates ./internal/gates`
- Broader read-only Docker `go test -short -timeout 30s ./...` did not complete cleanly because unrelated packages assume a writable/buildable checkout or non-root chmod semantics (`cmd/fest`, `internal/commands/workflow`, `internal/scaffold`).

obey-agent:
Pushed commit 8fba0b4 addressing both review findings.

1. `.md` suffix double-resolution: `ExtractPhaseAndGate` now strips a trailing `.md` from the gate name, so templates like `gates/problem-mining/QUALITY_GATE_ARTIFACT_CHECK.md` resolve to the correct file path instead of `.md.md`. Regression test: `TestGenerateForSequence_ResolvesTemplateWithMdSuffix`.

2. Parser mismatch: `extractPhaseFromTemplate` in commands.go now delegates to `gatescore.ExtractPhaseAndGate`, so built-in `agent/gates/implementation/...` templates bucket as `implementation` rather than `other`. The previous test that asserted `other` for that path was wrong and has been corrected.

`go build ./...`, `just test unit-raw`, `just lint vet`, and `just lint gopls-tests` on both gate packages are clean.

obey-agent (commented):
**Verdict: Approve**

The follow-up commit addresses the issues from the earlier review. The gate template parser is now shared between display bucketing and generation, built-in `agent/gates/...` templates bucket as implementation gates, and configured template paths with a trailing `.md` suffix no longer resolve to `*.md.md`.

**Key Findings**
- No remaining blocking findings in this diff.

**What's Done Well**
- The parser behavior is now centralized enough to keep `gates show` and `gates apply` aligned.
- The new tests cover the reported custom `gates/problem-mining/...md` shape, built-in `agent/gates/...` paths, and deterministic bucket ordering.
- The JSON skip-result fix preserves the same `template` field contract as create results.

**Staff Standard**
Yes. I would be comfortable submitting this change after the current focused verification.

**Verification**
- Passed in read-only Docker checkout: `go test ./internal/commands/gates ./internal/gates`
- Passed: `git diff --check origin/main...HEAD`

PR #181, opened 2026-05-21
Support standalone workflow show and watch

lancekrogers:
## Summary

- add top-level `fest show` support for standalone `WORKFLOW.md` directories
- add `fest watch` / `fest show --watch` support for standalone workflow progress rendering
- keep standalone watch event-driven with dynamic fsnotify path registration as `.workflow/` appears
- preserve festival selector, linked project, and picker behavior while allowing bare workflow directories to bypass workspace scope

## Validation

- `go test -short ./...`
- `just build test-binary`
- `go test -count=1 -v -tags=integration -run TestStandaloneWorkflow_TopLevelShowAndWatch ./tests/integration -timeout 5m`
- `just test integration-dev-watch`
- local smoke: `fest show --no-color` and `fest show --no-color --json` from a temp directory containing only `WORKFLOW.md`

### DIFF
--- a/internal/commands/show/show.go
+++ b/internal/commands/show/show.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/id"
 	"github.com/Obedience-Corp/fest/internal/pathutil"
+	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/workspace"
 	"github.com/spf13/cobra"
 )
@@ -46,6 +47,9 @@ SUBCOMMANDS:
   fest show all          List all festivals grouped by status
   fest show <name>       Show details of a specific festival by name
   fest show --festival <selector>  Show a festival by explicit selector (campaign workspace)`,
+		Annotations: map[string]string{
+			"scope": string(scope.Global),
+		},
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if opts.festival != "" && len(args) > 0 {
@@ -188,6 +192,11 @@ func runShowCurrent(ctx context.Context, opts *showOptions) error {
 	cwd, err := os.Getwd()
 	if err != nil {
 		return errors.IO("getting current directory", err)
+	}
+
+	handled, err := runShowStandaloneCurrent(ctx, cwd, opts)
+	if err != nil || handled {
+		return err
 	}
 
 	campaignRoot, _ := workspace.DetectCampaign(ctx, "")
--- a/internal/commands/show/standalone.go
+++ b/internal/commands/show/standalone.go
@@ -1,0 +1,376 @@
+package show
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/Obedience-Corp/fest/internal/commands/shared"
+	"github.com/Obedience-Corp/fest/internal/errors"
+	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
+	"github.com/Obedience-Corp/fest/internal/ui"
+	"github.com/Obedience-Corp/fest/internal/workflow/localstore"
+	"github.com/Obedience-Corp/fest/internal/workflow/standalone"
+)
+
+const (
+	standaloneModeAnonymous = "standalone-anonymous"
+	standaloneModeTracked   = "standalone-tracked"
+
+	renderModeNoRun      = "no_run"
+	renderModeNextUp     = "next_up"
+	renderModeInProgress = "in_progress"
+	renderModeComplete   = "complete"
+)
+
+// StandaloneWorkflowInfo is the display-ready view of a standalone WORKFLOW.md.
+type StandaloneWorkflowInfo struct {
+	Mode           string
+	StartDir       string
+	WorkflowDoc    string
+	RuntimeDir     string
+	RunID          string
+	RunStatus      string
+	CurrentStep    int
+	TotalSteps     int
+	CompletedSteps int
+	Blocked        bool
+	DocHashChanged bool
+	RenderMode     string
+	Steps          []shared.WorkflowStepView
+}
+
+// ResolveStandaloneWorkflow resolves and loads a standalone WORKFLOW.md view.
+// Festival contexts return nil so callers can continue with festival handling.
+func ResolveStandaloneWorkflow(ctx context.Context, startDir string) (*StandaloneWorkflowInfo, error) {
+	res, err := standalone.Resolve(ctx, startDir)
+	if err != nil {
+		return nil, err
+	}
+	switch res.Mode {
+	case standalone.ModeAnonymous, standalone.ModeTracked:
+		return loadStandaloneWorkflow(ctx, res)
+	default:
+		return nil, nil
+	}
+}
+
+func loadStandaloneWorkflow(ctx context.Context, res *standalone.Result) (*StandaloneWorkflowInfo, error) {
+	steps, err := wf.NewParser().Parse(ctx, res.WorkflowDoc)
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing WORKFLOW.md").WithField("path", res.WorkflowDoc)
+	}
+	if len(steps) == 0 {
+		return nil, errors.Validation("WORKFLOW.md has no parseable steps").
+			WithField("path", res.WorkflowDoc).
+			WithHint("add at least one '## Step 1: NAME' header")
+	}
+
+	mode := standaloneModeAnonymous
+	runStatus := "not-started"
+	var state *localstore.RunState
+	if res.Mode == standalone.ModeTracked {
+		mode = standaloneModeTracked
+		store := localstore.Open(res.RuntimeDir, res.WorkflowDoc)
+		var err error
+		state, err = store.LoadActive(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if state != nil {
+			runStatus = state.Status
+		}
+	}
+
+	views, current, completed, renderMode := buildStandaloneStepViews(steps, state)
+	info := &StandaloneWorkflowInfo{
+		Mode:           mode,
+		StartDir:       res.StartDir,
+		WorkflowDoc:    res.WorkflowDoc,
+		RuntimeDir:     res.RuntimeDir,
+		RunStatus:      runStatus,
+		CurrentStep:    current,
+		TotalSteps:     len(steps),
+		CompletedSteps: completed,
+		RenderMode:     renderMode,
+		Steps:          views,
+	}
+	if state != nil {
+		info.RunID = state.RunID
+		info.Blocked = state.Blocked
+		info.DocHashChanged = state.DocHashChanged
+	}
+	return info, nil
+}
+
+func buildStandaloneStepViews(steps []wf.WorkflowStep, state *localstore.RunState) ([]shared.WorkflowStepView, int, int, string) {
+	total := len(steps)
+	current, renderMode := standaloneRenderPosition(state, total)
+	completed := 0
+	if state != nil {
+		completed = clampStepCount(state.CompletedSteps, total)
+		if state.Status == "completed" {
+			completed = total
+		}
+	}
+
+	views := make([]shared.WorkflowStepView, len(steps))
+	for i, step := range steps {
+		position := i + 1
+		status := wf.StepStatusPending
+		if position <= completed || renderMode == renderModeComplete {
+			status = wf.StepStatusCompleted
+		} else if position == current {
+			switch {
+			case state != nil && state.Blocked:
+				status = wf.StepStatusBlocked
+			case renderMode == renderModeInProgress:
+				status = wf.StepStatusInProgress
+			default:
+				status = wf.StepStatusPending
+			}
+		}
+
+		views[i] = shared.WorkflowStepView{
+			Number:        step.Number,
+			Name:          step.Name,
+			Status:        status,
+			IsCurrent:     position == current && renderMode != renderModeComplete,
+			HasCheckpoint: step.HasCheckpoint(),
+			Goal:          step.Goal,
+		}
+	}
+	return views, current, completed, renderMode
+}
+
+func standaloneRenderPosition(state *localstore.RunState, totalSteps int) (int, string) {
+	if totalSteps == 0 {
+		return 0, renderModeNoRun
+	}
+	if state == nil {
+		return 1, renderModeNextUp
+	}
+	if state.Status == "completed" || state.CompletedSteps >= totalSteps {
+		return totalSteps, renderModeComplete
+	}
+	if state.CurrentStep > state.CompletedSteps {
+		return clampStepCount(state.CurrentStep, totalSteps), renderModeInProgress
+	}
+	next := state.CompletedSteps + 1
+	if next > totalSteps {
+		return totalSteps, renderModeComplete
+	}
+	if next < 1 {
+		return 1, renderModeNextUp
+	}
+	return next, renderModeNextUp
+}
+
+func clampStepCount(n, total int) int {
+	if n < 0 {
+		return 0
+	}
+	if total > 0 && n > total {
+		return total
+	}
+	return n
+}
+
+func emitStandaloneWorkflow(info *StandaloneWorkflowInfo, opts *showOptions) error {
+	renderOpts := standaloneRenderOptionsFromShow(opts)
+	if opts != nil && opts.json {
+		return emitStandaloneWorkflowJSON(info)
+	}
+	if renderOpts.Summary {
+		fmt.Print(FormatStandaloneWorkflowSummary(info))
+		return nil
+	}
+	fmt.Print(formatStandaloneWorkflowProgress(info, renderOpts))
+	return nil
+}
+
+func runShowStandaloneCurrent(ctx context.Context, cwd string, opts *showOptions) (bool, error) {
+	standaloneWorkflow, err := ResolveStandaloneWorkflow(ctx, cwd)
+	if err != nil {
+		return false, err
+	}
+	if standaloneWorkflow == nil {
+		return false, nil
+	}
+	if opts != nil && opts.watch {
+		return true, WatchStandaloneWorkflow(ctx, standaloneWorkflow, WatchOptions{
+			Summary:    opts.summary,
+			Goals:      opts.goals,
+			Collapsed:  opts.collapsed,
+			InProgress: opts.inProgress,
+		})
+	}
+	return true, emitStandaloneWorkflow(standaloneWorkflow, opts)
+}
+
+type standaloneRenderOptions struct {
+	Summary    bool
+	ShowGoals  bool
+	Collapsed  bool
+	InProgress bool
+}
+
+func standaloneRenderOptionsFromShow(opts *showOptions) standaloneRenderOptions {
+	if opts == nil {
+		return standaloneRenderOptions{}
+	}
+	return standaloneRenderOptions{
+		Summary:    opts.summary,
+		ShowGoals:  opts.goals,
+		Collapsed:  opts.collapsed,
+		InProgress: opts.inProgress,
+	}
+}
+
+func standaloneRenderOptionsFromWatch(opts WatchOptions) standaloneRenderOptions {
+	return standaloneRenderOptions{
+		Summary:    opts.Summary,
+		ShowGoals:  opts.Goals,
+		Collapsed:  opts.Collapsed,
+		InProgress: opts.InProgress,
+	}
+}
+
+// FormatStandaloneWorkflowSummary renders aggregate standalone workflow state.
+func FormatStandaloneWorkflowSummary(info *StandaloneWorkflowInfo) string {
+	var sb strings.Builder
+	writeStandaloneHeader(&sb, info)
+	return sb.String()
+}
+
+// FormatStandaloneWorkflowProgress renders standalone workflow state plus steps.
+func FormatStandaloneWorkflowProgress(info *StandaloneWorkflowInfo) string {
+	return formatStandaloneWorkflowProgress(info, standaloneRenderOptions{})
+}
+
+func formatStandaloneWorkflowProgress(info *StandaloneWorkflowInfo, opts standaloneRenderOptions) string {
+	var sb strings.Builder
+	writeStandaloneHeader(&sb, info)
+	current := currentStandaloneStep(info)
+	if opts.Collapsed {
+		if current != nil {
+			fmt.Fprintf(&sb, "Current: Step %d - %s\n", current.Number, current.Name)
+		}
+		return sb.String()
+	}
+
+	sb.WriteString("\n")
+	sb.WriteString(renderStandaloneSteps(info.Steps, opts))
+	if current != nil {
+		fmt.Fprintf(&sb, "\nCurrent: Step %d - %s\n", current.Number, current.Name)
+	}
+	if info.DocHashChanged {
+		fmt.Fprintf(&sb, "%s WORKFLOW.md has changed since this run started\n", ui.Warning("!"))
+	}
+	if info.RenderMode != renderModeComplete {
+		if info.Mode == standaloneModeAnonymous {
+			sb.WriteString("First mutating command: ")
+		} else {
+			sb.WriteString("Next: ")
+		}
+		sb.WriteString(ui.Accent("fest workflow advance"))
+		sb.WriteString("\n")
+	}
+	return sb.String()
+}
+
+func renderStandaloneSteps(steps []shared.WorkflowStepView, opts standaloneRenderOptions) string {
+	filtered := steps
+	if opts.InProgress {
+		filtered = make([]shared.WorkflowStepView, 0, len(steps))
+		for _, step := range steps {
+			if step.Status == wf.StepStatusCompleted || step.Status == wf.StepStatusSkipped {
+				continue
+			}
+			filtered = append(filtered, step)
+		}
+	}
+	return shared.RenderWorkflowSteps(filtered, !opts.ShowGoals)
+}
+
+func writeStandaloneHeader(sb *strings.Builder, info *StandaloneWorkflowInfo) {
+	sb.WriteString("Workflow Progress\n")
+	sb.WriteString("-----------------\n")
+	fmt.Fprintf(sb, "Workflow: %s\n", info.WorkflowDoc)
+	fmt.Fprintf(sb, "Mode: %s\n", strings.TrimPrefix(info.Mode, "standalone-"))
+	if info.RunID != "" {
+		fmt.Fprintf(sb, "Run: %s (%s)\n", info.RunID, info.RunStatus)
+	} else {
+		fmt.Fprintf(sb, "Status: %s\n", info.RunStatus)
+	}
+	if info.Blocked {
+		fmt.Fprintf(sb, "Blocked: true\n")
+	}
+	fmt.Fprintf(sb, "Progress: %s steps\n", shared.RenderWorkflowProgress(info.CompletedSteps, info.TotalSteps))
+}
+
+func currentStandaloneStep(info *StandaloneWorkflowInfo) *shared.WorkflowStepView {
+	if info == nil || info.RenderMode == renderModeComplete {
+		return nil
+	}
+	for i := range info.Steps {
+		if info.Steps[i].IsCurrent {
+			return &info.Steps[i]
+		}
+	}
+	return nil
+}
+
+type standaloneWorkflowJSON struct {
+	Mode           string                   `json:"mode"`
+	WorkflowDoc    string                   `json:"workflow_doc"`
+	RuntimeDir     string                   `json:"runtime_dir,omitempty"`
+	RunID          string                   `json:"run_id,omitempty"`
+	RunStatus      string                   `json:"run_status"`
+	CurrentStep    int                      `json:"current_step"`
+	TotalSteps     int                      `json:"total_steps"`
+	CompletedSteps int                      `json:"completed_steps"`
+	Blocked        bool                     `json:"blocked,omitempty"`
+	DocHashChanged bool                     `json:"doc_hash_changed,omitempty"`
+	Steps          []standaloneWorkflowStep `json:"steps"`
+}
+
+type standaloneWorkflowStep struct {
+	Number        int    `json:"number"`
+	Name          string `json:"name"`
+	Status        string `json:"status"`
+	Current       bool   `json:"current,omitempty"`
+	HasCheckpoint bool   `json:"has_checkpoint,omitempty"`
+	Goal          string `json:"goal,omitempty"`
+}
+
+func emitStandaloneWorkflowJSON(info *StandaloneWorkflowInfo) error {
+	out := standaloneWorkflowJSON{
+		Mode:           info.Mode,
+		WorkflowDoc:    info.WorkflowDoc,
+		RuntimeDir:     info.RuntimeDir,
+		RunID:          info.RunID,
+		RunStatus:      info.RunStatus,
+		CurrentStep:    info.CurrentStep,
+		TotalSteps:     info.TotalSteps,
+		CompletedSteps: info.CompletedSteps,
+		Blocked:        info.Blocked,
+		DocHashChanged: info.DocHashChanged,
+		Steps:          make([]standaloneWorkflowStep, 0, len(info.Steps)),
+	}
+	for _, step := range info.Steps {
+		out.Steps = append(out.Steps, standaloneWorkflowStep{
+			Number:        step.Number,
+			Name:          step.Name,
+			Status:        string(step.Status),
+			Current:       step.IsCurrent,
+			HasCheckpoint: step.HasCheckpoint,
+			Goal:          step.Goal,
+		})
+	}
+	if err := shared.EncodeJSON(os.Stdout, out); err != nil {
+		return errors.Wrap(err, "encoding JSON output")
+	}
+	return nil
+}
--- a/internal/commands/show/watch.go
+++ b/internal/commands/show/watch.go
@@ -70,6 +70,9 @@ func runWatchMode(ctx context.Context, festival *FestivalInfo, opts *showOptions
 	w, err := watch.New(watch.Config{
 		Paths:    watchPaths,
 		Debounce: 100 * time.Millisecond,
+		OnError: func(err error) {
+			fmt.Fprintf(os.Stderr, "%s file watch error: %v\n", ui.Warning("Warning:"), err)
+		},
 	}, func() {
 		clearScreen()
 		// Refresh festival info to get latest data
@@ -79,7 +82,7 @@ func runWatchMode(ctx context.Context, festival *FestivalInfo, opts *showOptions
 		}
 		if err := renderFestivalView(ctx, festival, opts); err != nil {
 			// Log error but don't fail - just show stale data
-			_ = err
+			fmt.Fprintf(os.Stderr, "%s could not refresh festival view: %v\n", ui.Warning("Warning:"), err)
 		}
 		printWatchFooter(false)
 	})
--- a/internal/commands/progress/watch.go
+++ b/internal/commands/progress/watch.go
@@ -45,6 +45,9 @@ func runWatchMode(ctx context.Context, mgr *progress.Manager, loc *show.Location
 	w, err := watch.New(watch.Config{
 		Paths:    watchPaths,
 		Debounce: 100 * time.Millisecond,
+		OnError: func(err error) {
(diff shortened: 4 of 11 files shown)


internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+			timer = nil
+			timerC = nil
+			if err := renderAndWatch(); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+func renderStandaloneWorkflowFromStartDir(ctx context.Context, startDir string, opts WatchOptions) error {
```
obey-agent: **Dead code.** `renderStandaloneWorkflowFromStartDir` is defined but never referenced anywhere in the repo (verified via grep). Either wire it into a code path that needs it or delete it. Either way, do not ship it as-is — dead code rots and confuses future maintainers.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+		return err
+	}
+	if info == nil {
+		return errors.NotFound("standalone workflow").WithField("path", startDir)
+	}
+	return renderStandaloneWorkflowView(info, opts)
+}
+
+func renderStandaloneWorkflowView(info *StandaloneWorkflowInfo, opts WatchOptions) error {
+	showOpts := &showOptions{summary: opts.Summary}
```
obey-agent: **Watch flags silently ignored for standalone workflows.** `renderStandaloneWorkflowView` builds `showOptions{summary: opts.Summary}` and discards `Goals`, `Collapsed`, and `InProgress`. Festival watch honors all four (see `WatchFestival` in `watch.go`). A user running `fest watch --goals --collapsed` against a standalone WORKFLOW.md gets no error and no behavior change, which is exactly the kind of silent inconsistency that erodes trust in the CLI.

Either map the flags through to the standalone renderer (even if the current renderer doesn't yet differentiate, the showOptions should at least be populated so future renderer work can read them), or explicitly reject the flags with a clear error when paired with a standalone context.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Remove) || event.Has(fsnotify.Rename) {
+				scheduleRender()
+			}
+
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return nil
+			}
+			_ = err
```
obey-agent: **`_ = err` silently drops fsnotify errors.** When the OS reports an inotify/kqueue error (queue overflow, descriptor exhaustion, permission revoked mid-watch), the watcher keeps spinning and the user sees stale output with no diagnostic. Per the staff-level question about "our observability system remaining useful" — this is exactly where a transient watcher error needs to surface, either as a one-line stderr write under verbose mode or as a tracked counter. `_ = err` is also a code-smell pattern reviewers will flag.

At minimum, write the error to stderr when verbose, and consider emitting a warning footer alongside `printWatchFooter` when fsnotify is degraded.

internal/commands/show/standalone.go
obey-agent: **Dead code flagged by lint.** `renderStandaloneWorkflowFromStartDir` at line 382 is defined but never called anywhere in the codebase, and `just lint all` (golangci-lint with the project's `unused` linter) fails specifically on this function. Looking at the diff: the previous commit (`a31785`) called this helper from the polling-fallback path in `runStandalonePollingMode`, but the latest commit (`37b5f0d`) inlined the fsnotify watch loop and removed the polling fallback, leaving this helper orphaned. Either delete `renderStandaloneWorkflowFromStartDir` (the call site is gone), or fold its contents into the one place that would still benefit from a resolve-and-render helper. Lint must pass clean — the project's golangci-lint config explicitly catches unused functions, and this is the only violation in the entire codebase right now.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+
+func addStandaloneWatchPaths(watcher *fsnotify.Watcher, watched map[string]bool, info *StandaloneWorkflowInfo) {
+	add := func(path string) {
+		if path == "" || watched[path] {
+			return
+		}
+		if stat, err := os.Stat(path); err == nil && (stat.IsDir() || filepath.Base(path) == "WORKFLOW.md") {
+			if err := watcher.Add(path); err == nil {
+				watched[path] = true
+			}
```
obey-agent: **`watcher.Add` errors silently dropped during dynamic registration.** `addStandaloneWatchPaths` checks `err == nil` and updates the `watched` map only on success, but the failure case produces no diagnostic and no retry. If the `.workflow/` directory appears mid-session but `Add` fails (e.g. due to inotify limits), the user gets a watch session that never refreshes for that path and no signal that anything went wrong.

This is the second silent-error pattern in this file. Combined with line 370, the watch loop loses the ability to tell a user "the OS is no longer notifying me about your runtime directory." Track failures so they can be surfaced in the footer or via verbose logging.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Reviewed head `37b5f0d8a185e92717f4cfcb6f160333a46e544d` (head shifted from `a31785` to `37b5f0d` mid-review when an fsnotify-inline refactor was pushed; re-fetched and re-validated against the new head before posting). The PR extends `fest show` and `fest watch` to support standalone `WORKFLOW.md` directories outside a campaign workspace, with full unit and integration test coverage. The architecture is sound: standalone resolution runs ahead of festival resolution, returns nil to indicate "not standalone, fall through," and the watch path inlines fsnotify with a 100ms debounce. One blocking finding from lint.

## Key Findings
- **Dead code: `renderStandaloneWorkflowFromStartDir` is unused.** `just lint all` fails on `internal/commands/show/standalone.go:382` — the function is defined but has no call sites. The mid-review refactor commit (`37b5f0d`) replaced the polling-fallback path with an inline fsnotify loop and removed `runStandalonePollingMode`, which was the only caller. The helper was left behind. Delete it (or inline it if there's a real reason to keep the resolve-then-render seam). Inline detail at `standalone.go:382`.

## What's Done Well
- Scope handling is right: both `fest show` and `fest watch` were promoted from `scope.Workspace` to `scope.Global` so they can run outside a campaign workspace, which is exactly what the standalone WORKFLOW.md feature needs.
- Standalone resolution precedes festival resolution at every entry point (`runShowCurrent` in `show.go`, `runWatch` in `watch/command.go`). The `ResolveStandaloneWorkflow` contract returning `nil, nil` for non-standalone contexts is a clean "fall through to existing behavior" signal that doesn't require touching every downstream branch.
- Dependency injection in `commandDeps` is extended consistently: `resolveStandalone` and `watchStandalone` are added alongside the existing `resolve` and `watch` fields, with `defaultCommandDeps` wiring up real implementations. The new tests `TestWatchCommandStandaloneWorkflowDelegatesBeforeFestivalResolution` and `TestWatchCommandSelectorSkipsStandaloneResolution` pin both the "use standalone when no selector" and "explicit selector skips standalone resolver" paths via mock delegates that fail the test if called wrongly.
- Three unit tests (`standalone_test.go`) cover the rendering math: anonymous-mode first-step-current, tracked-mode in-progress step marking, and the formatted-output happy path. The integration test suite adds `TestStandaloneWorkflow_TopLevelShowAndWatch` with four subtests covering anonymous show, tracked show, watch initial render, and watch-refresh-after-bootstrap.
- The watch-refresh-after-bootstrap test (`runStandaloneFestWatchWithAdvance`) is the harder one to get right: it launches `fest watch` in the foreground and `fest workflow advance` in the background after a 500ms delay, then asserts the watch output picks up the bootstrap and re-renders with `Mode: tracked` and `Progress: 1/2`. That's a real end-to-end check of fsnotify event handling, not just initial render.
- `addStandaloneWatchPaths` is correctly idempotent (the `watched` map prevents re-adding), filters to dirs and `WORKFLOW.md` files only, and silently skips paths that don't exist yet — appropriate for a watch loop where `.workflow/runs/<run-id>/` may not exist on first render.

## Validation
- `just test all` on `37b5f0d`: 2087 unit tests + 57 integration tests passed.
- `go build ./...` clean.
- `just lint all`: 1 unused-function violation (the blocking finding above).
- Re-fetched and re-diffed against the mid-review head shift; verified the refactor from `filewatch.New` (polling-fallback abstraction) to inline `fsnotify.NewWatcher` is internally consistent — the only loose end is the orphaned helper.

## Staff Standard
Not yet. The feature, the tests, and the scope changes are well thought out. But shipping with a known lint failure (the only one in the entire repo right now) is below the bar; the project's golangci-lint config exists specifically to catch this class of issue. Delete the dead helper and this is a clean Approve.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+
+		case <-timerC:
+			timer = nil
+			timerC = nil
+			if err := renderAndWatch(); err != nil {
+				return err
+			}
+		}
+	}
+}
```
lancekrogers: **`runStandaloneWatchMode` is 71 lines, over the 50-line target stated in the project's Go standards.** The function does five distinct things: watcher setup, render closure construction, debounce timer construction, the select loop, and post-cancellation cleanup. Extract `scheduleRender` and the select-loop body into named helpers, or factor the render closure into a method on a small `standaloneWatchSession` struct. The current shape is easy enough to follow now but will rot the first time someone needs to add a polling fallback or a different event filter.

Also: a transient error from `renderAndWatch` inside the `<-timerC` branch terminates the whole watch session (line 375-377). Festival watch tolerates render errors. This means a momentary fs hiccup ends the user's watch session. Catch the error, surface it in the footer, and continue.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,419 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing WORKFLOW.md").WithField("path", res.WorkflowDoc)
+	}
+	if len(steps) == 0 {
+		return nil, errors.Validation("WORKFLOW.md has no parseable steps").
+			WithField("path", res.WorkflowDoc).
+			WithHint("add at least one '## Step 1: NAME' header")
+	}
+
+	mode := "standalone-anonymous"
```
lancekrogers: **Magic strings for mode and renderMode.** Five distinct string literals are scattered across this file (`standalone-anonymous`, `standalone-tracked`, `no_run`, `next_up`, `in_progress`, `complete`) and at least two appear in conditionals (`if info.Mode == "standalone-anonymous"` line 205, `if info.RenderMode != "complete"` line 204, `renderMode == "in_progress"` line 121). One typo and the user sees the wrong footer with no compile-time signal.

Define typed constants (`type StandaloneMode string`, `type RenderMode string`) and have the test surface compare against them. The JSON marshaling stays string-valued; only the in-process comparisons get safer.

internal/commands/show/show.go
```diff
@@ -190,6 +194,22 @@ func runShowCurrent(ctx context.Context, opts *showOptions) error {
...
+		if opts.watch {
+			return WatchStandaloneWorkflow(ctx, standaloneWorkflow, WatchOptions{
+				Summary:    opts.summary,
+				Goals:      opts.goals,
+				Collapsed:  opts.collapsed,
+				InProgress: opts.inProgress,
+			})
+		}
+		return emitStandaloneWorkflow(standaloneWorkflow, opts)
+	}
```
lancekrogers: **Consider extracting the standalone branch.** This block is the only chunk of new logic in `show.go` and `runShowCurrent` is now 50 lines (right at the project's function-size limit). Moving the standalone routing into `standalone.go` as e.g. `HandleStandaloneIfPresent(ctx, opts) (handled bool, err error)` keeps the standalone wiring co-located with its other functions and pulls `show.go` back under its previous size. The file crossed 500 lines (from 483 to 503) with this PR — a small extraction reverses that without touching behavior.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Reviewed head `37b5f0d8a185e92717f4cfcb6f160333a46e544d` under the staff-level standard. The PR adds standalone WORKFLOW.md support to `fest show` and `fest watch`, with integration tests that exercise both anonymous and tracked modes and verify the event-driven refresh path. The intent and the test surface are right. Implementation has several real issues that need to land before this is ready to merge.

## Key Findings
- **Dead code shipped.** `renderStandaloneWorkflowFromStartDir` (standalone.go:382) is defined and never called. See inline.
- **Watch flags silently ignored for standalone workflows.** `renderStandaloneWorkflowView` discards `Goals`, `Collapsed`, and `InProgress` from the public `WatchOptions` struct (standalone.go:393-396). `fest watch --goals --collapsed` is inert for standalone but works for festivals. Same flag, two behaviors, no error. See inline.
- **Silent error swallowing in the watch loop.** `_ = err` in the `watcher.Errors` case (standalone.go:370) and discarded `watcher.Add` failures (standalone.go:404) mean the user has no signal when fsnotify degrades mid-session. The PR claims event-driven dynamic registration; if registration silently fails, the user sees a stale view with no diagnostic. See inlines.
- **`runStandaloneWatchMode` is 71 lines** (over the project's 50-line target). The function bundles watcher setup, render closure, debounce, select loop, and cleanup. Worth extracting before someone needs to add a polling fallback. A transient render error inside the timer branch (standalone.go:375-377) also terminates the entire watch session — festival watch tolerates these. See inline.
- **Magic strings for `Mode` and `RenderMode`.** Six distinct string literals (`standalone-anonymous`, `standalone-tracked`, `no_run`, `next_up`, `in_progress`, `complete`) compared in conditionals across the file. One typo flips a code path with no compile-time signal. See inline.
- **`show.go` crossed the 500-line threshold** (483 → 503). The standalone branch in `runShowCurrent` is the natural extraction point. Not a blocker on its own but worth doing while the change is fresh. See inline.

## What's Done Well
- The integration test `WatchAnonymousWorkflowRefreshesAfterBootstrap` correctly exercises the dynamic-registration path with a background `fest workflow advance`, asserting the watch session refreshes after `.workflow/` appears mid-session. This is exactly the test the implementation needs to pin its core claim.
- Unit tests around `buildStandaloneStepViews` cover both anonymous (state == nil) and tracked-in-progress cases, asserting the renderMode/current/completed triplet directly. Pure unit tests against a small focused function — right level.
- Watch command uses dependency injection (`commandDeps`) with separate `resolve`/`resolveStandalone`/`watch`/`watchStandalone` knobs. Tests can verify the routing decisions (standalone wins when selector empty; selector skips standalone) without touching the filesystem. Good shape.
- Errors throughout `loadStandaloneWorkflow` use the project `errors` package with `WithField`/`WithHint` — proper observability surface for the cases where it matters most.
- Context is plumbed through every entry point and forwarded into `standalone.Resolve` / `store.LoadActive` / `wf.NewParser().Parse`. Cancellation will propagate correctly from the CLI surface down to the parser.
- `WatchFestival(nil, ...)` and `WatchStandaloneWorkflow(nil, ...)` both validate the nil case with a typed `errors.Validation`. Defensive in the right places.
- The container-based integration tests respect the filesystem-safety rule documented in CLAUDE.md — host-side unit tests stay pure logic, mutation tests live in the container harness.

## Staff Standard
**Would I be proud to submit this as my own work?** Not yet. The functional surface is correct and the tests are good, but the silent error swallowing on the watch loop is the kind of detail a staff-level review will flag and a future on-call incident will surface. The dead-code shipping and the watch-flag inconsistency are also things I'd want fixed before a maintenance engineer with less context inherits this file. Address the inlines and this becomes a clean +1.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,517 @@
...
+		startDir: startDir,
+		opts:     opts,
+		watcher:  watcher,
+		watched:  map[string]bool{},
+	}
+	if err := session.render(ctx); err != nil {
+		return err
+	}
+	return session.loop(ctx)
+}
```
obey-agent: **Parallel implementation of fsnotify watch loop.**

`internal/watch.Watcher` (used by `runWatchMode` in `watch.go`) already implements fsnotify init, mutex-guarded debouncing, and event filtering. `runStandaloneWatchMode` re-implements all three. The PR body justifies this with "dynamic fsnotify path registration as `.workflow/` appears" — but the right fix is to extend `internal/watch` (a method to add paths after construction, and/or tolerating non-existent paths at `New` time), not to fork the loop.

Two paths to fix:
- Add `(*Watcher).AddPath(string) error` and let `watch.New` accept an empty initial path set, then drive a single watcher implementation from both call sites.
- Or move the per-render path-registration logic into the existing watcher via an `OnRefresh` hook.

As-is, any future fix to debouncing, event filtering, error handling, or polling fallback has to be made in two places. That divergence is already visible in this PR: festival watch falls back to polling on fsnotify init failure (`watch.go:87-90`) and standalone watch returns the raw error (see the comment on line 396).

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,517 @@
...
+	if startDir == "" {
+		startDir = filepath.Dir(info.WorkflowDoc)
+	}
+	return runStandaloneWatchMode(ctx, startDir, opts)
+}
+
+func runStandaloneWatchMode(ctx context.Context, startDir string, opts WatchOptions) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
```
obey-agent: **Raw fsnotify error returned without project error wrapping, and no polling fallback.**

Two issues on this line:

1. `return err` propagates the raw `fsnotify.NewWatcher()` error. The project standard is to wrap with the `errors` package (e.g. `errors.IO("creating fsnotify watcher", err)` or `errors.Wrap(err, "...")`). Other failures in this same file correctly use `errors.Wrap`/`errors.Validation`/`errors.NotFound` — this one slipped through.

2. `runWatchMode` (festival path, watch.go:87-90) falls back to polling when fsnotify init fails:
```go
if err != nil {
    fmt.Fprintf(os.Stderr, "File watching unavailable (%v), using polling fallback\n", err)
    return runPollingMode(ctx, festival, opts)
}
```
The standalone path has no equivalent — users on systems where fsnotify is unavailable get a hard error instead of a working polling experience. This is exactly the divergence the parallel-implementation comment above warns about.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,517 @@
...
+			return
+		}
+		if !stat.IsDir() && filepath.Base(path) != "WORKFLOW.md" {
+			return
+		}
+		if err := s.watcher.Add(path); err != nil {
+			fmt.Fprintf(os.Stderr, "%s could not watch %s: %v\n", ui.Warning("Warning:"), path, err)
+			return
+		}
+		s.watched[path] = true
```
obey-agent: **`watched` map persists across remove/recreate cycles, defeating dynamic re-registration.**

Once a path is added it's permanently marked `watched[path] = true`. If the watched directory (or `.workflow/`, or `runs/<id>/`) is removed and recreated, the stale map entry causes the next `addWatchPaths` call to skip re-registering it. fsnotify drops the underlying watch when an inode is removed, so after a remove/recreate cycle, events from the new inode are silently lost and the view stops refreshing.

In practice this matters for `fest workflow abandon` → `fest workflow start` cycles and any tooling that rewrites `.workflow/` atomically (write-temp + rename, which is a common pattern for crash-safe writers).

Either drop the cache entry on remove events (track the fsnotify Remove/Rename and clear `s.watched[path]`) or re-stat on every call and re-Add when the inode is gone.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,517 @@
+package show
```
obey-agent, line 1: **File is 517 lines — over the 500-line project guideline, and there's a natural seam to split on.**

Suggest extracting `runStandaloneWatchMode` + `standaloneWatchSession` + `shouldRefreshStandaloneWatch` + `addWatchPaths` into `standalone_watch.go`. That leaves `standalone.go` focused on resolve/load/render and isolates the watch loop (which is also the surface most likely to change if it gets refactored onto `internal/watch` per the comment above).

internal/commands/show/watch.go
```diff
@@ -79,7 +79,7 @@ func runWatchMode(ctx context.Context, festival *FestivalInfo, opts *showOptions
 		}
 		if err := renderFestivalView(ctx, festival, opts); err != nil {
 			// Log error but don't fail - just show stale data
-			_ = err
+			fmt.Fprintf(os.Stderr, "%s could not refresh festival view: %v\n", ui.Warning("Warning:"), err)
```
obey-agent, line 85: **Nice observability fix.** Replacing the silent `_ = err` with a visible warning is exactly the kind of cleanup that pays off the next time someone has to diagnose a stale watch view in the wild. Keep this pattern when you rationalize the standalone watch loop onto `internal/watch` — it should bubble watcher errors with the same warning prefix.

internal/commands/show/standalone.go
```diff
@@ -0,0 +1,517 @@
...
+
+func runShowStandaloneCurrent(ctx context.Context, cwd string, opts *showOptions) (bool, error) {
+	standaloneWorkflow, err := ResolveStandaloneWorkflow(ctx, cwd)
+	if err != nil {
+		return false, err
+	}
+	if standaloneWorkflow == nil {
+		return false, nil
+	}
+	if opts.watch {
```
obey-agent: **`opts.watch` is dereferenced without a nil guard, but `emitStandaloneWorkflow` two lines down guards `opts != nil`.**

Pick one contract: either `runShowStandaloneCurrent` requires non-nil `opts` (and the downstream nil-guards in `standaloneRenderOptionsFromShow` / `emitStandaloneWorkflow` should be removed for consistency), or this branch needs a nil check too. The current caller (`runShowCurrent` in show.go) always passes non-nil, so this isn't a live bug — but defensive checks that exist in some branches and not others are exactly how live bugs get introduced later by someone refactoring the call site.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The feature itself is the right shape — standalone WORKFLOW.md needs first-class `show`/`watch` support, the resolver gating is clean, and the watch-command DI surface (`resolveStandalone` + `watchStandalone`) makes the new path testable. But the standalone watch implementation forks fsnotify handling that already lives in `internal/watch`, and the divergence is already producing observable behavior gaps. That needs to be unified before merge.

### Key Findings
- **Parallel fsnotify implementation.** `runStandaloneWatchMode` re-implements the loop, debounce, and event filter that `internal/watch.Watcher` already provides. Inline comment on `standalone.go:410`.
- **Raw error + no polling fallback.** `fsnotify.NewWatcher()` failure returns an unwrapped error; festival watch falls back to polling, standalone watch does not. Inline comment on `standalone.go:396`.
- **Stale `watched` map.** Permanent map entries break re-registration after remove/recreate cycles (`workflow abandon`+`workflow start`, atomic rewrites). Inline comment on `standalone.go:505`.
- **File over 500-line guideline.** `standalone.go` is 517 lines with an obvious split (watch loop into `standalone_watch.go`). Inline comment on file header.
- **Inconsistent nil-guarding** between `runShowStandaloneCurrent` and its callees. Inline comment on `standalone.go:204`.

### What's Done Well
- Proper use of the project `errors` package almost everywhere (`Wrap`, `Validation`, `NotFound`, `IO`) with `WithField`/`WithHint` context.
- Context propagation is correct throughout: `ctx` as first param, `ctx.Err()` checked in resolver, `ctx.Done()` honored in the watch loop, no leaked goroutines.
- Replacing the silent `_ = err` in `watch.go` with a real warning is a small but meaningful observability win.
- DI surface for the watch command is clean and the unit tests exercise both the standalone-takes-precedence and selector-skips-standalone paths.
- Integration tests cover the bootstrap case (anonymous → tracked while watching) end-to-end in the container.

### Staff Standard
Would I be proud to submit this as my own work? Not yet. The product behavior is right and the seams are in the right places, but knowingly forking a shared watcher to work around a constructor constraint — rather than extending the shared watcher — is the kind of decision a staff review should catch every time. The fact that festival watch already has polling fallback and standalone watch silently doesn't is the cost of that decision showing up in the same diff. Address the parallel-implementation issue (extend `internal/watch` to accept dynamic paths, then route both call sites through it), wrap the fsnotify error, and fix the stale-watched map, and this is a clean merge.

obey-agent:
Apologies for the noise above — multiple duplicate review summaries posted during a tooling retry. The canonical review summary is review 4342516821 (or any of the duplicates from 04:45-04:46 UTC; they all have identical content). Inline comments are unique and posted once each.

obey-agent (commented):
**Verdict:** Approve

**Overview:** This PR adds standalone `WORKFLOW.md` support to top-level `fest show` and `fest watch` without forking the festival watch path unnecessarily. The standalone resolver remains scoped, festival context still wins, and the shared watcher now has a small dynamic path API that supports the `.workflow/` bootstrap path.

**Key Findings:** No blocking or follow-up findings. I did not find correctness, error-handling, watch-lifecycle, or maintainability issues worth inline comments.

**What's Done Well:** The implementation keeps display-state construction testable, propagates parse/runtime errors instead of silently degrading, adds non-fatal watcher error reporting, and covers anonymous-to-tracked watch refresh behavior in the container harness.

**Validation:**
- `go test ./internal/commands/show ./internal/commands/watch ./internal/watch`
- `go test -short ./...`
- `just build test-binary`
- `go test -count=1 -v -tags=integration -run TestStandaloneWorkflow_TopLevelShowAndWatch ./tests/integration -timeout 5m`

**Staff Standard:** Yes. I would be comfortable submitting this as my own production change.

PR #182, opened 2026-05-22
Start standalone workflow runs on create

lancekrogers:
## Summary
- start a tracked standalone workflow run when `fest create workflow` initializes `.workflow/`
- keep `fest next` working immediately after the default create workflow path
- reject pre-existing `.workflow/` before creating runtime state so rollback cannot delete files owned by a prior run or partial setup
- update JSON suggestions and integration coverage for the next-ready flow

## Validation
- `go test ./internal/commands/festival ./internal/commands/next ./internal/commands/workflow ./internal/workflow/localstore`
- `go test -short ./...`
- `just build test-binary`
- `go test -tags integration ./tests/integration -run TestCreateWorkflowStandalone -count=1`

### DIFF
--- a/internal/commands/festival/create_workflow_standalone.go
+++ b/internal/commands/festival/create_workflow_standalone.go
@@ -18,8 +18,8 @@ import (
 
 // runStandaloneCreateWorkflow implements `fest create workflow <name>` outside
 // a festival context (D009). It writes <cwd>/WORKFLOW.md from the same
-// WorkflowInput shape festival mode uses, then optionally initializes
-// <cwd>/.workflow/workflow.yaml via localstore.Init.
+// WorkflowInput shape festival mode uses, then initializes
+// <cwd>/.workflow/workflow.yaml and starts a run unless --no-init is set.
 //
 // Festival-only flags (--position, --path, --festival) are rejected here so
 // users get an explicit error rather than confusing silent ignores.
@@ -49,6 +49,17 @@ func runStandaloneCreateWorkflow(ctx context.Context, opts *CreateWorkflowOption
 				WithField("path", target).
 				WithHint("remove it first or pick another directory"))
 	}
+	runtimeDir := filepath.Join(cwd, ".workflow")
+	if !opts.NoInit {
+		if _, err := os.Stat(runtimeDir); err == nil {
+			return emitWorkflowError(opts,
+				errors.Validation(".workflow/ already exists").
+					WithField("path", runtimeDir).
+					WithHint("use `fest workflow start` for an existing tracked workflow, or remove .workflow/ before creating a new one"))
+		} else if !os.IsNotExist(err) {
+			return emitWorkflowError(opts, errors.IO("checking .workflow/", err).WithField("path", runtimeDir))
+		}
+	}
 
 	workflowID := "wf-" + workflow.SanitizeBasenameAsSlug(opts.Name)
 	if err := workflow.ValidateWorkflowID(workflowID); err != nil {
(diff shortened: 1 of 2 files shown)


internal/commands/festival/create_workflow_standalone.go
```diff
@@ -66,22 +66,35 @@ func runStandaloneCreateWorkflow(ctx context.Context, opts *CreateWorkflowOption
...
 	if !opts.NoInit {
-		store := localstore.Open(filepath.Join(cwd, ".workflow"), target)
+		runtimeDir := filepath.Join(cwd, ".workflow")
+		store := localstore.Open(runtimeDir, target)
 		if err := store.Init(ctx, localstore.InitOptions{WorkflowID: workflowID}); err != nil {
 			return emitWorkflowError(opts, errors.Wrap(err, "initialize .workflow/"))
 		}
+		runID, err := store.StartRun(ctx, "")
+		if err != nil {
+			_ = os.RemoveAll(runtimeDir)
```
obey-agent, line 88: This rollback is too broad for the failure mode. `store.Init` does not prove this invocation created `.workflow/`; it only creates `workflow.yaml` inside `runtimeDir`, and `os.MkdirAll` is happy when the directory already exists. If a user has a pre-existing `.workflow/` directory without `workflow.yaml` (for example partial state, notes, or a `runs` path that later makes `StartRun` fail), this error path deletes the whole directory, including files this command did not create. `EnsureTracked` avoids that by refusing to bootstrap when `.workflow/` already exists before using `RemoveAll`. Please either reject pre-existing `runtimeDir` in this standalone create path too, or track and remove only the files/directories created during this attempt. That keeps rollback safe under the staff-level filesystem/data-ownership standard.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** The main behavior is correct: `fest create workflow` now starts a tracked standalone run, and the PR updates `fest next`/`workflow advance` integration coverage around that path. The implementation is small and uses the existing `localstore.Init` + `StartRun` APIs rather than creating a parallel runtime flow. One rollback edge case needs to be fixed before merge because it can delete pre-existing `.workflow/` content on a start failure.

**Key Findings:**
- Inline: `create_workflow_standalone.go:78` uses `os.RemoveAll(runtimeDir)` after `StartRun` fails, but this path does not prove the command created `runtimeDir`. Since `Store.Init` accepts an existing `.workflow/` directory when `workflow.yaml` is absent, the rollback can remove files this command did not create. Reject pre-existing `.workflow/` before init, or track and remove only the files/directories created by this attempt.

**What's Done Well:** The happy path is cleanly integrated with `localstore`, JSON now reports `active_run_id`, suggestions no longer tell users to start an already-started run, and the integration test now verifies immediate `fest next --json` and `fest workflow advance` after create.

**Validation:**
- `go test ./internal/commands/festival ./internal/commands/next ./internal/commands/workflow ./internal/workflow/localstore`
- `go test -short ./...`
- `just build test-binary`
- `go test -tags integration ./tests/integration -run TestCreateWorkflowStandalone -count=1 -timeout 5m`
- `git diff --check refs/pr-review/182^ refs/pr-review/182`

**Staff Standard:** Not yet. The feature shape and coverage are good, but the rollback needs data-ownership safety before I would be comfortable merging it.

obey-agent:
Addressed the review finding: standalone create now refuses a pre-existing .workflow/ directory before writing WORKFLOW.md, so rollback only applies to runtime state created by the current invocation. Added integration coverage that preserves existing .workflow/ content and verifies no WORKFLOW.md or workflow.yaml is written on that error path.

obey-agent (approved):
**Verdict: Approve**

This is a fresh pass against head `d82e937` after the latest merge from `main`. The PR cleanly extends standalone `fest create workflow` so the default path initializes `.workflow/`, starts an active run, and leaves `fest next` immediately usable without adding a parallel runtime path.

**Key Findings**
- No blocking or actionable findings. The pre-existing `.workflow/` guard is in the right place, before `WORKFLOW.md` is written, so rollback cannot delete runtime state owned by an earlier setup.
- The create -> start-run -> next -> advance lifecycle is covered by the integration test and matches the localstore invariants around a single active run.

**What's Done Well**
- Reuses `localstore.Init` and `StartRun` instead of inventing a separate create-time runtime path.
- Keeps structured JSON output current by exposing `active_run_id` and removing the stale default suggestion to run `fest workflow start`.
- Adds regression coverage for immediate `fest next`, `workflow advance`, JSON output, and preserving pre-existing `.workflow/` contents.

**Verification**
- `go test ./internal/commands/festival ./internal/commands/next ./internal/commands/workflow ./internal/workflow/localstore`
- `go test -short ./...`
- `just build test-binary`
- `go test -tags integration ./tests/integration -run TestCreateWorkflowStandalone -count=1`

**Staff Standard**
Yes, I would be comfortable submitting this as my own work. The change is focused, uses the existing runtime contracts, handles the meaningful rollback risk, and has the right coverage for the workflow it changes.

PR #188, opened 2026-05-27
Validate WORKFLOW.md step numbering (closes #186)

lancekrogers:
## Summary

Closes #186.

`fest validate` previously did not catch WORKFLOW.md files whose `## Step N:` headings did not form a contiguous 1..N sequence. The CW0003 repro had `Step 0..Step 7` because an agent added Step 0 retroactively instead of renumbering. The runtime then silently mis-tracked progress (skipped the orphaned Step 0, advanced past the last step).

This PR adds a shared, deliberately narrow step-numbering validator and wires it into two surfaces.

### Shared validator
`internal/workflow/standalone/validate.go`
- New `ValidateStepNumbering(ctx, path)` function. Reuses the existing `internal/guidance/workflow` parser. Reports the first failure with the actual found sequence and concrete remediation options (renumber vs merge Step 0 into Step 1).
- Single check by design: contiguous 1..N. No section name / state table / checkpoint syntax validation. WORKFLOW.md is the lightweight customizable surface; stricter checks would fight that.
- Kept the existing `ValidateWorkflowDoc` untouched so `bootstrap.go` / `init.go` callers do not start failing mid-creation on numbering drift.

### New subcommand
`internal/commands/workflow/validate.go`
- `fest workflow validate [path]` validates a standalone WORKFLOW.md.
- Defaults to `./WORKFLOW.md` when no arg given.
- Exits non-zero on validation failure.
- Distinguishes missing file (NOT_FOUND) from directory-as-path (VALIDATION).

### Festival-tree integration
`internal/commands/validation/validate_workflow.go`
- New `validateWorkflowDocsChecks` walks the festival tree (at any depth: phase, sequence, task) for any `WORKFLOW.md` and runs the shared validator on each.
- Skips `.fest`, `.workflow`, and `.git` directories.
- Surfaces diagnostics under a new `WORKFLOW` section in the `fest validate` report.
- New error code `workflow_numbering` (plus `workflow_scan` for walk failures).
- Adds a suggestion pointing operators at `fest workflow validate <path>` for detailed remediation.

## Test plan

- `just test unit` -> 89 packages, 2108/2108 tests passed.
- `just lint all` -> 0 issues.
- `just build quick` -> binary built.
- Manual CLI exercise:
  - `fest workflow validate` with Step 0..2 -> exits 1, message includes "must start at 1", hint offers renumber-or-merge.
  - `fest workflow validate` with gap 1,2,4 -> exits 1, message says "not contiguous (jumped from 2 to 4)".
[...]

PR #189, opened 2026-05-27
Add fest workflow renumber (closes #187)

### LINKED ISSUE #187: Add fest workflow renumber command for renumbering steps in WORKFLOW.md
## Summary

`fest` already has `renumber` for phases, sequences, and tasks. Add `fest workflow renumber` to renumber `## Step N:` headings inside a `WORKFLOW.md` file so they form a contiguous 1-indexed sequence. Pairs with #186 (`fest workflow validate`): validate detects the drift, renumber fixes it in one command.

## Motivation

Real instance from `obey-campaign/festivals/active/camp-workitem-linking-and-commits-CW0003/004_PR_302_SCOPE_REVIEW/WORKFLOW.md`: an agent retroactively inserted `Step 0` instead of renumbering the existing 1..7 steps, producing `Step 0..Step 7`. Manual renumbering is mechanical and easy to get wrong. A dedicated command makes the fix trivial for agents and humans.

## Proposed behavior

- Default target: `./WORKFLOW.md`. Accepts an explicit path argument.
- Walks `## Step N:` headings in document order and rewrites them to `## Step 1:`, `## Step 2:`, ..., `## Step K:`.
- Preserves the heading suffix verbatim (`## Step 0: PRECONDITION - Confirm...` becomes `## Step 1: PRECONDITION - Confirm...`).
- `--dry-run` prints the rename plan without writing.
- Idempotent on already-correct docs.

## Scope: headings only

Do **not** touch:

- Step body text (no rewriting of cross-references like "see Step 3"). Document as a known v1 limitation.
- The "Workflow State Tracking" table. That table is agent decoration, not source of truth (state lives in `.fest/progress_events.jsonl`).
- Any other markdown content.

This keeps the command predictable and aligned with the lightweight/customizable design of `WORKFLOW.md`.

## Error cases

- **Duplicate step numbers** (`Step 1, Step 2, Step 2, Step 3`): exit non-zero, print offending headings with line numbers, instruct the user to resolve manually. Auto-merging is ambiguous.
- **Malformed `## Step` headings** (non-numeric): exit non-zero with line number.
- **No `## Step N:` headings found**: exit non-zero, "nothing to renumber."

## Success cases

- `Step 0..N` -> renumber to `Step 1..N+1`. The CW0003 case.
- Gaps (`Step 1, 2, 4, 5`) -> compact to `Step 1..4`.
- Already-correct -> no-op.

## Acceptance

- Running on a CW0003-style doc with `Step 0..7` produces `Step 1..8`, and the doc passes `fest workflow validate` (#186) afterward.
- `--dry-run` prints the rename plan without writing.
- Idempotent on already-correct docs.
- Refuses to write when duplicates are present.

## Relationship to #186

Once this command exists, the validator's remediation text in #186 should point at it:

```
Fix one of:
  a) Run `fest workflow renumber` to renumber to Step 1..8
  b) Merge Step 0's content into Step 1 and delete the heading
```

## Known limitations (v1)

- Inline cross-references like "see Step 3" in step bodies are **not** rewritten. A future `--rewrite-refs` flag could address this; not required for v1.
- Generated state-tracking tables in the doc are not updated. They're decoration; agents can regenerate or delete them.

lancekrogers:
## Summary

Adds `fest workflow renumber [path]`. Walks `## Step N:` headings in a WORKFLOW.md and rewrites the numbers to form a contiguous 1-indexed sequence in document order. Closes #187.

Default target is `./WORKFLOW.md`. Accepts an explicit path argument.

## Behavior

What it touches:
- Only `## Step N:` heading lines. The heading suffix is preserved verbatim (`## Step 0: PRECONDITION - Confirm...` becomes `## Step 1: PRECONDITION - Confirm...`).

What it leaves alone:
- Step body text (cross-references inside bodies are a known v1 limitation, see below).
- The Workflow State Tracking table inside the doc.
- Every other markdown construct in the file.

Flag ergonomics mirror the existing `fest renumber phase/sequence/task` commands so users have one mental model across all renumber subcommands:
- `--dry-run` (default true) prints the rename plan without writing.
- `--skip-dry-run` applies changes immediately.
- `--backup` writes a `.bak` copy of the original alongside before applying.

Error cases (exit non-zero, no write):
- Duplicate step numbers (e.g. `1, 2, 2, 3`). The offending step numbers and their line numbers are reported.
- No `## Step N:` headings found.
- File not found or unreadable.

Success cases:
- `Step 0..N` -> `Step 1..N+1` (the CW0003 case).
- Gaps `Step 1, 2, 4, 5` -> compact to `Step 1..4`.
- Already-correct `Step 1..K` -> no-op, exit 0, no write.

## Implementation notes

- Reuses the existing step-header regex from `internal/guidance/workflow/parser.go` via a new `StepHeaderRegexp()` accessor. The pattern stays in one place.
- Uses `festerrors` for all error returns (NotFound, IO, Validation, Parse).
- The new command is registered with `scope.Global` because it operates on an explicit file path, not a resolved festival context. This lets you renumber a WORKFLOW.md from any directory.

## Known v1 limitations

- Cross-references inside step bodies (for example `see Step 3`) are not rewritten. Update those by hand after renumbering if needed. The `Long` help text calls this out.

## Test plan

- [x] `just build` passes.
- [x] `just test unit` passes (2105 tests).
- [x] `just lint all` reports 0 issues.
- [x] Unit tests cover: Step 0..N shifts to 1..N+1, gaps compact, already-correct no-op, duplicates error, empty file error, file-not-found error, dry-run does not write, success writes, suffix preservation, backup file written when requested.
- [x] Smoke test on a sample doc: `Step 0,1,2,4` -> `Step 1,2,3,4` end-to-end (dry-run preview matches written output).

## Files

- `internal/commands/workflow/renumber.go` (new)
- `internal/commands/workflow/renumber_test.go` (new)
- `internal/commands/workflow/workflow.go` (wire up subcommand)
- `internal/guidance/workflow/parser.go` (add `StepHeaderRegexp()` accessor)

Closes #187.

### DIFF
--- a/internal/commands/workflow/renumber.go
+++ b/internal/commands/workflow/renumber.go
@@ -1,6 +1,7 @@
 package workflow
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -78,7 +79,7 @@ Examples:
 				return errors.Wrap(err, "resolving path").WithField("path", path)
 			}
 
-			return runWorkflowRenumber(cmd.OutOrStdout(), absPath, dryRun, backup)
+			return runWorkflowRenumber(cmd.Context(), cmd.OutOrStdout(), absPath, dryRun, backup)
 		},
 	}
 
@@ -89,7 +90,14 @@ Examples:
 	return cmd
 }
 
-func runWorkflowRenumber(out io.Writer, path string, dryRun, backup bool) error {
+func runWorkflowRenumber(ctx context.Context, out io.Writer, path string, dryRun, backup bool) error {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
+		return errors.Wrap(err, "context cancelled before reading workflow file").WithField("path", path)
+	}
+
 	content, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -100,7 +108,7 @@ func runWorkflowRenumber(out io.Writer, path string, dryRun, backup bool) error 
 		return errors.IO("reading workflow file", err).WithField("path", path)
 	}
 
-	plan, planErr := buildRenumberPlan(string(content))
+	plan, planErr := buildRenumberPlan(ctx, string(content))
 	if planErr != nil {
 		return planErr.WithField("path", path)
 	}
@@ -124,13 +132,17 @@ func runWorkflowRenumber(out io.Writer, path string, dryRun, backup bool) error 
 	}
 
 	if backup {
-		if err := os.WriteFile(path+".bak", content, 0o644); err != nil {
+		if err := writeFileAtomic(ctx, path+".bak", content, 0o644); err != nil {
 			return errors.IO("writing backup", err).WithField("path", path+".bak")
 		}
 	}
 
 	updated := applyRenumber(string(content), plan)
-	if err := os.WriteFile(path, []byte(updated), 0o644); err != nil {
+	mode := os.FileMode(0o644)
+	if info, statErr := os.Stat(path); statErr == nil {
+		mode = info.Mode().Perm()
+	}
+	if err := writeFileAtomic(ctx, path, []byte(updated), mode); err != nil {
 		return errors.IO("writing workflow file", err).WithField("path", path)
 	}
 
@@ -138,7 +150,14 @@ func runWorkflowRenumber(out io.Writer, path string, dryRun, backup bool) error 
 	return nil
 }
 
-func buildRenumberPlan(content string) (renumberPlan, *errors.Error) {
+func buildRenumberPlan(ctx context.Context, content string) (renumberPlan, *errors.Error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
+		return renumberPlan{}, errors.Wrap(err, "context cancelled before building renumber plan")
+	}
+
 	re := wf.StepHeaderRegexp()
 	matches := re.FindAllStringSubmatchIndex(content, -1)
 
@@ -188,6 +207,49 @@ func buildRenumberPlan(content string) (renumberPlan, *errors.Error) {
 	}
 
 	return plan, nil
+}
+
+func writeFileAtomic(ctx context.Context, path string, content []byte, mode os.FileMode) error {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	dir := filepath.Dir(path)
+	tmp, err := os.CreateTemp(dir, "."+filepath.Base(path)+".tmp-*")
+	if err != nil {
+		return err
+	}
+	tmpName := tmp.Name()
+	cleanup := true
+	defer func() {
+		if cleanup {
+			_ = os.Remove(tmpName)
+		}
+	}()
+
+	if _, err := tmp.Write(content); err != nil {
+		_ = tmp.Close()
+		return err
+	}
+	if err := ctx.Err(); err != nil {
+		_ = tmp.Close()
+		return err
+	}
+	if err := tmp.Chmod(mode); err != nil {
+		_ = tmp.Close()
+		return err
+	}
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+	if err := os.Rename(tmpName, path); err != nil {
+		return err
+	}
+	cleanup = false
+	return nil
 }
 
 type duplicate struct {
--- a/internal/commands/workflow/renumber.go
+++ b/internal/commands/workflow/renumber.go
@@ -1,0 +1,260 @@
+package workflow
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/Obedience-Corp/fest/internal/errors"
+	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
+	"github.com/Obedience-Corp/fest/internal/scope"
+	"github.com/spf13/cobra"
+)
+
+const defaultWorkflowFile = "WORKFLOW.md"
+
+type renumberPlan struct {
+	headings []renumberHeading
+	changes  int
+}
+
+type renumberHeading struct {
+	line      int
+	numStart  int
+	numEnd    int
+	oldNumber int
+	newNumber int
+	suffix    string
+}
+
+func newRenumberCmd() *cobra.Command {
+	var (
+		dryRun     bool
+		skipDryRun bool
+		backup     bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "renumber [path]",
+		Short: "Renumber step headings in a WORKFLOW.md to a contiguous 1-indexed sequence",
+		Annotations: map[string]string{
+			"scope": string(scope.Global),
+		},
+		Long: `Renumber the "## Step N:" headings inside a WORKFLOW.md so they form a
+contiguous 1-indexed sequence (Step 1, Step 2, ..., Step K) in document order.
+
+Only the heading lines are touched. Step body text, the Workflow State Tracking
+table, and any other markdown content are left exactly as written.
+
+The default target is ./WORKFLOW.md. Pass an explicit path to target another file.
+
+Like other renumber commands, --dry-run is the default so you can preview the
+plan; pass --skip-dry-run to write changes immediately.
+
+Known v1 limitations:
+  - Cross-references inside step bodies (for example "see Step 3") are not
+    rewritten. Update those references by hand after renumbering if needed.
+
+Examples:
+  fest workflow renumber                              # preview ./WORKFLOW.md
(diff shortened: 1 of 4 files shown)


obey-agent (commented):
**Verdict:** Approve

**Overview:** This PR adds `fest workflow renumber` with a small, focused rewrite strategy: locate shared parser step headings and replace only the numeric span. The dry-run default and optional backup behavior fit the risk profile of a command that mutates user-authored workflow docs.

**Key Findings:** None blocking. I did not find correctness issues worth inline comments.

**What's Done Well:** The command keeps the regex source centralized via `StepHeaderRegexp()`, refuses ambiguous duplicate renumbering, and has coverage for dry-run, write, no-op, backup, missing file, duplicate, and no-step cases.

**Staff Standard:** Yes. I would be comfortable submitting this as my own change.

**Verification:** Ran `go test ./internal/commands/workflow ./internal/guidance/workflow` on the PR worktree.

internal/commands/workflow/workflow.go
```diff
@@ -77,6 +77,7 @@ Examples:
 		newInitCmd(),
 		newStartCmd(),
 		newRunsCmd(),
+		newRenumberCmd(),
```
obey-agent, line 80: This line now conflicts with `main` after #188 landed, because `main` registers `newValidateCmd()` in the same subcommand list. GitHub reports this PR as `DIRTY`, and a local merge-tree confirms a content conflict in this file.

Please refresh the branch onto current `main` and keep both workflow subcommands registered (`newValidateCmd()` and `newRenumberCmd()`).

obey-agent (changes requested):
**Verdict: Request Changes**

**Overview**
The `fest workflow renumber` implementation is narrow and generally well scoped: it reuses the workflow step-header parser, preserves non-heading content, defaults to dry-run, and has good coverage for gaps, Step 0, duplicate detection, backups, no-op behavior, and missing files. I did not find a runtime correctness issue in the renumber logic itself.

**Key Findings**
- Inline: this PR is now merge-conflicted with current `main` after #188 landed. `main` registers `newValidateCmd()` in the same workflow subcommand list where this PR adds `newRenumberCmd()`. Please refresh the branch and keep both commands registered.

**What's Done Well**
- The rewrite is limited to `## Step N:` headings and does not attempt unsafe body cross-reference rewriting.
- Duplicate step numbers fail closed instead of guessing how to merge ambiguous headings.
- The command uses `festerrors` and mirrors the existing renumber dry-run/apply ergonomics.

**Validation**
- `go test ./internal/commands/workflow ./internal/guidance/workflow`
- `just test::unit`
- `just lint::all`
- `just build::quick`
- Local merge-tree against current `origin/main` reports a content conflict in `internal/commands/workflow/workflow.go`.

**Staff Standard**
Not ready to merge until the branch is refreshed against current `main` and the conflict is resolved. After that, the implementation should be straightforward to re-review.

internal/commands/workflow/renumber.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	cmd.Flags().BoolVar(&dryRun, "dry-run", true, "preview changes without writing")
+	cmd.Flags().BoolVar(&skipDryRun, "skip-dry-run", false, "skip preview and apply changes immediately")
+	cmd.Flags().BoolVar(&backup, "backup", false, "write a .bak copy of the original before applying")
+
+	return cmd
+}
+
+func runWorkflowRenumber(out io.Writer, path string, dryRun, backup bool) error {
```
obey-agent: **Context propagation.** This new path drops `context.Context` entirely, but the sibling renumber family it mirrors threads ctx through and checks `ctx.Err()` at each entry (`internal/festival/renumber.go`: `RenumberPhases/Sequences/Tasks(ctx, ...)`). The project standard (CLAUDE.md: "every I/O or long-running function takes context.Context") applies here too — there's real file I/O below. `cmd.Context()` is already available in `RunE`; thread it as the first param of `runWorkflowRenumber`/`buildRenumberPlan` and check `ctx.Err()` before the read/write so cancellation propagates and the command stays consistent with its siblings.

internal/commands/workflow/renumber.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	if backup {
+		if err := os.WriteFile(path+".bak", content, 0o644); err != nil {
+			return errors.IO("writing backup", err).WithField("path", path+".bak")
+		}
+	}
+
+	updated := applyRenumber(string(content), plan)
+	if err := os.WriteFile(path, []byte(updated), 0o644); err != nil {
```
obey-agent: **Non-atomic rewrite of the user's source doc.** This is a truncate-then-write in place on a file whose whole purpose is to be a hand-authored source of truth, and `--backup` defaults to false. If the process is interrupted mid-write (signal, disk full, crash) the WORKFLOW.md is left truncated/corrupted with no copy. A tool whose entire job is rewriting the user's file should make the replace atomic: write to a temp file in the same directory, then `os.Rename` over the original (rename is atomic within a filesystem). That makes `--backup` a true belt-and-suspenders option rather than the only thing standing between an interrupt and data loss.

obey-agent (commented):

**Verdict: Request Changes** (two small, concrete fixes; the feature itself is solid)

**Overview.** A focused, well-scoped addition: `fest workflow renumber` rewrites `## Step N:` headings to a contiguous 1-indexed sequence. The byte-offset rewrite is correct (matches are ascending and only the digit span is replaced, so the suffix and all body text are preserved verbatim), the regex is reused via the new `StepHeaderRegexp()` accessor rather than redefined, error handling uses `festerrors` throughout, and the test suite is genuinely thorough — Step 0 shift, gap compaction, no-op, duplicates, empty file, file-not-found, dry-run-doesn't-write, backup, and suffix preservation are all covered. Flag ergonomics correctly mirror the existing renumber commands.

**Key findings (see inline):**
- **Context propagation** — the new path drops `context.Context`, but the sibling renumber family it mirrors threads ctx and checks `ctx.Err()`, and CLAUDE.md mandates it for I/O. `cmd.Context()` is already in hand; thread it.
- **Non-atomic in-place rewrite** — truncate-then-write on the user's source-of-truth doc with `--backup` defaulting off. An interrupt mid-write corrupts the file. Use temp-file + `os.Rename` for an atomic replace.

**What's done well.** Single-source-of-truth regex via the accessor; honest, well-documented v1 limitation on body cross-references; refuses to act on duplicates rather than guessing; dry-run-by-default matches sibling muscle memory; tests assert behavior, not internals.

**Staff standard.** Close. With ctx threaded and the write made atomic, I'd be proud to ship this. Both fixes are small and mechanical.

internal/commands/workflow/renumber.go
obey-agent: Addressed in 2fc451a. The rewrite now uses a temp file in the same directory, applies the desired mode, closes it, then `os.Rename`s over the original. `--backup` uses the same atomic helper, and the source document preserves its existing file mode.

internal/commands/workflow/renumber.go
obey-agent: Addressed in 2fc451a. `cmd.Context()` now threads through `runWorkflowRenumber` and `buildRenumberPlan`, with cancellation checked before reading/planning and again before the atomic write path. Added a cancelled-context regression.

internal/commands/workflow/workflow.go
obey-agent: Current head is merge-clean and keeps both workflow subcommands registered: `newRenumberCmd()` and `newValidateCmd()` are present in the workflow command list.

obey-agent (commented):
## Staff Re-Review — Approve

**Verdict: Approve**

**Overview.** Re-reviewed the delta in `fix: harden workflow renumber writes` (2fc451a) against my prior round. Both findings are fully and correctly addressed; verified locally with a clean `go build ./...` and a passing `internal/commands/workflow` test run.

**Prior findings — resolved:**
- **Context propagation** — `runWorkflowRenumber` and `buildRenumberPlan` now take `context.Context` as the first parameter, `cmd.Context()` is threaded from `RunE`, and `ctx.Err()` is checked before the read and before plan construction. The new `TestRunWorkflowRenumberCancelledContext` proves a cancelled context returns `context.Canceled` and writes nothing. Matches the sibling renumber family's convention.
- **Non-atomic write** — replaced with `writeFileAtomic`: temp file in the same directory, write, `ctx.Err()` checkpoint, `Chmod`, close, then `os.Rename` over the target, with deferred temp cleanup on any failure path. Also now preserves the original file mode via `os.Stat(...).Mode().Perm()` instead of hardcoding 0644. The interrupt-corruption window is closed, and `--backup` is now genuinely belt-and-suspenders.

**What's done well.** The fix is idiomatic and minimal: same-dir temp guarantees a same-filesystem (atomic) rename, the cleanup flag flips only after a successful rename, and cancellation is checked at the right boundaries. Tests were updated coherently rather than just made to pass.

**Optional (non-blocking).** If you want power-loss durability (not just interrupt-safety), an `tmp.Sync()` before close — and optionally a directory fsync after rename — would harden it further. Not required for this tool.

**Staff standard.** Yes — I'd be proud to submit this. Ship it.

PR #190, opened 2026-05-27
Failed gate remediation routing in fest next (closes #185)

### LINKED ISSUE #185: Support failed gate remediation routing in fest next
## Problem

Today a phase gate can be approved or rejected, but rejection blocks the current gate in place. That is accurate for simple revision loops, but it does not model this common festival case:

1. A gate is evaluated and should **not** pass.
2. The failure is real and should remain recorded as not approved.
3. The operator adds a new remediation phase to correct the issues that caused the gate to fail.
4. `fest next` should continue into that remediation phase.
5. After remediation completes, `fest next` should route back to the failed gate for re-evaluation.

Current workarounds are awkward:

- `fest workflow reject --reason ...` records the failure, but keeps `fest next` stuck at the rejected gate.
- `fest workflow skip --reason ...` can unblock traversal, but it does not accurately record that the gate failed/not approved.
- Approving the gate after adding remediation blurs the audit trail: the gate did not actually pass; it was deferred to a corrective phase.

## Desired behavior

Add first-class support for failed gate remediation routing.

A gate should be able to enter a state like `failed_with_remediation` that records:

- failed phase and gate step
- failure reason
- remediation phase or task path
- timestamp/operator context
- whether the remediation has completed
- whether the failed gate has been rechecked afterward

`fest next` should then route agents clearly:

```text
Previous gate failed:
004_PR_302_SCOPE_REVIEW / COMPLETENESS

Remediation phase:
005_FIX_PR_302_AGGREGATE

Next task:
005_FIX_PR_302_AGGREGATE/01_scope_contract_cleanup/01_prepare_pr302_branch.md
```

After the remediation phase completes, `fest next` should route back to the failed gate for re-evaluation instead of treating the original gate as approved.

## Possible CLI shape

One option:

```bash
fest workflow reject \
  --reason "PR 302 is not ready for approval" \
  --remediation-phase 005_FIX_PR_302_AGGREGATE
```

Another option:

```bash
fest workflow fail \
  --reason "PR 302 is not ready for approval" \
  --continue-with 005_FIX_PR_302_AGGREGATE
```

The exact command name is less important than preserving the semantics:

- not approved
- not skipped
- remediation linked
- `fest next` stays useful
- gate can be rechecked after remediation

## Agent UX requirement

Agents should not need to infer this policy. `fest next` should show the remediation route explicitly and provide the exact next task. The feature should avoid arbitrary graph semantics unless needed; this should stay a narrow gate-failure remediation path.

## Acceptance criteria

- [ ] A gate/checkpoint can be recorded as failed with a linked remediation phase or task.
- [ ] The failed gate remains auditable as not approved.
- [ ] `fest next` routes to the remediation work instead of remaining stuck at the failed gate.
- [ ] After remediation completion, `fest next` routes back to the failed gate for re-evaluation.
- [ ] `fest workflow status` shows the failed gate, reason, and remediation link.
- [ ] The behavior is covered by integration tests for reject/fail -> remediation -> recheck.
- [ ] Docs explain when to use normal `reject`, `skip`, and failed-gate remediation.

## Concrete motivating case

In CW0003, PR 302 aggregate review gates found real blockers. A new phase (`005_FIX_PR_302_AGGREGATE`) was added to correct those blockers. The desired audit trail is: the original readiness gate did not pass, remediation was planned, `fest next` should advance into that new phase, and the readiness gate should be re-evaluated after the remediation phase completes.

lancekrogers:
## Summary

Closes #185. Adds first-class support for failed-with-remediation gate routing so a rejected phase gate can advance into a linked remediation phase, then route back for re-evaluation, without lying about the audit trail.

## CLI shape

Extend the existing reject command with an additive flag:

```
fest workflow reject --reason "PR 302 is not ready" --remediation-phase 005_FIX_PR_302_AGGREGATE
```

Chosen over a new `fail` verb because it is additive, requires no new vocabulary, and keeps agent learning cost low. Behavior without the flag is unchanged.

## State machine

- New step status `failed_with_remediation` (non-terminal) and a `RemediationPhase` field on `StepState`.
- New `WorkflowState.RejectWithRemediation(reason, phase)` and `ClearFailedRemediation()` transitions.
- Two new progress events recorded in `.fest/progress_events.jsonl`:
  - `wf_step_fail_remediation` (carries `remediation_phase` and `feedback`)
  - `wf_step_recheck`
- Audit trail for a full loop: `wf_step_fail_remediation` -> `wf_step_recheck` -> `wf_step_done`. The gate is never silently approved.

## fest next routing

Before normal phase resolution, `fest next` scans gate-prefixed phase state for any step in `failed_with_remediation`. If found:

- If the remediation phase is incomplete (tasks and any WORKFLOW.md not done), print the failed-gate banner and render the remediation phase next step. WORKFLOW.md, GATES.md, and task-based remediation phases are all supported.
- If the remediation phase is complete, print a recheck banner, emit a `wf_step_recheck` event so the step returns to in-progress, and render the gate step for re-evaluation. Approve or reject then closes the record.

`fest workflow status` now shows failed-gate entries with reason and remediation link.

## Test plan

- [x] `just build`
- [x] `just test unit` (2112 tests pass)
- [x] `just lint all` (0 issues)
- [x] Unit tests for the new state transitions in `internal/guidance/workflow/failed_remediation_test.go`.
- [x] Progress event materialization tests in `internal/progress/failed_remediation_test.go` including the full fail -> recheck -> done sequence.
- [x] Reject command tests in `internal/commands/workflow/reject_remediation_test.go` covering the happy path, missing-phase validation, non-numbered-phase validation, and the default (no flag) path.
- [x] `fest next` discovery and routing tests in `internal/commands/next/remediation_test.go` for find, clear-after-done, and remediation-phase completion detection.
- [x] End-to-end loop in `internal/e2e/failed_remediation_test.go`: reject with remediation, complete the remediation workflow, recheck, approve, verify the audit-event sequence and final state.

## Docs

`methodology/README.md` gained a "When to use reject, skip, and failed-gate remediation" section with the CW0003-style example flow.

## Deviations

None of substance. The recheck transition is automatic when `fest next` detects a complete remediation phase (the operator does not run a separate `recheck` command). This was not specified in the issue but follows the spirit of "preserving useful fest next" without adding a new verb. The acceptance criteria are met as written.

Closes #185

### DIFF
--- a/internal/commands/next/next.go
+++ b/internal/commands/next/next.go
@@ -184,9 +184,15 @@ func runNext(cmd *cobra.Command, args []string) error {
 	// Failed-gate remediation routing takes priority over normal phase
 	// resolution. When a phase gate was rejected with --remediation-phase,
 	// route into the remediation phase until it completes, then auto-recheck
-	// the failed gate.
-	if gate, gateErr := findFailedRemediationGate(ctx, festivalPath); gateErr == nil && gate != nil {
-		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate); handled {
+	// the failed gate. The progress event log is the source of truth for active
+	// failed gates, so a load/materialization failure here must fail closed
+	// rather than silently falling through to normal routing.
+	gate, gateErr := findFailedRemediationGate(ctx, festivalPath)
+	if gateErr != nil {
+		return errors.Wrap(gateErr, "checking for failed-gate remediation state")
+	}
+	if gate != nil {
+		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate, opts); handled {
 			return err
 		}
 	}
@@ -563,6 +569,16 @@ type RenderOptions struct {
 }
 
 func (o RenderOptions) showInlineContext() bool { return !o.NoContext }
+
+// isMachineOutput reports whether the active render mode expects pipeable or
+// structured output (path, cd, project-dir, short, or json) rather than the
+// default human banner plus instructions. Failed-gate remediation routing must
+// suppress its banner and navigator instructions in these modes so that
+// `fest next --path` and friends keep honoring their output contracts while a
+// remediation loop is active.
+func (o RenderOptions) isMachineOutput() bool {
+	return o.JSON || o.Path || o.CD || o.ProjectDir || o.Short
+}
 
 // runStandaloneNext renders the next step for a tracked standalone workflow.
 func runStandaloneNext(ctx context.Context, res *standalone.Result, opts RenderOptions) error {
--- a/internal/commands/next/next.go
+++ b/internal/commands/next/next.go
@@ -181,6 +181,16 @@ func runNext(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Failed-gate remediation routing takes priority over normal phase
+	// resolution. When a phase gate was rejected with --remediation-phase,
+	// route into the remediation phase until it completes, then auto-recheck
+	// the failed gate.
+	if gate, gateErr := findFailedRemediationGate(ctx, festivalPath); gateErr == nil && gate != nil {
+		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate); handled {
+			return err
+		}
+	}
+
 	// If mode flag is provided or --navigator flag is set, use guidance navigator
 	if modeFlag != "" || useNavigator {
 		return runNavigatorMode(ctx, cwd, festivalPath)
--- a/internal/commands/next/next.go
+++ b/internal/commands/next/next.go
@@ -187,14 +187,17 @@ func runNext(cmd *cobra.Command, args []string) error {
 	// the failed gate. The progress event log is the source of truth for active
 	// failed gates, so a load/materialization failure here must fail closed
 	// rather than silently falling through to normal routing.
-	gate, gateErr := findFailedRemediationGate(ctx, festivalPath)
+	remediationStore, gateErr := loadProgressStore(ctx, festivalPath)
+	if gateErr != nil {
+		return errors.Wrap(gateErr, "checking for failed-gate remediation state")
+	}
+	gate, gateErr := findFailedRemediationGateInStore(ctx, festivalPath, remediationStore)
 	if gateErr != nil {
 		return errors.Wrap(gateErr, "checking for failed-gate remediation state")
 	}
 	if gate != nil {
-		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate, opts); handled {
-			return err
-		}
+		_, err := routeFailedRemediationGate(ctx, festivalPath, remediationStore, gate, opts)
+		return err
 	}
 
 	// If mode flag is provided or --navigator flag is set, use guidance navigator
--- a/internal/commands/next/remediation.go
+++ b/internal/commands/next/remediation.go
@@ -100,9 +100,11 @@ func lookupGateStepName(ctx context.Context, phasePath string, stepNum int) (str
 }
 
 // isRemediationPhaseComplete reports whether the remediation phase has all
-// task-based and workflow work complete. Sequence/task and WORKFLOW.md
-// completion both qualify; phase gates on the remediation phase are not
-// required (they would block remediation completion in a circular fashion).
+// task-based and workflow work complete. A remediation phase is driven by its
+// WORKFLOW.md and/or numbered sequences; it is validated to contain at least
+// one of these at link time (see validateRemediationPhase). A GATES.md on the
+// remediation phase is not part of the remediation loop: the original failed
+// gate is the verification step and is rechecked once remediation work is done.
 func isRemediationPhaseComplete(ctx context.Context, festivalPath, remediationPhase string) (bool, error) {
 	if remediationPhase == "" {
 		return false, nil
@@ -151,9 +153,6 @@ func renderRemediationRoute(ctx context.Context, festivalPath string, gate *fail
 
 	if _, err := os.Stat(filepath.Join(remPath, "WORKFLOW.md")); err == nil {
 		return runWorkflowMode(ctx, festivalPath, remPath)
-	}
-	if _, err := os.Stat(filepath.Join(remPath, "GATES.md")); err == nil {
-		return runPhaseGateMode(ctx, festivalPath, remPath)
 	}
 	return renderFirstIncompleteTask(ctx, festivalPath, remPath)
 }
--- a/internal/commands/next/remediation.go
+++ b/internal/commands/next/remediation.go
@@ -29,9 +29,30 @@ type failedGate struct {
 	RemediationPhase string
 }
 
-func findFailedRemediationGate(ctx context.Context, festivalPath string) (*failedGate, error) {
+func loadProgressStore(ctx context.Context, festivalPath string) (*progress.Store, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
 	store := progress.NewStore(festivalPath)
 	if err := store.Load(ctx); err != nil {
+		return nil, err
+	}
+	return store, nil
+}
+
+func findFailedRemediationGate(ctx context.Context, festivalPath string) (*failedGate, error) {
+	store, err := loadProgressStore(ctx, festivalPath)
+	if err != nil {
+		return nil, err
+	}
+	return findFailedRemediationGateInStore(ctx, festivalPath, store)
+}
+
+func findFailedRemediationGateInStore(ctx context.Context, festivalPath string, store *progress.Store) (*failedGate, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
 		return nil, err
 	}
 
@@ -108,17 +129,28 @@ func lookupGateStepName(ctx context.Context, phasePath string, stepNum int) (str
 // remediation phase is not part of the remediation loop: the original failed
 // gate is the verification step and is rechecked once remediation work is done.
 func isRemediationPhaseComplete(ctx context.Context, festivalPath, remediationPhase string) (bool, error) {
+	store, err := loadProgressStore(ctx, festivalPath)
+	if err != nil {
+		return false, err
+	}
+	phasePath, err := resolveRemediationPhasePath(festivalPath, &failedGate{RemediationPhase: remediationPhase})
+	if err != nil {
+		return false, err
+	}
+	return isRemediationPhaseCompleteWithStore(ctx, store, phasePath, remediationPhase)
+}
+
+func isRemediationPhaseCompleteWithStore(ctx context.Context, store *progress.Store, phasePath, remediationPhase string) (bool, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
+		return false, err
+	}
 	if remediationPhase == "" {
 		return false, nil
 	}
-	phasePath := filepath.Join(festivalPath, remediationPhase)
-	if _, err := os.Stat(phasePath); err != nil {
-		return false, nil
-	}
-
-	store := progress.NewStore(festivalPath)
-	storeLoaded := store.Load(ctx) == nil
-	return shared.ArePhaseTasksAndWorkflowComplete(ctx, storeLoaded, store, phasePath, remediationPhase), nil
+	return shared.ArePhaseTasksAndWorkflowComplete(ctx, store != nil, store, phasePath, remediationPhase), nil
 }
 
 // routeFailedRemediationGate produces the appropriate `fest next` output for
@@ -128,31 +160,53 @@ func isRemediationPhaseComplete(ctx context.Context, festivalPath, remediationPh
 // and renders the gate step for re-evaluation. Returns (true, nil) when it
 // fully handled the request. opts controls output formatting so machine-output
 // modes are honored while remediation is active.
-func routeFailedRemediationGate(ctx context.Context, festivalPath string, gate *failedGate, opts RenderOptions) (bool, error) {
+func routeFailedRemediationGate(ctx context.Context, festivalPath string, store *progress.Store, gate *failedGate, opts RenderOptions) (bool, error) {
 	if gate == nil {
 		return false, nil
 	}
+	if store == nil {
+		var err error
+		store, err = loadProgressStore(ctx, festivalPath)
+		if err != nil {
+			return true, err
+		}
+	}
 
-	complete, err := isRemediationPhaseComplete(ctx, festivalPath, gate.RemediationPhase)
+	remPath, err := resolveRemediationPhasePath(festivalPath, gate)
+	if err != nil {
+		return true, err
+	}
+
+	complete, err := isRemediationPhaseCompleteWithStore(ctx, store, remPath, gate.RemediationPhase)
 	if err != nil {
 		return true, err
 	}
 
 	if complete {
-		return true, renderRecheckRoute(ctx, festivalPath, gate, opts)
+		return true, renderRecheckRoute(ctx, festivalPath, store, gate, opts)
 	}
-	return true, renderRemediationRoute(ctx, festivalPath, gate, opts)
+	return true, renderRemediationRoute(ctx, festivalPath, store, gate, remPath, opts)
 }
 
-func renderRemediationRoute(ctx context.Context, festivalPath string, gate *failedGate, opts RenderOptions) error {
-	remPath := filepath.Join(festivalPath, gate.RemediationPhase)
-	if _, err := os.Stat(remPath); err != nil {
-		return festerrors.NotFound("remediation phase directory").
-			WithField("phase", gate.RemediationPhase).
-			WithField("path", remPath)
+func resolveRemediationPhasePath(festivalPath string, gate *failedGate) (string, error) {
+	if gate == nil || gate.RemediationPhase == "" {
+		return "", festerrors.Validation("failed gate is missing a remediation phase").
+			WithHint("Reject the gate again with --remediation-phase, or clear the failed-remediation state with 'fest workflow reject --reason \"...\"'")
 	}
 
-	useWorkflow := shouldRouteToRemediationWorkflow(ctx, festivalPath, remPath, gate.RemediationPhase)
+	remPath := filepath.Join(festivalPath, gate.RemediationPhase)
+	info, err := os.Stat(remPath)
+	if err != nil || !info.IsDir() {
+		return "", festerrors.NotFound("remediation phase directory").
+			WithField("phase", gate.RemediationPhase).
+			WithField("path", remPath).
+			WithHint(fmt.Sprintf("The failed gate still references %s, but that phase no longer exists. Restore or rename the phase, or clear the failed-remediation state with 'fest workflow reject --phase %s --reason \"...\"'.", gate.RemediationPhase, gate.PhaseName))
+	}
+	return remPath, nil
+}
+
+func renderRemediationRoute(ctx context.Context, festivalPath string, store *progress.Store, gate *failedGate, remPath string, opts RenderOptions) error {
+	useWorkflow := shouldRouteToRemediationWorkflowWithStore(ctx, store, remPath, gate.RemediationPhase)
 
 	if opts.isMachineOutput() {
 		view, err := remediationRouteView(ctx, festivalPath, remPath, gate, useWorkflow)
@@ -164,7 +218,7 @@ func renderRemediationRoute(ctx context.Context, festivalPath string, gate *fail
 
 	printFailedGateBanner(gate)
 	if useWorkflow {
-		return runWorkflowMode(ctx, festivalPath, remPath)
+		return runWorkflowModeWithStore(ctx, festivalPath, remPath, store)
 	}
 	return renderFirstIncompleteTask(ctx, festivalPath, remPath, opts)
 }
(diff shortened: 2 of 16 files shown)


internal/commands/workflow/reject.go
```diff
@@ -55,27 +76,42 @@ func runReject(ctx context.Context, reason string) error {
...
 
-	// Verify this is a checkpoint step
 	if !step.Checkpoint.IsBlocking() {
-		return fmt.Errorf("step %d does not have a blocking checkpoint\n\nReject is only for checkpoint steps", currentStepNum)
+		return festerrors.Validation("step does not have a blocking checkpoint").
+			WithField("step", currentStepNum).
+			WithHint("Reject is only for checkpoint steps")
+	}
+
+	if remediationPhase != "" {
```
obey-agent, line 98: This flag path needs to reject non-gate navigators or wire routing for them. `getWorkflowNavigator` can target a normal `WORKFLOW.md` checkpoint, and this branch will then persist `failed_with_remediation` under the regular workflow phase key. `fest next` only scans `gate:`-prefixed state in `findFailedRemediationGate`, so that user is left in a non-terminal state with no remediation routing or recheck path. Since the feature is explicitly phase-gate remediation, validate that the active navigator is in gate mode before accepting `--remediation-phase`.

internal/commands/next/remediation.go
```diff
@@ -0,0 +1,238 @@
...
+		return false, nil
+	}
+	phasePath := filepath.Join(festivalPath, remediationPhase)
+	if _, err := os.Stat(phasePath); err != nil {
+		return false, nil
+	}
+
+	store := progress.NewStore(festivalPath)
+	storeLoaded := store.Load(ctx) == nil
+	return shared.ArePhaseTasksAndWorkflowComplete(ctx, storeLoaded, store, phasePath, remediationPhase), nil
```
obey-agent: This completion check makes `GATES.md` remediation phases complete before their gate ever runs. `ArePhaseTasksAndWorkflowComplete` returns true when the phase has no `WORKFLOW.md` and no numbered sequence dirs, so a gate-only remediation phase goes straight to `renderRecheckRoute`; the `GATES.md` branch in `renderRemediationRoute` is unreachable for the advertised case. Either include remediation gate state in this completion test, or remove the unsupported `GATES.md` routing/docs so `fest next` cannot silently skip the remediation phase's gate.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** The failed-gate remediation model is directionally strong and the event/audit trail design is the right shape, but I found two routing holes that can leave users with misleading or stuck workflow state. Both issues are around enforcing the intended gate-only contract and matching the advertised remediation phase types.

**Key Findings:**
- Inline: `--remediation-phase` can be accepted for a normal `WORKFLOW.md` checkpoint, but `fest next` only routes `gate:`-prefixed failed-remediation state, leaving the workflow non-terminal with no recheck path.
- Inline: a gate-only remediation phase is considered complete before its `GATES.md` runs, so `fest next` skips straight back to the original failed gate despite the PR claiming `GATES.md` remediation phases are supported.

**What's Done Well:** The new `failed_with_remediation` status is non-terminal, the JSONL audit sequence is explicit, and the happy-path e2e coverage exercises fail -> remediate -> recheck -> approve.

**Staff Standard:** Not yet. I would want the two routing invariants fixed and covered before shipping this.

**Verification:** Ran `go test ./internal/guidance/workflow ./internal/progress ./internal/commands/workflow ./internal/commands/next ./internal/e2e` on the PR worktree.

internal/commands/workflow/reject.go
obey-agent: Fixed in d039577. `runRejectWithRemediation` now rejects `--remediation-phase` unless the active navigator is in gate mode (new `Navigator.IsGate()`), so the flag can no longer persist `failed_with_remediation` under a non-`gate:` workflow key. The error directs operators to plain `reject --reason` (revise in place) or `skip` (done elsewhere) for normal checkpoints. Covered by `TestRunRejectWithRemediation_RejectsNonGateNavigator`, which also asserts no gate-prefixed state is written.

internal/commands/next/remediation.go
obey-agent: Fixed in d039577. Removed the unreachable `GATES.md` branch in `renderRemediationRoute` and corrected the `isRemediationPhaseComplete` comment: the original failed gate is the verification step and is rechecked once remediation work completes, so a remediation phase is driven only by its `WORKFLOW.md`/sequences. To close the silent-skip hole at the source, `validateRemediationPhase` now requires the linked phase to contain actionable work (`WORKFLOW.md` or numbered sequence dirs); a `GATES.md`-only phase is rejected at `reject` time rather than being treated as instantly complete. Covered by `TestValidateRemediationPhase_RequiresActionableWork` and `_AcceptsSequenceDirs`.

internal/commands/next/remediation.go
```diff
@@ -0,0 +1,237 @@
...
+	remPath := filepath.Join(festivalPath, gate.RemediationPhase)
+	if _, err := os.Stat(remPath); err != nil {
+		return festerrors.NotFound("remediation phase directory").
+			WithField("phase", gate.RemediationPhase).
+			WithField("path", remPath)
+	}
+
+	printFailedGateBanner(gate)
+
+	if _, err := os.Stat(filepath.Join(remPath, "WORKFLOW.md")); err == nil {
```
obey-agent: This unconditionally routes to `WORKFLOW.md` whenever the remediation phase has one. That breaks phases that contain both a remediation workflow and sequence/task work: `isRemediationPhaseComplete` correctly returns false until both are complete, but once the workflow is done this branch still calls `runWorkflowMode`, which prints `WORKFLOW COMPLETE` and returns instead of surfacing the remaining incomplete task.

Please only call `runWorkflowMode` while the remediation workflow is incomplete; after it completes, fall through to `renderFirstIncompleteTask`. Add a regression test for a remediation phase with a completed `WORKFLOW.md` plus an incomplete numbered task.

internal/commands/next/next.go
```diff
@@ -181,6 +181,16 @@ func runNext(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Failed-gate remediation routing takes priority over normal phase
+	// resolution. When a phase gate was rejected with --remediation-phase,
+	// route into the remediation phase until it completes, then auto-recheck
+	// the failed gate.
+	if gate, gateErr := findFailedRemediationGate(ctx, festivalPath); gateErr == nil && gate != nil {
```
obey-agent: If `findFailedRemediationGate` fails to load/materialize the progress store, this currently ignores the error and falls through into normal `fest next` routing. For failed-gate remediation, the progress event log is the source of truth for an active gate failure; failing open can route users past a remediation loop or hide a corrupted/unreadable event log.

Please propagate `gateErr` instead of silently ignoring it. A missing event log already loads as empty state, so an error here should be treated as a real operability problem.

obey-agent (changes requested):
**Verdict: Request Changes**

**Overview**
This PR moves the failed-gate remediation loop in the right direction: the new non-terminal state, explicit remediation phase link, recheck event, and documentation are all coherent. The latest head is cleanly merged with current `main`, but there are still two routing/operability issues that can mislead `fest next` in real remediation loops.

**Key Findings**
- Inline: `renderRemediationRoute` always chooses `WORKFLOW.md` when present. A remediation phase with both a completed workflow and incomplete sequence/task work will keep printing `WORKFLOW COMPLETE` instead of routing to the remaining task.
- Inline: `runNext` ignores `findFailedRemediationGate` errors. Since the progress event log is the source of truth for active failed gates, an unreadable/corrupt store should fail closed instead of silently falling through normal routing.

**What's Done Well**
- `failed_with_remediation` is correctly non-terminal, so gates are not silently approved.
- The audit trail shape (`wf_step_fail_remediation` -> `wf_step_recheck` -> `wf_step_done`) is explicit and covered by tests.
- The reject command now blocks `--remediation-phase` on non-gate workflow checkpoints and rejects GATES-only remediation phases, which avoids two easy dead ends.

**Validation**
Latest head reviewed: `11560739542e33d96dbafb61a129a75b364d32fb`.
- `go test ./internal/commands/next ./internal/commands/workflow ./internal/guidance/workflow ./internal/progress ./internal/e2e`
- `just test::unit` (2129/2129 passing)
- `just lint::all`
- `just build::quick`

**Staff Standard**
Not ready to merge yet. I would be comfortable after the mixed workflow-plus-task remediation routing case is fixed and progress-store load errors fail closed.

internal/commands/next/next.go
```diff
@@ -181,6 +181,16 @@ func runNext(cmd *cobra.Command, args []string) error {
...
+		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate); handled {
```
obey-agent: Major: this failed-remediation branch bypasses the `RenderOptions` captured above. While a failed gate is active, `fest next --json`, `--path`, `--short`, `--cd`, and `--no-context` all route through `routeFailedRemediationGate` and print banners plus workflow/`FormatText` output instead of the normal machine-readable or terse forms. That breaks automation that relies on `fest next --path`/JSON exactly when the remediation loop is active. Thread `opts` through the remediation route and add regressions for at least `--json` and `--path` with an active failed-remediation gate.

obey-agent (commented):
**Verdict: Request Changes**

I re-reviewed PR #190 at `11560739542e33d96dbafb61a129a75b364d32fb`. The gate-only remediation guard and GATES-only remediation rejection from `d039577` are good fixes, and the event model remains directionally solid. The PR still should not merge yet.

**Key findings**
- Existing inline: `renderRemediationRoute` always routes to `WORKFLOW.md` when present, so a remediation phase with a completed workflow plus incomplete sequence/task work prints `WORKFLOW COMPLETE` instead of surfacing the remaining task.
- Existing inline: `runNext` ignores errors from `findFailedRemediationGate`; failed-remediation routing should fail closed when the progress store cannot be loaded/materialized.
- New inline: the failed-remediation route bypasses `RenderOptions`, so `fest next --json`, `--path`, `--short`, `--cd`, and `--no-context` produce banner/text output while remediation is active instead of honoring the normal output contracts.

**What’s done well**
`failed_with_remediation` is non-terminal, the audit events preserve the fail → recheck → done sequence, and the reject command now blocks two dead-end states that were present in the first pass.

**Verification**
Passed on this head: focused remediation packages, `just test unit` (2129/2129), `just lint all`, `just build quick`, and `git diff --check origin/main...HEAD`.

**Staff standard**
Still request changes. I’d be comfortable after the mixed workflow-plus-task routing, progress-store error handling, and render-option contracts are fixed and covered.

internal/commands/next/remediation.go
```diff
@@ -0,0 +1,398 @@
...
+func renderRecheckRoute(ctx context.Context, festivalPath string, gate *failedGate, opts RenderOptions) error {
+	// Record the recheck transition first so the gate leaves
+	// failed_with_remediation regardless of output mode. Machine-output callers
+	// then receive a clean view; human callers get the banner plus the gate
+	// step instructions.
+	nav, err := newRecheckNavigator(ctx, festivalPath, gate)
+	if err != nil {
+		return err
+	}
+	if err := nav.Recheck(ctx); err != nil {
```
obey-agent, line 296: **Primary concern: `fest next` now mutates state, including in the pipe-oriented machine modes.** Before this PR the `next` package has zero state-writing calls — `fest next` is a pure report command, and `--path`/`--cd`/`--project-dir` are documented as pipe helpers (`methodology/README.md`: `fest next --path  # Just the file path (for piping)`). The recheck transition is recorded *first*, before rendering, so even `fest next --cd` / `--path` / `--json` fire a `wf_step_recheck` event and move the gate out of `failed_with_remediation`.\n\nConsequences:\n- `cd "$(fest next --cd)"` silently consumes the recheck as a side effect of navigating.\n- An agent that calls `fest next --json` to *inspect* state mutates it; a second call yields different output (the gate is gone from failed-remediation). The query is non-idempotent.\n- The human "RECHECK GATE" banner + instructions render only on the first call; a second `fest next` shows the normal step, losing the recheck context.\n\nI understand the deliberate design choice to avoid a new `recheck` verb (per the PR's Deviations section), and auto-recheck in *human* `fest next` is defensible. But a read/pipe query firing a state transition is a correctness hazard for fest's core agent-loop use case. At minimum, suppress the transition in `isMachineOutput()` modes (emit the view describing 'remediation complete, recheck pending' without recording the event), and only record `wf_step_recheck` on an explicit operator action (e.g. the next `fest workflow approve`/`reject`, or human-mode `fest next`). Would like your call on this before merge.

internal/commands/next/remediation.go
```diff
@@ -0,0 +1,398 @@
...
+// WORKFLOW.md and/or numbered sequences; it is validated to contain at least
+// one of these at link time (see validateRemediationPhase). A GATES.md on the
+// remediation phase is not part of the remediation loop: the original failed
+// gate is the verification step and is rechecked once remediation work is done.
+func isRemediationPhaseComplete(ctx context.Context, festivalPath, remediationPhase string) (bool, error) {
+	if remediationPhase == "" {
+		return false, nil
+	}
+	phasePath := filepath.Join(festivalPath, remediationPhase)
+	if _, err := os.Stat(phasePath); err != nil {
```
obey-agent: **Stale remediation-phase pointer can dead-end `fest next`.** The link is stored as a phase-name string (`RemediationPhase`). If that phase is later renamed or renumbered — and phase renumbering is a first-class fest operation (`fest renumber phase`, and now `fest workflow renumber` in #189) — this stat fails, returns `(false, nil)`, and routing falls to `renderRemediationRoute`, which returns a hard `NotFound`. Because the failed-gate check runs *before* normal phase resolution, `fest next` then errors on every invocation with no obvious way forward. The only escape (a bare `fest workflow reject` from inside the gate phase, which resets status to `blocked`) is non-obvious. Consider: (a) detecting the missing linked phase and surfacing an actionable recovery hint ('remediation phase X no longer exists; run ... to clear'), and/or (b) validating/repairing the stored link when phases are renumbered.

internal/commands/next/remediation.go
```diff
@@ -0,0 +1,398 @@
...
+		return false, nil
+	}
+
+	store := progress.NewStore(festivalPath)
+	storeLoaded := store.Load(ctx) == nil
```
obey-agent: **Redundant store loads per `fest next`.** When a failed gate is active, the JSONL event log is read and fully materialized 3–4 times in a single `fest next`: `findFailedRemediationGate` loads it, then `isRemediationPhaseComplete` loads it again (here), `shouldRouteToRemediationWorkflow` loads it again, and `newRecheckNavigator` loads it once more. It's correct (idempotent reads) but the event log grows for the life of a festival, so this is avoidable O(n) re-parsing on the hot path. Consider loading the store once at the top of the routing entrypoint and threading the loaded `*Store`/state down, rather than each helper opening its own.

internal/commands/next/next.go
```diff
@@ -181,6 +181,22 @@ func runNext(cmd *cobra.Command, args []string) error {
...
+	// route into the remediation phase until it completes, then auto-recheck
+	// the failed gate. The progress event log is the source of truth for active
+	// failed gates, so a load/materialization failure here must fail closed
+	// rather than silently falling through to normal routing.
+	gate, gateErr := findFailedRemediationGate(ctx, festivalPath)
+	if gateErr != nil {
+		return errors.Wrap(gateErr, "checking for failed-gate remediation state")
+	}
+	if gate != nil {
+		if handled, err := routeFailedRemediationGate(ctx, festivalPath, gate, opts); handled {
```
obey-agent: **Nit: this `if handled` guard can silently swallow an error.** `routeFailedRemediationGate` only returns `handled=false` when `gate == nil`, which is already excluded by the enclosing `if gate != nil`. So `handled` is always true here and the guard is dead. The risk is future drift: if the route ever returns `(false, someErr)`, this block would fall through to normal routing and drop `someErr` on the floor. Since the gate is guaranteed non-nil at this point, just `return routeFailedRemediationGate(...)` unconditionally.

obey-agent (commented):

**Verdict: Request Changes** (one design concern worth gating on; the rest is strong)

**Overview.** This is a substantial, carefully-built feature: a new non-terminal `failed_with_remediation` step status, additive `--remediation-phase` flag on `reject`, event-sourced state transitions (`wf_step_fail_remediation` -> `wf_step_recheck` -> `wf_step_done`), and `fest next` routing that drives the remediation loop. The audit trail is honest — the gate is logged as failed, never silently approved or skipped. Test coverage is excellent across every layer: state transitions, event materialization (including the full fail->recheck->done sequence), command validation, routing, fail-closed on an unreadable store, and a real end-to-end loop. Shared helpers are reused rather than reimplemented, ctx is threaded with `ctx.Err()` checks, and `festerrors` is used consistently (the diff also cleans up pre-existing `fmt.Errorf` in reject.go — nice).

**Key findings (see inline):**
- **(Gating) `fest next` now mutates state, including in pipe/machine modes.** The recheck transition is recorded before rendering, so `fest next --cd` / `--path` / `--json` fire a `wf_step_recheck` event and advance the gate. Before this PR the `next` package had zero state writes — it is a pure report command and `--path`/`--cd` are documented pipe helpers. A read/pipe query with a hidden state transition is a real correctness hazard for fest's agent-loop use case (inspect != act; `cd "$(fest next --cd)"` consumes the recheck). Suggest suppressing the transition in `isMachineOutput()` modes and recording it only on an explicit operator action. Want your call here.
- **(Robustness) Stale remediation-phase pointer can dead-end `fest next`.** The link is a phase-name string; if that phase is renamed/renumbered (a first-class fest op, incl. #189's `workflow renumber`), routing returns a hard NotFound on every `fest next` with no obvious escape. Surface a recovery hint and/or repair the link on renumber.
- **(Efficiency) Redundant store loads** — the event log is materialized 3–4x per `fest next` when a gate is active. Load once and thread it.
- **(Nit) Dead `if handled` guard** in next.go can swallow a future error; `gate` is already non-nil, so return unconditionally.

**What's done well.** The non-terminal-status design is exactly right — it makes "gate failed but being fixed" a first-class, auditable state instead of a lie. Fail-closed-on-load-error is the correct safety posture and is tested. Machine-output suppression of the human banner is handled and tested. The methodology doc clearly disambiguates reject vs skip vs failed-gate remediation.

**Staff standard.** The engineering quality is staff-level and I'd be proud of most of this. I can't approve as-is only because of the side-effect-in-a-read-query concern — that one genuinely affects correctness in the agent loop. Resolve (or consciously decide) that and it's a clean ship.

internal/commands/next/remediation.go
obey-agent: Addressed in e398916. Failed-remediation routing now resolves the linked remediation phase up front and returns an actionable `NotFound` when the stored phase name is stale, including restore/rename guidance and the `fest workflow reject --phase ... --reason` escape hatch. Added a regression for the stale pointer case.

internal/commands/next/next.go
obey-agent: Addressed in e398916. With a non-nil gate, `runNext` now returns the remediation route result directly, so there is no dead `handled` guard that could drop a future error.

internal/commands/next/remediation.go
obey-agent: Addressed in e398916. Machine-output recheck routes now only emit the read-only view and do not record `wf_step_recheck`; human/default `fest next` is the path that records the recheck transition. Added regressions for JSON staying non-mutating and human output clearing the failed-remediation state.

internal/commands/next/remediation.go
obey-agent: Addressed in e398916. `runNext` now loads the progress store once for failed-remediation routing, finds the failed gate from that loaded store, and threads the store through completion checks, workflow/task routing, and the recheck navigator. The remediation workflow renderer can now reuse the loaded store too.

obey-agent (commented):
## Staff Re-Review — Approve

**Verdict: Approve**

**Overview.** Re-reviewed the delta in `fix: harden failed gate remediation routing` (e398916) against my prior round. The gating concern and both secondary findings are resolved. Verified locally: clean `go build ./...` and passing tests across `internal/commands/next`, `internal/commands/workflow`, `internal/guidance/workflow`, `internal/progress`, and `internal/e2e`.

**Prior findings — resolved:**
- **(Was gating) `fest next` mutating state in pipe modes** — fixed exactly right. `renderRecheckRoute` now returns the machine view *before* any state change; the `nav.Recheck` transition fires only in human mode, explicitly framed as "human `fest next` is the operator action; pipe modes stay read-only." Two new tests lock this in: `RecheckMachineOutputDoesNotMutateState` (gate still failed-remediation after `--json`) and `RecheckHumanOutputRecordsSideEffect` (human mode clears it). `cd "$(fest next --cd)"` is now side-effect-free.
- **Stale remediation-phase pointer** — `resolveRemediationPhasePath` now returns an actionable error when the linked phase is gone, with a recovery hint (`fest workflow reject --phase <gate> --reason "..."`). I verified `--phase` is a real persistent flag honored by `getWorkflowNavigator` (Priority 1), and that rejecting the gate phase clears the `failed_with_remediation` state — so the hint actually works. `StaleRemediationPhasePointerIsActionable` covers it.
- **Redundant store loads** — the store is now loaded once in `runNext` (`loadProgressStore`) and threaded through `findFailedRemediationGateInStore` → `routeFailedRemediationGate` → completeness/route/recheck helpers and `runWorkflowModeWithStore`. The old standalone helpers remain as thin wrappers for test ergonomics. Hot path went from 3–4 materializations to one.
- **Dead `handled` branch** — gone; `runNext` now `return err` unconditionally since the gate is guaranteed non-nil.

**What's done well.** The read-only-vs-action split is a clean resolution that preserves the no-new-verb design while removing the correctness hazard. Store injection via `*WithStore` variants is a tidy DI pattern. Error hints are specific and verified-correct, not hand-wavy. Test discipline is excellent — behavior was re-specified, not just patched to pass.

**Staff standard.** Yes. The engineering here is staff-level and the audit-trail honesty is the right call. Ship it.

PR #191, opened 2026-05-30
Fix create festival agent rollback

lancekrogers:
## Summary

Fixes `fest create festival --agent` so strict agent-mode failures no longer leave behind a created festival directory that looks failed to the caller but exists on disk.

## Root Cause

`--agent` reused the normal human-oriented create pipeline: render files first, run validation later, then return `ok: false` if validation blocked. The JSON response did not disclose the created path or marker details, so agent callers could reasonably treat the failure as side-effect-free and retry, producing duplicate festival IDs.

## Changes

- Roll back the created festival directory on blocking `--agent` validation failures.
- Delay registry and campaign contract writes until after validation succeeds.
- Return structured failure JSON with `created_path`, `markers_unfilled`, `unfilled_markers`, `validation`, `rolled_back`, and rollback errors when relevant.
- Add reusable detailed template marker scanning with file, line, marker type, and marker content.
- Update `fest create festival --agent` help text.
- Add a containerized regression test proving rollback, structured marker details, and retry ID reuse.

## Validation

- `go test ./internal/validator ./internal/commands/festival`
- `just test unit-raw`
- `TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run TestCreateFestivalAgentValidationFailureRollsBackAndReportsMarkers -timeout 5m ./tests/integration`
- `git diff --check`

### DIFF
--- a/internal/commands/festival/create_festival.go
+++ b/internal/commands/festival/create_festival.go
@@ -267,6 +267,7 @@ type createResult struct {
 	markersTotal      int
 	allMarkers        []map[string]any
 	validationResult  *ValidationSummary
+	registered        bool
 }
 
 // RunCreateFestival executes the create festival command logic.
@@ -320,7 +321,7 @@ func RunCreateFestival(ctx context.Context, opts *CreateFestivalOptions) error {
 		return emitCreateFestivalCreatedError(ctx, cfg, res, err)
 	}
 
-	registerFestival(ctx, cfg)
+	registerFestival(ctx, cfg, res)
 
 	// Ensure fest contract entries are present in .campaign/watchers.yaml.
 	// This is idempotent: if fest init already wrote them, WriteEntries
@@ -647,7 +648,9 @@ func recordInitialSize(ctx context.Context, cfg *createConfig, festConfig *confi
 }
 
 // registerFestival records the festival in the ID registry with event logging.
-func registerFestival(ctx context.Context, cfg *createConfig) {
+// On success it marks res.registered so rollback only cleans up the registry
+// when a write actually happened.
+func registerFestival(ctx context.Context, cfg *createConfig, res *createResult) {
 	regPath := registry.GetEventsPath(cfg.festivalsRoot)
 	reg, regErr := registry.Load(ctx, regPath)
 	if regErr != nil {
@@ -662,7 +665,9 @@ func registerFestival(ctx context.Context, cfg *createConfig) {
 		CreatedAt: now,
 		UpdatedAt: now,
 	}
-	_ = reg.AddWithEvent(ctx, regEntry)
+	if err := reg.AddWithEvent(ctx, regEntry); err == nil && res != nil {
+		res.registered = true
+	}
 }
 
 // processAllMarkers processes REPLACE markers in all created files.
@@ -917,13 +922,15 @@ func rollbackCreatedFestival(ctx context.Context, cfg *createConfig, res *create
 		}
 	}
 
-	regPath := registry.GetEventsPath(cfg.festivalsRoot)
-	reg, err := registry.Load(ctx, regPath)
-	if err != nil {
-		failures = append(failures, fmt.Sprintf("loading registry for rollback: %v", err))
-	} else if reg.Exists(ctx, cfg.festivalID) {
-		if err := reg.DeleteWithEvent(ctx, cfg.festivalID); err != nil {
-			failures = append(failures, fmt.Sprintf("deleting registry entry: %v", err))
+	if res != nil && res.registered {
+		regPath := registry.GetEventsPath(cfg.festivalsRoot)
+		reg, err := registry.Load(ctx, regPath)
+		if err != nil {
+			failures = append(failures, fmt.Sprintf("loading registry for rollback: %v", err))
+		} else if reg.Exists(ctx, cfg.festivalID) {
+			if err := reg.DeleteWithEvent(ctx, cfg.festivalID); err != nil {
+				failures = append(failures, fmt.Sprintf("deleting registry entry: %v", err))
+			}
 		}
 	}
 
--- a/internal/commands/festival/create_festival.go
+++ b/internal/commands/festival/create_festival.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 	tpl "github.com/Obedience-Corp/fest/internal/template"
 	"github.com/Obedience-Corp/fest/internal/types"
 	"github.com/Obedience-Corp/fest/internal/ui"
+	"github.com/Obedience-Corp/fest/internal/validator"
 	"github.com/Obedience-Corp/fest/internal/workspace"
 	"github.com/Obedience-Corp/obey-shared/contract"
 	"github.com/google/uuid"
@@ -51,6 +53,7 @@ type createFestivalResult struct {
 	OK                bool               `json:"ok"`
 	Action            string             `json:"action"`
 	Festival          map[string]string  `json:"festival,omitempty"`
+	CreatedPath       string             `json:"created_path,omitempty"`
 	Created           []string           `json:"created,omitempty"`
 	AutoPhasesCreated []string           `json:"auto_phases_created,omitempty"`
 	GatesDirectory    string             `json:"gates_directory,omitempty"`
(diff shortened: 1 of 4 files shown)


obey-agent (commented):
**Verdict: Approve**

I re-reviewed the PR after reading the changed files in full. The agent-mode rollback flow is now ordered correctly: scaffold/render happens first, strict validation runs before registry and contract writes, and blocking agent validation failures remove the created festival directory while returning structured JSON that includes marker details and rollback status.

I did not find any blocking correctness issues. The new integration coverage exercises the failure, rollback, marker reporting, and retry/ID reuse path directly.

Validation run locally:
- `go test ./internal/validator ./internal/commands/festival`
- `just test unit-raw`
- `TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run TestCreateFestivalAgentValidationFailureRollsBackAndReportsMarkers -timeout 5m ./tests/integration`
- `go vet ./...`

internal/commands/festival/create_festival.go
```diff
@@ -815,6 +834,233 @@ func emitCreateJSON(cfg *createConfig, res *createResult, gatesDir string, remai
...
+			failures = append(failures, fmt.Sprintf("deleting registry entry: %v", err))
+		}
+	}
+
+	if err := os.RemoveAll(cfg.destDir); err != nil {
+		failures = append(failures, fmt.Sprintf("removing created festival directory: %v", err))
+	}
+
+	if len(failures) > 0 {
+		return fmt.Errorf("%s", strings.Join(failures, "; "))
```
obey-agent: **Use the project error package, not `fmt.Errorf`.**

`rollbackCreatedFestival` builds its aggregate error with the stdlib:

```go
return fmt.Errorf("%s", strings.Join(failures, "; "))
```

Repo standard (CLAUDE.md, zero-tolerance) is to never use `fmt.Errorf` in production code — always go through `internal/errors`. This matters here specifically because this error becomes the `rollback_error` field an agent reads: routing it through the project package keeps the code/op/field metadata and JSON marshaling consistent with every other error surface in this file.

Swap to:

```go
return errors.New(strings.Join(failures, "; "))
```

(`errors` is already imported as the internal package in this file.) The inner `fmt.Sprintf` calls that build the individual `failures` fragments are fine — those are message strings, not errors.

obey-agent (commented):

### Verdict
**Request Changes** — one zero-tolerance standard violation, otherwise strong work.

### Overview
This is a well-scoped, well-reasoned fix. The root cause analysis is correct: `--agent` reused the human create pipeline, registered the festival and wrote contract entries *before* validation, and on a blocking failure left a created-but-failed directory on disk while returning a side-effect-free-looking JSON error. The fix reorders the pipeline so registry/contract writes happen only after validation succeeds, rolls back the directory (plus nav link and any registry entry) on blocking agent-mode failures, and returns structured failure JSON with `created_path`, marker detail, and `rolled_back`. I traced the execution paths and verified the design holds.

### What I Verified
- **ID reuse on retry is real.** `id.FindNextCounter` scans festival *directories* (not the registry), so removing `destDir` in rollback genuinely frees the counter. The test's `AR0001`-reuse assertion is sound, not coincidental.
- **Reordering is safe.** Registration/contract writes moved after validation; nothing between (auto-phases, size recording) depends on registration. Build passes, `go vet` clean, `go test ./internal/validator ./internal/commands/festival` passes.
- **Rollback ordering is correct.** Markers are scanned while the directory still exists, then rollback runs. Nil-guard accessors (`festConfigPath`, `markersFilled`, etc.) are justified because `res` is legitimately nil on early render/yaml failures.
- **Marker scan refactor is a genuine consolidation,** not a parallel implementation. `ValidateTemplateMarkers` and `CheckTemplatesFilled` now both flow through `ScanTemplateMarkers`, deleting the duplicated walk. Good DRY win.

### Required Change (blocking)
- `rollbackCreatedFestival` builds its aggregate error with `fmt.Errorf("%s", ...)` (create_festival.go:935). Repo standard is never `fmt.Errorf` in production code — use `errors.New(...)` from the already-imported internal package. This is the only thing gating the merge; it is a one-line fix. Inline comment posted.

### Non-blocking notes (your call, don't need a re-review)
- `scanTemplateMarkersInContent`: the inner `for _, content := range ...` loop variable shadows the `content` function parameter. Harmless today but a readable rename (`markerText`) avoids a future footgun.
- `create_festival.go` is now 1141 lines (was already well over the 500 guideline before this PR). Not introduced here and not worth blocking a bugfix, but the new failure/rollback emit helpers are a clean seam — extracting them to `create_festival_failure.go` in a follow-up would start paying down the file.
- `extractMarkerContents` truncates on pathologically nested markers (e.g. `[FILL: [FILL: x]`), under-counting vs the old `strings.Count`. Not realistic in templates; flagging only for completeness.

### What's Done Well
Containerized integration test (correct per the filesystem-safety rule), exercising the full loop: rollback, structured marker detail with file/line/type, empty-planning-dir assertion, and retry ID reuse. The PR description's root-cause writeup is exactly the altitude I want. Error wrapping elsewhere in the new code consistently uses the internal package and propagates `context.Context`.

### Staff Standard
Would I submit this as my own work? Yes, after the one-line `fmt.Errorf` fix. The thinking is sound, the test proves the behavior, and it improved existing code rather than forking a parallel path. Fix that line and this is a clean approve.

internal/commands/festival/create_festival.go
```diff
@@ -815,6 +834,233 @@ func emitCreateJSON(cfg *createConfig, res *createResult, gatesDir string, remai
...
+			failures = append(failures, fmt.Sprintf("removing navigation link: %v", err))
+		} else if nav.RemoveLink(cfg.dirName) {
+			if err := nav.Save(); err != nil {
+				failures = append(failures, fmt.Sprintf("saving navigation rollback: %v", err))
+			}
+		}
+	}
+
+	regPath := registry.GetEventsPath(cfg.festivalsRoot)
+	reg, err := registry.Load(ctx, regPath)
```
obey-agent: `registry.Load` is not a read-only operation in legacy workspaces: `Load` migrates `id_registry.yaml` to `festival_events.jsonl` and removes the legacy file on access. Because this rollback path runs for the agent validation failure case after registration has been delayed, a failed `fest create festival --agent` can still mutate registry state even though the PR goal says registry writes wait until validation succeeds.

Please make this rollback path avoid the convert-on-access load unless a registry write actually happened. For example, track whether this create registered the festival, or skip registry rollback when only the validation-failure path can reach it. If you still need defensive cleanup, check for an existing events file before loading so legacy YAML is not migrated on a failed create.

obey-agent (commented):
## Staff Re-review — Request Changes

### Verdict
**Request Changes** — the previous `fmt.Errorf` blocker is fixed, but there is one remaining rollback side effect in legacy registry workspaces.

### Overview
I re-reviewed PR #191 at head `d041aab9aca6d72e96fecc7a0cfe482993c9d246` after the latest push. The main flow is still strong: agent validation now happens before registration/contract writes, validation failures roll back the created directory, and the structured JSON failure response includes created path, validation, marker detail, and rollback status. The new integration test continues to prove rollback, marker reporting, and retry ID reuse.

### Key Finding
- Inline comment posted on `internal/commands/festival/create_festival.go`: `rollbackCreatedFestival` still calls `registry.Load`, and `registry.Load` is convert-on-access for legacy `id_registry.yaml` files. That means a failed `fest create festival --agent` can still mutate registry state by migrating/deleting the legacy registry even though this PR's contract is that registry writes are delayed until validation succeeds. Please avoid that load on validation-failure rollback unless this create actually registered the festival, or guard it so legacy migration is not triggered by a failed create.

### What's Done Well
- The previous internal error-package issue is resolved with `errors.New(...)`.
- The marker scanner variable shadowing cleanup is good.
- The rollback test is correctly containerized and exercises the important user-visible behavior: nonzero agent failure, JSON marker details, directory removal, empty planning dir, retry success, and ID reuse.

### Validation Run
- `go test ./internal/validator ./internal/commands/festival`
- `go vet ./...`
- `just test unit-raw`
- `TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run TestCreateFestivalAgentValidationFailureRollsBackAndReportsMarkers -timeout 5m ./tests/integration`

### Staff Standard
Would I submit this as my own work? Yes after the registry rollback load is tightened. The implementation is close; this remaining issue is about preserving the no-registry-side-effects guarantee for older workspaces.

obey-agent (commented):
## Staff Re-review (round 3) — Approve

### Verdict
**Approve.** Both previously-blocking issues are resolved and verified in the code at head `19b282e`. No new blocking issues.

### What I verified this round
I re-read the full changed files at head and traced the create→validate→rollback path; `go build ./...` and `go vet ./internal/commands/festival ./internal/validator` are clean.

- **Blocker 1 (`fmt.Errorf` in rollback) — fixed.** `rollbackCreatedFestival` now aggregates failures via `errors.New(strings.Join(...))` (create_festival.go:942). No `fmt.Errorf` remains in the file; the surviving `fmt.Sprintf` calls only build message fragments, which is correct.

- **Blocker 2 (registry migration on failed agent create) — fixed correctly.** The fix is sound at the root, not patched at the symptom:
  - `registerFestival` is called only *after* `validateIfConfigured` succeeds (create_festival.go:324), and now records `res.registered = true` only on a successful `AddWithEvent` (create_festival.go:668).
  - `rollbackCreatedFestival` guards the `registry.Load`/`DeleteWithEvent` block behind `if res != nil && res.registered` (create_festival.go:925). On a validation failure, registration never ran, so `registered` is false and `registry.Load` is never called — the convert-on-access migration of a legacy `id_registry.yaml` can no longer be triggered by a failed create.
  - Because ID generation scans festival directories (not the registry), nothing earlier in the pipeline loads the registry either, so a failed `--agent` create is genuinely registry-side-effect-free.

### Tests
- `TestCreateFestivalAgentValidationFailureLeavesLegacyRegistryIntact` is a precise regression test for blocker 2: it seeds a legacy `id_registry.yaml`, runs a failing `--agent` create, and asserts the legacy YAML survives and no `festival_events.jsonl` is produced. This would fail if any pre-validation code path loaded the registry, so it pins the guarantee.
- `TestCreateFestivalAgentValidationFailureRollsBackAndReportsMarkers` continues to cover the full loop: non-zero exit, structured JSON marker detail, directory removal, empty planning dir, and ID reuse (`AR0001`) on retry.
- Both are containerized, consistent with the filesystem-safety rule.

### Non-blocking (already acknowledged in prior rounds, not gating)
- `create_festival.go` remains over the 500-line guideline; the new failure/rollback emit helpers are a clean future extraction seam.
- `extractMarkerContents` undercounts pathologically nested markers vs the old `strings.Count`; not reachable from real templates and does not affect the filled/unfilled decision.

### Staff Standard
Would I submit this as my own work? Yes. The pipeline reordering, the registration-tracking guard, and the targeted regression tests address the root cause cleanly and improve the existing code rather than forking a parallel path. Clean approve.

PR #193, opened 2026-06-04
fix(chain): use CH namespace for chain IDs

lancekrogers:
## Summary
- generate festival chain IDs from the fixed `CH` namespace instead of deriving prefixes from chain names
- update chain creation tests to assert the `CH####` convention
- keep existing chain parsing compatible with already-created non-CH files

## Verification
- `go test ./internal/commands/chain ./internal/chain`
- `just test unit`
- `just lint gopls-tests internal/commands/chain`
- `just lint vet`
- `just build quick`

### DIFF
--- a/internal/commands/chain/create.go
+++ b/internal/commands/chain/create.go
@@ -16,6 +16,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const chainIDPrefix = "CH"
+
 func newCreateCmd() *cobra.Command {
 	var (
 		name string
@@ -59,10 +61,9 @@ func runCreate(cmd *cobra.Command, name, goal string) error {
 		return errors.IO("creating chains directory", err)
 	}
 
-	// Generate chain ID from name prefix.
+	// Generate chain IDs from the fixed chain namespace.
 	slug := strings.ToLower(strings.ReplaceAll(name, " ", "-"))
-	prefix := chainIDPrefix(name)
-	id := nextChainID(chainsDir, prefix)
+	id := nextChainID(chainsDir, chainIDPrefix)
 
 	// Render embedded chain template.
 	tplData, err := chaintpl.Templates.ReadFile("chain_template.yaml")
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
## Verdict: Approve

### Overview
Clean, tightly-scoped fix that standardizes festival chain IDs onto a fixed `CH` namespace instead of deriving an arbitrary 2-letter prefix from the chain name. Intent is clear, execution is correct, and it's well covered by tests. Verified against head `6ec591e` with `go test ./internal/commands/chain/` passing.

### Key Findings
- **Correctness verified**: replacing `chainIDPrefix(name)` (which produced non-deterministic prefixes like `AB`/`XY` per name-family) with `const chainIDPrefix = "CH"` gives chains a single stable namespace consistent with the wider `FE####` ID convention. `nextChainID` now scans one shared counter, which is the intended behavior.
- **Backward compatibility holds**: chain discovery reads the `id:` field from inside each YAML (`internal/chain/discovery.go`), not from filenames, so existing non-`CH` chains still load. `nextChainID` safely ignores them since slugs are lowercased (uppercase `CH` only ever matches the ID suffix) and `O_EXCL` guards against filename collisions regardless.
- **Magic string removed**: `"CH"` is now a named constant rather than a buried literal.
- Minor (non-blocking): `TestChainIDPrefixUsesChainNamespace` asserts `chainIDPrefix == "CH"`, which is tautological. Harmless; not worth a change on its own.

### What's Done Well
- `ctx.Err()` checked up front; `internal/errors` used consistently with proper codes.
- No parallel implementation — the existing generation path was edited in place and dead code (`chainIDPrefix(name)`) removed cleanly, with `slug` still correctly used for filename and template `Name`.
- Tests updated to assert the new `CH####` convention, including a case proving other-namespace files don't bleed into the counter.

### Staff Standard
Yes — I'd be comfortable submitting this as my own work. It's small, correct, leaves the codebase cleaner than it found it, and is trivially maintainable by a less-experienced engineer.

PR #194, opened 2026-06-05
fix(chain,watch): dungeon-aware chain resolution + watch completion; chain context ergonomics

lancekrogers:
## Summary

Fixes two dungeon-correctness bugs in `fest` chain/watch handling and adds chain-command context ergonomics. Design workitem: `obey-campaign:workflow/design/fest-chain-and-watch-dungeon-2026-06-05/`.

Three commits, one per concern, reviewable independently.

## A — `fix(chain)`: resolve dungeoned festivals in chain status + promote gate

**Bug:** `fest promote` (ready → active) falsely blocked on a completed upstream festival that had been moved to `dungeon/completed/<date>/`, demanding `--force`.

`findFestivalDir` (`internal/chain/resolve.go`) scanned only one directory level, so festivals inside a dated dungeon bucket (`dungeon/<status>/YYYY-MM-DD/<festival>`) were never found. The promote chain-dependency gate, `fest next` chain context, and `fest chain status` all resolve festival status through this path, so a dungeoned-completed upstream read as `planning` → incomplete.

- `findFestivalDir` now descends one level into `YYYY-MM-DD` buckets for dungeon status dirs (mirrors the existing `FindNextRitualRun` dungeon-descent idiom).
- Added `ResolveAvailable`, a best-effort sibling of `Resolve` that omits unresolved refs instead of nil-ing the whole map (mirrors `DiscoverAll` vs `DiscoverAllStrict`), so one missing chain member no longer blanks the rest of the chain. `promote`, `next`, and `chain` live status use it.

Tests: dungeon-bucket resolution (strict + best-effort), non-date subdir is not descended, best-effort omits missing refs, and gate-level tests proving a dungeon-completed upstream does **not** block while an active upstream still does.

## C — `fix(watch)`: exclude dungeon festivals from completion and picker

**Bug:** `fest watch` tab-completion (and the interactive picker) offered dungeon festivals. Dungeon is terminal — nothing there is watched.

The watch enumeration passed no `PreferredStatuses`, so candidate collection fell back to the full `StatusDirectories` set (which includes `dungeon/*`).

- Completion now passes `id.WorkingStatusDirectories` → only `planning/ready/active/ritual`.
- The picker is constrained the same way: `preferredPickerStatuses` no longer derives a dungeon substatus, and `pickerStatuses` falls back to working statuses.
- The shared default and `fgo` dungeon navigation are untouched — this is a watch-local constraint.

Tests: a real disk-level test proving working-status collection excludes a dungeon festival while the default surfaces it; picker-status tests. The prior test that asserted dungeon festivals were *included* is replaced.

## B — `feat(chain)`: infer festival/chain context + interactive picker

Chain subcommands required the chain id (or festival ref) explicitly even when the festival context was already knowable — unlike `fest next/commit/watch`.

- `chain status/graph/validate/complete` now take an optional `[chain-id]`; when omitted they infer it from the current festival (a `fest.yaml` ancestor of the cwd, or a linked project dir via the navigation registry) and its chain via `FindForFestival`. `chain check` infers the festival ref/id the same way.
- When no context is available and stdout is a TTY, an interactive chain picker (reusing the shared bubbletea picker) lets the user select a non-terminal chain. In a non-interactive shell the previous missing-argument behavior is preserved, so scripts and agents still fail fast.
- `chain complete` is mutating, so an inferred or picked target must be confirmed, and a non-interactive run refuses an inferred target and asks for an explicit id — a chain is never archived on a guessed target without consent.

Tests: marker/walk-up festival detection, explicit-id passthrough, current-festival resolution, and the non-interactive picker/confirm/no-context paths.

## End-to-end verification

`fest chain status` with **no argument**, run from inside a `ready` festival whose only hard upstream is completed-and-dungeoned:

```
$ cd festivals/ready/onboarding-FA0010 && fest chain status
Chain: onboarding-readiness (CH0001)        # ← inferred, no chain-id passed (B)
Status: active

Edges:
  audit --hard--> onboarding

Progress: ███████████████░░░░░░░░░░░░░░░ [1/2] 50%   # ← dungeoned FA0009 resolves as completed (A)

Unblocked:
  onboarding (FA0010) onboarding-parity     # ← FA0010 unblocked: its hard dep is satisfied
```

Before this change the same command required an explicit chain id, and FA0009 resolved as `planning`, leaving FA0010 blocked (the same false negative that made `fest promote` demand `--force`).

## Gates

- `go build ./...`: clean
- `go vet ./...`: clean
- `go test ./...`: 90 packages pass
- `just lint all` (golangci-lint): 0 issues

### DIFF
--- a/internal/chain/resolve.go
+++ b/internal/chain/resolve.go
@@ -36,6 +36,32 @@ func Resolve(ctx context.Context, c *Chain, searchDirs []string) (map[string]Res
 	return resolved, nil
 }
 
+// ResolveAvailable resolves festival references best-effort: refs whose
+// directory cannot be found are omitted from the result rather than failing the
+// entire resolution (as Resolve does). Use this when partial chain state is
+// acceptable — computing live statuses for display or dependency gates, where
+// one missing or not-yet-created festival should not blank out the rest of the
+// chain. Use Resolve when every ref must resolve (e.g. chain validation).
+func ResolveAvailable(ctx context.Context, c *Chain, searchDirs []string) (map[string]ResolvedFestival, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	resolved := make(map[string]ResolvedFestival, len(c.Festivals))
+	for _, node := range c.Festivals {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		path, err := findFestivalDir(ctx, node.ID, searchDirs)
+		if err != nil {
+			continue // best-effort: omit unresolved refs
+		}
+		resolved[node.Ref] = ResolvedFestival{Node: node, Path: path}
+	}
+
+	return resolved, nil
+}
+
 // findFestivalDir searches the given directories for a subdirectory whose name
 // contains the festival ID (e.g., "hedera-foundation-HF0001" contains "HF0001").
 func findFestivalDir(ctx context.Context, festivalID string, searchDirs []string) (string, error) {
@@ -48,6 +74,7 @@ func findFestivalDir(ctx context.Context, festivalID string, searchDirs []string
 		if err != nil {
 			continue // skip unreadable directories
 		}
+		dungeon := isDungeonStatusDir(dir)
 		for _, entry := range entries {
 			if !entry.IsDir() {
 				continue
@@ -55,10 +82,56 @@ func findFestivalDir(ctx context.Context, festivalID string, searchDirs []string
 			if matchesFestivalID(entry.Name(), festivalID) {
 				return filepath.Join(dir, entry.Name()), nil
 			}
+			// Dungeon status dirs interpose a dated bucket between the status
+			// dir and the festival (dungeon/<status>/YYYY-MM-DD/<festival>), so
+			// descend one level into date buckets to match festivals there.
+			if dungeon && isDateBucket(entry.Name()) {
+				if path, ok := matchFestivalInDir(filepath.Join(dir, entry.Name()), festivalID); ok {
+					return path, nil
+				}
+			}
 		}
 	}
 
 	return "", errors.NotFound("festival").WithField("festivalID", festivalID)
+}
+
+// matchFestivalInDir returns the path of a festival directory matching
+// festivalID located directly inside dir, if present.
+func matchFestivalInDir(dir, festivalID string) (string, bool) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return "", false
+	}
+	for _, entry := range entries {
+		if entry.IsDir() && matchesFestivalID(entry.Name(), festivalID) {
+			return filepath.Join(dir, entry.Name()), true
+		}
+	}
+	return "", false
+}
+
+// isDungeonStatusDir reports whether dir is a dungeon status directory
+// (festivals/dungeon/{completed,archived,someday}), which holds festivals one
+// level deeper inside dated YYYY-MM-DD buckets.
+func isDungeonStatusDir(dir string) bool {
+	return filepath.Base(filepath.Dir(dir)) == "dungeon"
+}
+
+// isDateBucket reports whether name is a YYYY-MM-DD dungeon bucket directory.
+func isDateBucket(name string) bool {
+	if len(name) != 10 || name[4] != '-' || name[7] != '-' {
+		return false
+	}
+	for i := 0; i < len(name); i++ {
+		if i == 4 || i == 7 {
+			continue
+		}
+		if name[i] < '0' || name[i] > '9' {
+			return false
+		}
+	}
+	return true
 }
 
 // matchesFestivalID checks whether a directory name matches a festival ID.
--- a/internal/chain/resolve.go
+++ b/internal/chain/resolve.go
@@ -36,12 +36,9 @@ func Resolve(ctx context.Context, c *Chain, searchDirs []string) (map[string]Res
 	return resolved, nil
 }
 
-// ResolveAvailable resolves festival references best-effort: refs whose
-// directory cannot be found are omitted from the result rather than failing the
-// entire resolution (as Resolve does). Use this when partial chain state is
-// acceptable — computing live statuses for display or dependency gates, where
-// one missing or not-yet-created festival should not blank out the rest of the
-// chain. Use Resolve when every ref must resolve (e.g. chain validation).
+// ResolveAvailable is like Resolve but best-effort: unresolved refs are omitted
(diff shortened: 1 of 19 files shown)


internal/commands/chain/context.go
```diff
@@ -0,0 +1,173 @@
...
+// as-is. Otherwise the current festival context is used to find its chain;
+// failing that, an interactive picker is shown in a TTY. inferred reports
+// whether the id was inferred or picked rather than passed explicitly, so
+// mutating commands can confirm before acting.
+func resolveChainID(ctx context.Context, explicit string) (chainID string, inferred bool, err error) {
+	if explicit != "" {
+		return explicit, false, nil
+	}
+
+	root, err := festivalsRoot()
```
obey-agent, line 89: This resolves the festivals root from the current working directory before trying `resolveCurrentFestivalID`, which means the advertised linked-project path cannot work from a normal `projects/<repo>` checkout. `tpl.FindFestivalsRoot` only succeeds when cwd is inside the `festivals/` tree, so from a linked project it returns before `festivalIDFromLink` can load navigation and find the linked festival. The chain context helper needs to derive the root through campaign/workspace detection, or resolve the current festival/link first and use that festival path to find the festivals root. Please add a regression test for `resolveChainID` from a linked project cwd.

internal/chain/resolve.go
```diff
@@ -48,19 +74,66 @@ func findFestivalDir(ctx context.Context, festivalID string, searchDirs []string
...
+
+// isDungeonStatusDir reports whether dir is a dungeon status directory
+// (festivals/dungeon/{completed,archived,someday}), which holds festivals one
+// level deeper inside dated YYYY-MM-DD buckets.
+func isDungeonStatusDir(dir string) bool {
+	return filepath.Base(filepath.Dir(dir)) == "dungeon"
+}
+
+// isDateBucket reports whether name is a YYYY-MM-DD dungeon bucket directory.
+func isDateBucket(name string) bool {
```
obey-agent, line 113: This new chain-specific date predicate only accepts `YYYY-MM-DD`, but the rest of fest still treats `YYYY-MM` as a valid legacy dungeon bucket (`show.LooksLikeDateDir` / status helpers explicitly support both). A completed upstream in `dungeon/completed/2025-01/<festival>` will still resolve as missing/planning, so this fix is incomplete for existing campaigns. Please reuse the shared date-bucket predicate or extend this test/helper to cover the legacy `YYYY-MM` case.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This PR addresses the right chain/watch failures and the main flow is well covered by targeted tests plus a full `go test ./...` pass. The dungeon-completed chain gate fix is directionally correct, but two context/correctness gaps remain before I would merge it.

**Key Findings:**
- `internal/commands/chain/context.go`: linked-project chain inference currently fails before it reaches navigation lookup because `resolveChainID` calls `festivalsRoot()` from the project cwd first.
- `internal/chain/resolve.go`: the new date bucket predicate only accepts `YYYY-MM-DD`, while fest still supports legacy `YYYY-MM` dungeon buckets elsewhere.

**What's Done Well:** The PR keeps strict and best-effort chain resolution separate, preserves non-interactive fast-fail behavior for no-context commands, and adds focused regression tests around the FA0009/FA0010 completed-upstream case.

**Staff Standard:** Not yet. I would be proud of the direction and test shape, but I would not submit it as complete until linked-project inference actually works and dungeon bucket parsing stays consistent with the rest of fest.

Verification run locally: `go test ./...` passed; `git diff --check` passed.

lancekrogers:
Both findings fixed in `b0a8aef`.

**1. Linked-project inference failed before navigation lookup.** You were right, and the root cause was a bit deeper than just `resolveChainID`: resolving the festivals root from the cwd via `tpl.FindFestivalsRoot` only succeeds when cwd is already inside `festivals/`, so it failed not only in `resolveChainID` but also in the downstream lookups (`findChainByID`, chain live-status, `chain create`/`list`) — the id would get inferred and then resolution would still error. Fix: `festivalsRoot()` now falls back to `workspace.FindFestivals` (resolves from anywhere in the campaign, incl. a linked project dir or the campaign root), and all the chain-package root lookups route through it instead of calling `tpl.FindFestivalsRoot(cwd)` directly.

Verified end-to-end (not just unit-tested): `fest chain status` with **no arg**, run from inside a linked project working directory, now infers `CH0001` and resolves statuses:

```
$ cd campaign/projects/demo && fest chain status     # linked to FA0010
Chain: onboarding (CH0001)
Progress: ███████████████░░░░░░░░░░░░░░░ [1/2] 50%
Unblocked: onboarding (FA0010)
```

**2. Legacy `YYYY-MM` dungeon buckets.** `isDateBucket` now matches both `YYYY-MM-DD` and the legacy `YYYY-MM` form, mirroring `show.LooksLikeDateDir`'s `^\d{4}-\d{2}(-\d{2})?$` (kept local to avoid the low-level `chain` package importing a `commands` package). The e2e above uses a `2026-05` bucket, so it exercises the legacy path.

Tests added: `YYYY-MM` bucket resolution + `isDateBucket` cases; `festivalsRoot` resolving from a linked project dir and from inside `festivals/`.

Also trimmed the over-verbose doc/inline comments this PR introduced down to fest's terse one-line style.

Gates: `go build ./...`, `go test ./...` (90 pkgs), `just lint all` (golangci-lint) — all green.

obey-agent (commented):
**Verdict: Approve**

**Overview:** Rereviewed `b0a8aef` after the fixes for the prior findings. The chain resolver now handles completed festivals in both `YYYY-MM-DD` and legacy `YYYY-MM` dungeon buckets, and chain commands now resolve the festivals root from project/campaign contexts before doing chain lookup.

**Key Findings:**
- No blocking findings in this pass. The previous linked-project inference and legacy dungeon bucket issues are fixed.

**What's Done Well:** The PR keeps strict resolution for contract validation while using best-effort resolution only for live status/progress surfaces. The command root resolver is now shared across the chain command set, and the added tests cover the previous FA0009/FA0010 completed-upstream regression plus watch picker/completion exclusion for terminal dungeon festivals.

**Staff Standard:** Yes. I would be comfortable submitting this as my own change after this revision.

**Verification:** `git diff --check 8690934494dd50464b907b8493c9382ae781a074..HEAD`, `go test ./...`, `go build ./...`, and `just lint all` all pass. I also built a temporary PR binary and verified no-arg `chain status` / `chain check` from the linked FA0010 worktree resolves `CH0001` and treats FA0009 as completed, while explicit `chain check festival_authoring_modal --chain CH0001` remains blocked on active FA0010.

PR #195, opened 2026-06-08
feat(chain): add 'fest chain add' to add festivals to chains (WI-72f9af)

lancekrogers:
## Summary

Adds `fest chain add`, a command to add a festival to an existing chain as a node, optionally with prerequisite dependency edges. Until now `fest chain create` only made an empty chain shell and printed "add festivals and edges by editing the chain file directly" — there was no command to populate a chain, so membership was hand-edited YAML. This closes that gap.

The command parses the target chain, mutates it in memory (append node + edges, place in a trailing wave when the chain uses waves), validates the result with the existing `chain.Validate`, and writes only if valid via a new atomic `chain.Write`.

Design: `workflow/design/fest-chain-add-member/` in the obey-campaign repo (workitem WI-72f9af), grounded against this source.

## Command surface

```
fest chain add
  --chain string      target chain id or name
  --festival string   festival to add (id, name, or path); defaults to current festival context
  --after stringArray prerequisite festival ref/id (optional, repeatable -> one edge each)
  --type string       edge type: hard | soft (default hard)
  --note string       optional edge note
  --json              emit structured JSON result
```

Creating a brand-new chain stays in `fest chain create`; this command adds to an existing chain.

## Behavior

- Validate-then-write: the mutated chain is validated in memory and written only if valid. An invalid mutation exits non-zero and writes nothing (no partial file).
- Atomic write: new `chain.Write` re-marshals the chain and replaces the file via a temp-file-plus-rename. Clean re-marshal (drops the create template comments); the chain stays valid and readable.
- Waves: chains without waves need no wave handling; chains with waves get the added festival placed in a trailing wave so the S7/S8 validators still pass.
- Rejections: duplicate festival, self-edge, unknown prerequisite, invalid edge type.
- `--festival` defaults to the current festival context; non-TTY requires explicit flags.

## JSON result schema

```json
{
  "chain_id": "CH0001",
  "chain_name": "demo-chain",
  "node": { "ref": "gamma", "id": "G0001", "name": "gamma", "projects": ["projects/gamma"] },
  "edges": [ { "from": "beta", "to": "gamma", "type": "hard", "note": "needs beta" } ],
  "validation": { "valid": true, "score": 100 }
}
```

`node.projects` omitted when unknown; `edges` always present (empty array with no `--after`); `edge.note` omitted when empty; `validation.errors` (string array) omitted when valid.

## Changes

- `internal/chain/write.go` (+test): `chain.Write` clean re-marshal, atomic.
- `internal/commands/chain/add.go`, `add_helpers.go`, `add_festival.go` (+tests): the command, pure helpers (ref slug/dedup, node/edge build, edge-type parse, trailing-wave placement), and `--festival` resolution.
- `internal/commands/chain/chain.go`: register `add`.
- `internal/chain/types.go`: `omitempty` on `Metadata.created_at` and `status_history` so re-marshal of chains lacking those fields stays clean (`chain.Write` is the only Chain marshaler).

## Testing

- `go build ./...` clean; `golangci-lint run ./...` 0 issues; `go vet` and `gofmt` clean.
- Full unit suite green (no regressions). New tests use host `t.TempDir` per the repo convention, covering: `Write` round-trip and atomicity; node/edge/ref construction, dedup, self-edge rejection; validate-then-write (valid writes, invalid writes nothing); no-waves and waved (trailing-wave) cases; `--json`; context cancellation.
- Real-binary smoke: no-waves add, repeatable `--after`, root node, waved trailing-wave placement (then `fest chain validate` VALID 100/100), and the rejection paths.

## Notes / follow-ups

- Interactive TTY prerequisite picker not included (design marked it optional for this PR); non-TTY explicit-flag behavior is implemented, and `--chain` reuses the existing chain picker.
- Removing a member (`fest chain remove`) is a separate future concern.
- Downstream: this unblocks the festival-app create-festival modal (FA0011), whose chain wiring calls `fest chain add` rather than editing chain YAML.

### DIFF
--- a/internal/chain/write.go
+++ b/internal/chain/write.go
@@ -1,0 +1,57 @@
+package chain
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/Obedience-Corp/fest/internal/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// Write marshals a Chain to YAML and writes it atomically to path. It renders a
+// clean document from the struct (dropping any explanatory comments the create
+// template added) and replaces the destination via a temp file in the same
+// directory followed by an atomic rename, so a failed write never leaves a
+// partial or corrupt chain file in place.
+func Write(path string, c *Chain) error {
+	if c == nil {
+		return errors.Validation("cannot write nil chain")
+	}
+
+	data, err := yaml.Marshal(c)
+	if err != nil {
+		return errors.Wrap(err, "marshaling chain").WithCode(errors.ErrCodeParse)
+	}
+
+	dir := filepath.Dir(path)
+	tmp, err := os.CreateTemp(dir, ".chain-*.yaml.tmp")
+	if err != nil {
+		return errors.IO("creating temp chain file", err).WithField("dir", dir)
+	}
+	tmpPath := tmp.Name()
+
+	cleanup := func() {
+		_ = tmp.Close()
+		_ = os.Remove(tmpPath)
+	}
+
+	if _, err := tmp.Write(data); err != nil {
+		cleanup()
+		return errors.IO("writing temp chain file", err).WithField("path", tmpPath)
+	}
+	if err := tmp.Sync(); err != nil {
+		cleanup()
+		return errors.IO("syncing temp chain file", err).WithField("path", tmpPath)
+	}
+	if err := tmp.Close(); err != nil {
+		_ = os.Remove(tmpPath)
+		return errors.IO("closing temp chain file", err).WithField("path", tmpPath)
+	}
+
+	if err := os.Rename(tmpPath, path); err != nil {
+		_ = os.Remove(tmpPath)
+		return errors.IO("replacing chain file", err).WithField("path", path)
+	}
+
+	return nil
+}
--- a/internal/chain/write.go
+++ b/internal/chain/write.go
@@ -1,24 +1,29 @@
 package chain
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 
 	"github.com/Obedience-Corp/fest/internal/errors"
-	"gopkg.in/yaml.v3"
+	"github.com/Obedience-Corp/fest/internal/yamlutil"
 )
 
 // Write marshals a Chain to YAML and writes it atomically to path. It renders a
 // clean document from the struct (dropping any explanatory comments the create
 // template added) and replaces the destination via a temp file in the same
 // directory followed by an atomic rename, so a failed write never leaves a
-// partial or corrupt chain file in place.
-func Write(path string, c *Chain) error {
+// partial or corrupt chain file in place. It uses 2-space indentation
+// (yamlutil.Marshal) to match the rest of the repo's chain files.
+func Write(ctx context.Context, path string, c *Chain) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 	if c == nil {
 		return errors.Validation("cannot write nil chain")
 	}
 
-	data, err := yaml.Marshal(c)
+	data, err := yamlutil.Marshal(c)
 	if err != nil {
 		return errors.Wrap(err, "marshaling chain").WithCode(errors.ErrCodeParse)
 	}
--- a/internal/chain/types.go
+++ b/internal/chain/types.go
@@ -39,9 +39,9 @@ type Metadata struct {
 	ID            string        `yaml:"id"`
 	Name          string        `yaml:"name"`
 	Goal          string        `yaml:"goal,omitempty"`
-	CreatedAt     time.Time     `yaml:"created_at"`
+	CreatedAt     time.Time     `yaml:"created_at,omitempty"`
 	Status        ChainStatus   `yaml:"status"`
-	StatusHistory []StatusEntry `yaml:"status_history"`
+	StatusHistory []StatusEntry `yaml:"status_history,omitempty"`
 }
 
 // StatusEntry is a single lifecycle state transition in the audit trail.
--- a/internal/chain/write_test.go
+++ b/internal/chain/write_test.go
@@ -1,0 +1,98 @@
+package chain
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWrite_RoundTrip(t *testing.T) {
+	dir := t.TempDir()
+	src := filepath.Join(dir, "src.yaml")
+	require.NoError(t, os.WriteFile(src, []byte(validChainYAML), 0o644))
+
+	original, err := Parse(context.Background(), src)
+	require.NoError(t, err)
+
+	out := filepath.Join(dir, "out.yaml")
+	require.NoError(t, Write(out, original))
+
+	reparsed, err := Parse(context.Background(), out)
+	require.NoError(t, err)
+
+	assert.Equal(t, original.ChainVersion, reparsed.ChainVersion)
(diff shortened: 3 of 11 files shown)


internal/chain/write.go
```diff
@@ -0,0 +1,57 @@
...
+// clean document from the struct (dropping any explanatory comments the create
+// template added) and replaces the destination via a temp file in the same
+// directory followed by an atomic rename, so a failed write never leaves a
+// partial or corrupt chain file in place.
+func Write(path string, c *Chain) error {
+	if c == nil {
+		return errors.Validation("cannot write nil chain")
+	}
+
+	data, err := yaml.Marshal(c)
```
obey-agent: **4-space indent diverges from the repo standard (2-space).** `yaml.Marshal` here emits 4-space indentation (yaml.v3 default), but every other chain write path uses `yamlutil.Marshal` (`internal/yamlutil/marshal.go`, which does `SetIndent(2)`) — e.g. `chain complete` at complete.go:88. Verified empirically in this module: `yaml.Marshal` gives `sub:\n    value` vs yamlutil's `sub:\n  value`. Result: the first `fest chain add` reindents the entire chain file 2→4 space, producing a noisy diff and inconsistency with hand-authored chains and files written by `chain complete`. Use `yamlutil.Marshal(c)` instead of `yaml.Marshal(c)`.

internal/chain/write.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/Obedience-Corp/fest/internal/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// Write marshals a Chain to YAML and writes it atomically to path. It renders a
+// clean document from the struct (dropping any explanatory comments the create
+// template added) and replaces the destination via a temp file in the same
+// directory followed by an atomic rename, so a failed write never leaves a
+// partial or corrupt chain file in place.
+func Write(path string, c *Chain) error {
```
obey-agent: **Convention break: `Write` is the only exported I/O function in this package without a `context.Context` first parameter.** `Parse`, `ParseBytes`, `DiscoverAll`, `FindForFestival`, `Resolve`, `ResolveAvailable` all take `ctx context.Context` first; the campaign Go standard mandates context on I/O. Recommend `func Write(ctx context.Context, path string, c *Chain) error` with a `ctx.Err()` check before marshal, and update the caller in add.go. Keeps the package API uniform and lets cancellation propagate if `Write` is ever called from a long-running path.

obey-agent (commented):
**Verdict: Request Changes**

**Overview**
A well-structured feature: `fest chain add` cleanly reuses the existing chain model (`resolveChainID`/`findChainByID`, `chainpkg.Validate`, `festivalsRoot`, `resolveCurrentFestivalID`) and adds a single new serialization path via `chain.Write`, with a tight validate-before-write contract and genuinely atomic temp-file+rename writes. Test coverage is thorough (round-trip, atomicity, no-temp-left-behind, duplicate rejection, invalid-mutation guard, context cancellation, wave/no-wave, by-id/name/path resolution). I verified the build, full `go test ./...`, and `golangci-lint` all pass, and confirmed there is no forked YAML write path in the command layer. One real defect blocks merge.

**Key Findings**
- **(Blocking) 4-space YAML indentation in `chain.Write`** (write.go:21, inline): `yaml.Marshal` emits yaml.v3's default 4-space indent, while the rest of the repo standardizes on 2-space via `yamlutil.Marshal` (`SetIndent(2)`) — e.g. `chain complete`. I confirmed the difference empirically in this module. The first `fest chain add` will silently reindent the whole chain file 2→4 space, producing noisy diffs and inconsistency with hand-authored chains and `chain complete` output. Fix: use `yamlutil.Marshal(c)`.
- **(Should fix) `Write` lacks a `context.Context` parameter** (write.go:16, inline): it is the only exported I/O function in `internal/chain` without `ctx` first. The campaign Go standard mandates context on I/O. Add `ctx` + a `ctx.Err()` check and update the caller.
- **(Optional, pre-existing)** `fest chain create`'s success message ("Add festivals and edges by editing the chain file directly," create.go:107) is now stale given this command exists. That file isn't in this PR, so it's not a regression here, but a one-line update in this PR would close the loop. Also a minor test-robustness nit in `add_test.go`'s `captureStdout` (restore `os.Stdout` via `t.Cleanup` so a failed `w.Close()` can't leak the override); harmless while tests are serial.

**What's Done Well**
- Validate-before-write is correctly positioned so an invalid mutation never reaches disk (`TestRunAdd_InvalidMutationWritesNothing` proves it).
- Atomic write ordering is correct (Sync → Close → Rename, with cleanup on every failure path).
- `mutateChain` appends edges before `waveUnlock` runs, so wave placement sees the full edge set.
- The `omitempty` additions to `Metadata.CreatedAt`/`StatusHistory` are narrowly scoped and backward-compatible (marshal-only zero-value suppression).
- No `fmt.Errorf` in new production code; the `errors` package is used consistently.

**Staff Standard**
Close. The design and tests are at the bar I'd submit, but the indentation divergence is exactly the kind of silent formatting regression that erodes a file format over time, and it would ship under my name as a defect. Fix the `yamlutil.Marshal` swap (and ideally the `ctx` param) and this is a clean Approve.

lancekrogers:
Addressed the review in d176280:

- **Blocking (2-space indent):** `chain.Write` now uses `yamlutil.Marshal` (SetIndent(2)) instead of `yaml.v3`'s 4-space default, so it matches `chain complete` and hand-authored chains. No more whole-file reindent on first add.
- **ctx param:** `chain.Write(ctx, path, c)` with a `ctx.Err()` check, matching the rest of the `internal/chain` I/O API; caller in `add.go` and all `write_test.go` call sites updated.
- **Stale message:** `fest chain create` now points at `fest chain add` instead of "edit the chain file directly".
- **Test robustness:** `captureStdout` restores `os.Stdout` via `t.Cleanup`.
- Also trimmed redundant doc comments on unexported helpers/structs (kept the exported `Write` doc and the trailing-wave reasoning).

Verified: `go build ./...`, fresh `go test ./internal/chain/... ./internal/commands/chain/...`, and `golangci-lint run` on the changed packages all pass; gofmt clean.

obey-agent (commented):
**Re-review (delta since 92a20463 → d1762803): Approve**

The single new commit resolves every finding from my prior review. I verified each fix against the real code, not just the diff:

- **2-space YAML indentation (blocking finding) — fixed and confirmed empirically.** `chain.Write` now uses `yamlutil.Marshal` instead of raw `yaml.Marshal`. I ran the actual `Write` path against the package's `validChainYAML` fixture and inspected the bytes: output is 2-space per level (`metadata:` → 2-space `id`, nested list items at 4/6 spaces), matching `chain complete` and hand-authored chains. No more reindentation on `fest chain add`.
- **`context.Context` on `Write` — fixed.** Signature is now `Write(ctx context.Context, path string, c *Chain) error` with a `ctx.Err()` check before marshal. The sole caller (`add.go:113`) passes `ctx`, and all `write_test.go` call sites were updated. Build is clean and I confirmed there are no other callers left unthreaded.
- **Stale `chain create` message — fixed.** It now prints `fest chain add --chain <id> --festival <id> [--after <ref>]` instead of pointing users at manual file editing.
- **`captureStdout` test robustness — fixed.** `t.Cleanup(func() { os.Stdout = old })` now guarantees stdout is restored even if a `require` fails mid-helper.

Verification run on d1762803: `go build ./...` clean, `go test ./internal/chain/... ./internal/commands/chain/...` pass, `golangci-lint` 0 issues. (The function doc-comment removals are a stylistic choice and don't affect correctness.)

**Staff Standard:** Yes. The defect is gone, the convention break is closed, and the fix is verified end-to-end. Clean approve.

PR #196, opened 2026-06-09
feat(chain): add --json to 'fest chain list' (WI-04f47a)

lancekrogers:
## Summary

Adds a `--json` flag to `fest chain list`, emitting a stable structured result so machine consumers can read the chain list without screen-scraping the human table.

```json
{ "chains": [ { "id", "name", "status", "festival_count", "refs": [...] } ] }
```

- Empty workspace returns `{ "chains": [] }` (not the human "No chains found" message).
- Human table output is unchanged.
- JSON conventions mirror `fest chain add --json` (snake_case keys, top-level object wrapper).

## Why

festival-app's create-festival modal (FA0011) needs a machine-readable chain list to populate its chain picker. Per the facade principle already applied to `fest chain add` (#195 / WI-72f9af), the app must consume a real fest JSON contract rather than parse the human table. Tracked as design workitem WI-04f47a.

## Tests

`internal/commands/chain/list_test.go`: JSON shape, empty workspace, status filter, context cancellation. `go test ./internal/commands/chain/` passes; `golangci-lint run ./...` reports 0 issues; gofmt/vet clean.

### DIFF
--- a/internal/commands/chain/list.go
+++ b/internal/commands/chain/list.go
@@ -2,6 +2,7 @@ package chain
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview**
A clean, single-purpose addition: `--json` on `fest chain list` emitting a stable `{ "chains": [...] }` contract for machine consumers (the festival-app create-festival chain picker), leaving the human table untouched. It mirrors the facade pattern and JSON conventions established by `fest chain add --json` (#195), so the app consumes a real fest contract instead of screen-scraping. I verified the build, `go test ./internal/commands/chain/...`, and `golangci-lint` all pass on the head commit.

**Key Findings**
No issues. I checked the behaviors most likely to be subtly wrong:
- **Empty/filtered output is correct and tested.** The human "No chains found" message is gated on `!jsonOut && len(unfiltered) == 0`, so JSON always reports the filtered set; an empty result serializes as `{"chains":[]}` rather than the human message. `TestRunList_JSONEmptyWorkspace` and `TestRunList_JSONStatusFilter` cover both, and `TestRunList_HumanStatusFilterEmptyPrintsNoMessage` pins the human path's filter-vs-unfiltered distinction.
- **JSON hygiene is right.** `items` and `refs` are built with `make([]T, 0, ...)` / `make([]T, len)`, so empty collections marshal as `[]`, never `null` — important for a stable wire contract.
- **Conventions match the existing facade.** snake_case keys, top-level object wrapper, and the same `chainListResult`/`chainListItem` shape family as `chain add`. `printChainListJSON` is factored out of `runList`, and marshal errors are wrapped with the project `errors.IO`.

**What's Done Well**
- Context cancellation is honored (`ctx.Err()` at entry) and explicitly tested (`TestRunList_ContextCancelled`).
- Tests reuse the existing package helpers (`addTestEnv`, `captureStdout`) rather than standing up parallel test infrastructure, and assert the full JSON shape (id/name/status/festival_count/refs).
- Scope is tight: human output path is byte-for-byte unchanged; the JSON path is additive.

**Staff Standard**
Yes, I'd submit this as my own work. It is a minimal, well-tested contract addition that stays consistent with the sibling `chain add --json` facade and handles the empty/null edge cases that matter for a machine-readable API. Approve.

PR #197, opened 2026-06-09
feat(festival): add --seed/--seed-file to 'fest create festival' (WI-88956e)

lancekrogers:
## Summary

Adds type-aware seeding of a new festival's **ingest phase** `input_specs/` at creation time:

```
--seed string        inline seed content (markdown)
--seed-file string   file whose contents seed the ingest phase (mutually exclusive with --seed)
```

On success the content is written to `<festival>/<NNN_INGEST>/input_specs/seed.md` and the path is reported in `--json` as `seed_path`.

## Type-awareness (the important part)

`input_specs/` is an **ingest-phase** concept, and festival types differ — `standard`/`research` have an ingest phase; `implementation` (`skip_ingestion`) and others may not. Seeding is **validated before any scaffolding** (fail-fast): a type with no auto ingest phase is refused with a clear error and **no partial festival is created**. fest owns this because only fest knows each type's phase layout (`FestivalType.GetAutoPhases()`).

## Why

festival-app's create-festival modal (FA0011) wants to let users seed starting context. Per the facade principle (and to avoid the app writing into fest's festival directory, which it cannot lay out correctly per-type), fest owns seeding and the app calls `fest create festival --seed`. Tracked as design workitem WI-88956e.

## Tests

`create_festival_seed_test.go`: standard happy path (inline + file), no-seed leaves no file, implementation-type rejection (fail-fast, no festival created), `--seed`+`--seed-file` mutual exclusion, and pure unit tests for `resolveSeedContent` / `ingestAutoPhaseID`. Full festival package tests pass; `golangci-lint run ./...` 0 issues; gofmt/vet clean. Existing create behavior unchanged.

### DIFF
--- a/internal/commands/festival/create_festival.go
+++ b/internal/commands/festival/create_festival.go
@@ -316,12 +316,27 @@ func RunCreateFestival(ctx context.Context, opts *CreateFestivalOptions) error {
 
 	created, res.autoPhasesCreated = autoScaffoldPhases(ctx, cfg, created)
 
-	recordInitialSize(ctx, cfg, res.festConfig)
-
 	res.created = created
 	res.copiedGates = copiedGates
+
+	// Seed before the initial-size snapshot so seeded content is part of the
+	// creation baseline, not later counted as post-create growth. Skipped on
+	// dry-run. The seed file is registered in res.created after marker processing
+	// so user content is never marker-substituted.
+	if !opts.DryRun {
+		if err := writeSeedFile(cfg, res); err != nil {
+			return emitCreateFestivalCreatedError(ctx, cfg, res, err)
+		}
+	}
+
+	recordInitialSize(ctx, cfg, res.festConfig)
+
 	if err := processAllMarkers(ctx, cfg, res); err != nil {
 		return emitCreateFestivalCreatedError(ctx, cfg, res, err)
+	}
+
+	if res.seedPath != "" {
+		res.created = append(res.created, res.seedPath)
 	}
 
 	if opts.DryRun && res.markersTotal > 0 {
@@ -330,10 +345,6 @@ func RunCreateFestival(ctx context.Context, opts *CreateFestivalOptions) error {
 			return emitCreateFestivalCreatedError(ctx, cfg, res, err)
 		}
 		return nil
-	}
-
-	if err := writeSeedFile(cfg, res); err != nil {
-		return emitCreateFestivalCreatedError(ctx, cfg, res, err)
 	}
 
 	if err := validateIfConfigured(ctx, cfg, res); err != nil {
@@ -714,7 +725,6 @@ func writeSeedFile(cfg *createConfig, res *createResult) error {
 	}
 
 	res.seedPath = seedPath
-	res.created = append(res.created, seedPath)
 	return nil
 }
 
--- a/internal/commands/festival/create_festival.go
+++ b/internal/commands/festival/create_festival.go
@@ -42,6 +42,8 @@ type CreateFestivalOptions struct {
 	VarsFile    string
 	Markers     string // Inline JSON with hint→value mappings
 	MarkersFile string // JSON file path with hint→value mappings
+	Seed        string // Inline seed/input-spec content for the ingest phase
+	SeedFile    string // File whose contents seed the ingest phase
 	SkipMarkers bool   // Skip marker processing
 	DryRun      bool   // Show markers without creating file
 	JSONOutput  bool
@@ -59,6 +61,7 @@ type createFestivalResult struct {
 	GatesDirectory    string             `json:"gates_directory,omitempty"`
 	FestYAML          string             `json:"fest_yaml,omitempty"`
 	GateTemplates     []string           `json:"gate_templates,omitempty"`
+	SeedPath          string             `json:"seed_path,omitempty"`
 	ProjectPath       string             `json:"project_path,omitempty"`
 	ProjectLinked     bool               `json:"project_linked,omitempty"`
 	Markers           []map[string]any   `json:"markers,omitempty"`
@@ -107,6 +110,8 @@ type createConfig struct {
 	tmplCtx          *tpl.Context
 	festivalType     *types.FestivalType
 	festivalTypeName string
+	seedContent      string
+	seedRequested    bool
 }
 
 // NewCreateFestivalCommand adds 'create festival'
@@ -138,6 +143,8 @@ func NewCreateFestivalCommand() *cobra.Command {
 	cmd.Flags().StringVar(&opts.VarsFile, "vars-file", "", "JSON file with variables")
 	cmd.Flags().StringVar(&opts.Markers, "markers", "", "JSON string with REPLACE marker hint→value mappings")
 	cmd.Flags().StringVar(&opts.MarkersFile, "markers-file", "", "JSON file with REPLACE marker hint→value mappings")
+	cmd.Flags().StringVar(&opts.Seed, "seed", "", "Inline seed content written to the ingest phase input_specs/ (requires a type with an ingest phase)")
+	cmd.Flags().StringVar(&opts.SeedFile, "seed-file", "", "File whose contents seed the ingest phase input_specs/ (mutually exclusive with --seed)")
 	cmd.Flags().BoolVar(&opts.SkipMarkers, "skip-markers", false, "Skip REPLACE marker processing")
 	cmd.Flags().BoolVar(&opts.DryRun, "dry-run", false, "Show template markers without creating file")
 	cmd.Flags().BoolVar(&opts.JSONOutput, "json", false, "Emit JSON output")
@@ -268,6 +275,7 @@ type createResult struct {
 	allMarkers        []map[string]any
 	validationResult  *ValidationSummary
 	registered        bool
+	seedPath          string
 }
 
 // RunCreateFestival executes the create festival command logic.
@@ -276,8 +284,17 @@ func RunCreateFestival(ctx context.Context, opts *CreateFestivalOptions) error {
 		return errors.Wrap(err, "context cancelled").WithOp("RunCreateFestival")
 	}
 
+	if opts.Seed != "" && opts.SeedFile != "" {
+		return emitCreateFestivalError(opts, errors.Validation(
+			"--seed and --seed-file are mutually exclusive"))
+	}
+
 	cfg, err := resolveCreateConfig(ctx, opts)
 	if err != nil {
+		return emitCreateFestivalError(opts, err)
+	}
+
+	if err := resolveAndValidateSeed(cfg); err != nil {
 		return emitCreateFestivalError(opts, err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/commands/festival/create_festival.go
```diff
@@ -314,6 +332,10 @@ func RunCreateFestival(ctx context.Context, opts *CreateFestivalOptions) error {
 		return nil
 	}
 
+	if err := writeSeedFile(cfg, res); err != nil {
```
obey-agent: This is a creation-time write, but it happens after `recordInitialSize` snapshots the festival directory. Because `seed.md` is a `.md` file, progress/content-delta reporting will later treat the entire seed as post-create growth instead of part of the baseline. Please write the seed before `recordInitialSize` (or recompute/save the initial size after seeding) and add a regression test that `fest.yaml`'s `initial_size_bytes` includes `input_specs/seed.md`.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This is a focused addition that keeps seeding inside `fest`, which is the right ownership boundary for type-aware festival layout. The implementation validates type compatibility before scaffolding, handles mutual exclusion, and adds good coverage for the main happy/error paths.

**Key Findings:**
- `seed.md` is written after `recordInitialSize`, so seeded markdown is excluded from `metadata.initial_size_bytes` and will later be reported as post-create content growth. I left an inline comment on `writeSeedFile`; please move the seed write before the initial-size snapshot or recompute the snapshot after seeding, and add a regression test for the saved `initial_size_bytes`.

**What's Done Well:**
- The type-aware ownership is correct: festival-app should not guess the ingest phase path.
- The fail-fast validation for types without auto ingest avoids the worst partial-create path.
- The tests cover inline seed, seed-file, no-seed, mutual exclusion, and implementation-type rejection.

**Validation Run:**
- `go test ./internal/commands/festival` passed.
- `go test ./internal/types ./internal/progress` passed.
- `just test unit` passed: 89 packages, 2219/2219 tests.

**Staff Standard:** Not quite ready to merge as-is because the progress baseline regression is user-visible in metrics. After that ordering/test fix, I would be comfortable with the shape of this PR.

obey-agent (commented):
**Verdict: Approve**

**Overview**
Adds `--seed`/`--seed-file` to `fest create festival`, writing type-aware starting context into the ingest phase's `input_specs/seed.md` and reporting `seed_path` in `--json`. The facade is correct: fest owns the seeding because only fest knows each type's auto-phase layout, so the festival-app modal (FA0011) consumes a real contract instead of guessing the directory structure. I verified build, `go test ./internal/commands/festival/...`, and `golangci-lint` all pass on the head commit.

**Key Findings**
No issues. I traced the parts most likely to be subtly wrong:
- **Fail-fast is real and proven.** Mutual exclusion of `--seed`/`--seed-file` is checked first, then `resolveAndValidateSeed` reads the seed (so a missing `--seed-file` fails early) and refuses a type with no ingest phase BEFORE `scaffoldFestivalDirectory` runs. `TestCreateFestivalSeed_ImplementationTypeRejected` asserts zero festivals are left in `planning/` on rejection, so the "no partial festival" guarantee is enforced, not just claimed.
- **Phase-ID derivation is consistent.** `ingestAutoPhaseID` computes `tpl.FormatPhaseID(i+1, name)` over `GetAutoPhases()` using the same range index as `autoScaffoldPhases` (create_festival.go:656 vs :745). I confirmed the two match, so the `slices.Contains(res.autoPhasesCreated, ingestPhaseID)` guard in `writeSeedFile` succeeds for valid types rather than spuriously failing.
- **Defensive write guard is well-judged.** Because `autoScaffoldPhases` swallows per-phase failures, `writeSeedFile` confirms the ingest phase actually landed before writing, so it can't write `seed.md` under a missing phase and falsely report success. On that rare path it returns through `emitCreateFestivalCreatedError`, which is the right call (report, don't silently succeed).

**What's Done Well**
- The project error package is used throughout with actionable context (`errors.Validation(...).WithField("type", ...).WithHint(...)`, `errors.IO(...).WithField("path", ...)`).
- Test coverage is thorough and behavioral: standard seed, seed-file, no-seed-leaves-nothing, implementation-type-rejected (fail-fast), mutual-exclusion, plus focused unit tests for `resolveSeedContent` (inline/file/missing) and `ingestAutoPhaseID` (nil/standard/impl/non-auto).
- The seed path is surfaced in both `res.created` and the `--json` `seed_path`, so the new artifact is discoverable by humans and machines.

**Staff Standard**
Yes, I'd submit this. The type-awareness, the before-scaffolding fail-fast, and the defensive "was the phase actually created" check show real care about partial-failure states, and the tests pin every branch. Approve.

internal/commands/festival/create_festival_seed_test.go
```diff
@@ -0,0 +1,220 @@
...
+		Seed:       seed,
+		JSONOutput: true,
+	})
+	seededCfg, err := config.LoadFestivalConfig(seededDir, "")
+	if err != nil {
+		t.Fatalf("load seeded config: %v", err)
+	}
+
+	growth := seededCfg.Metadata.InitialSizeBytes - baselineCfg.Metadata.InitialSizeBytes
+	if growth < int64(len(seed)) {
```
obey-agent: **This assertion is flaky: it fails ~80% of full-package `go test` runs (passes only in isolation).** I reproduced it repeatedly: `growth=8998, seed len=9000 (baseline=873, seeded=9871)`, so `growth < len(seed)` trips by ~2 bytes. The feature is correct (seeded=9871 clearly includes the 9000-byte seed); the *test* is the problem. It diffs two independently-scaffolded festivals and assumes their non-seed bytes are byte-identical, but scaffold content (timestamps/dates/etc.) jitters a couple bytes between the two creations, so the diff is non-deterministic. A flaky test that lands green locally but breaks CI intermittently is exactly what staff review should stop.

Recommend dropping the baseline comparison and asserting the single seeded festival directly:
```go
if seededCfg.Metadata.InitialSizeBytes < int64(len(seed)) {
    t.Errorf("seed not counted in initial_size_bytes: got %d, want >= %d",
        seededCfg.Metadata.InitialSizeBytes, len(seed))
}
```
This is deterministic: the ~873-byte unseeded baseline proves the scaffold alone can never reach 9000, so `InitialSizeBytes >= len(seed)` is a sufficient and stable proof that the seed is in the creation baseline. (You can drop the baseline-festival creation entirely.)

obey-agent (commented):
**Re-review (delta since 26eb8b65 → 7d66f5c5): Request Changes**

The delta's intent is good and the production change is correct: seeding now happens before the `initial_size_bytes` snapshot so seeded content is part of the creation baseline (not later miscounted as agent-produced growth), while the seed is still appended to `res.created` only after `processAllMarkers`, so user seed content is correctly excluded from REPLACE-marker substitution. I verified that exclusion holds (`processAllMarkers` ranges `res.created`, and the seed is appended after that loop), and `go build` + `golangci-lint` are clean.

One blocking issue: the new test is flaky.

**Key Findings**
- **(Blocking) `TestCreateFestivalSeed_CountedInInitialSizeBytes` is non-deterministic** (create_festival_seed_test.go:121, inline). It failed 4 of 5 full-package `go test` runs for me (passes only when run in isolation). The reproduced failure is `growth=8998, seed len=9000 (baseline=873, seeded=9871)` — the assertion `growth < len(seed)` trips by ~2 bytes. The feature works (seeded=9871 includes the 9000-byte seed); the test is wrong: it diffs two independently-created festivals and assumes their non-seed scaffold bytes are identical, but that content jitters a couple bytes between creations. Fix: assert the single seeded festival's `InitialSizeBytes >= len(seed)` and drop the baseline comparison (the ~873-byte unseeded scaffold proves 9000 is unreachable without the seed being counted). This is deterministic. A test that's green locally but breaks CI intermittently is the one thing that must not merge.

**What's Done Well**
- The reordering is well-reasoned and the rationale is captured in the comment: seed before the size snapshot, register in `res.created` after marker processing so user content is never substituted. I confirmed both properties hold.
- Dry-run correctly skips the seed write (`if !opts.DryRun`), preserving prior behavior, and the seed is still surfaced in `res.created`/`seed_path` exactly once.
- The earlier-reviewed fail-fast validation, phase-ID consistency, and defensive "was the ingest phase scaffolded" guard are all unchanged and still sound.

**Staff Standard**
The production code I'd ship. The test I would not — I would not submit a suite that fails ~80% of the time under a normal `go test ./...`. Make the size assertion deterministic (single-festival lower bound) and this is a clean Approve.

obey-agent (changes requested):
**Verdict: Request Changes**

This supersedes the earlier automated "Approve" text review. I re-reviewed current head `7d66f5c5` and reproduced the blocking issue in the current test suite.

The production change that moves seeding before `recordInitialSize` is directionally correct. The remaining blocker is the regression test at `internal/commands/festival/create_festival_seed_test.go:121`: it compares two independently-created festivals and assumes their non-seed scaffold byte counts are identical. They are not stable enough for that assertion. I reproduced the failure under normal package execution: run 4 of 5 of `go test ./internal/commands/festival -count=1` failed with `growth=8999, seed len=9000 (baseline=873, seeded=9872)`.

There is already an inline comment on the failing assertion with the specific fix: drop the baseline-festival diff and assert the seeded festival directly, e.g. `seededCfg.Metadata.InitialSizeBytes >= int64(len(seed))`. That proves the seed is included in the creation baseline without depending on byte-identical scaffold content across two separate festival creations.

**Validation Run**
- `go test ./internal/commands/festival -count=1` repeated 5 times: failed on run 4 due to `TestCreateFestivalSeed_CountedInInitialSizeBytes`.

**Staff Standard**
The production path is close, but I would not merge a known flaky test. Please make the regression assertion deterministic, then this should be ready for approval.

obey-agent (approved):
**Verdict: Approve**

**Overview:** Re-reviewed current head `d7e1c084`. The deterministic test fix addresses the previous blocker: `TestCreateFestivalSeed_CountedInInitialSizeBytes` now uses a single seeded festival lower-bound assertion instead of diffing two independently scaffolded festivals, so it no longer depends on byte-identical generated metadata.

**Key Findings:**
- No remaining blocking issues. The seed is written before `recordInitialSize`, so `seed.md` is included in `metadata.initial_size_bytes`.
- The seed path is still appended to `res.created` only after marker processing, so user-provided seed content is not marker-substituted.
- Type-aware validation, mutual exclusion, missing seed-file handling, and the actual-ingest-phase-created guard remain sound.

**Validation Run:**
- `go test ./internal/commands/festival -count=1` repeated 5 times: passed 5/5.
- `go test ./internal/commands/festival -run 'TestCreateFestivalSeed|TestIngestAutoPhaseID|TestResolveSeedContent' -count=20`: passed.
- `go test ./internal/types ./internal/progress`: passed.
- `just test unit`: passed, 89 packages, 2220/2220 tests.
- `go build ./...`: passed.
- `golangci-lint run ./...`: passed, 0 issues.

**Staff Standard:** Yes, I would submit this now. The ownership boundary is correct, the failure modes are handled, and the previous flaky regression test has been made deterministic.

PR #198, opened 2026-06-15
feat(watch): order festival picker by status priority then recency

lancekrogers:
## Summary

Orders the `fest watch` festival selection picker by lifecycle priority instead
of the current order-by-accident:

1. **active**, then **ready**, then **planning**
2. within each group, **most recently modified first**

It also excludes **ritual** festivals from `fest watch` (picker and shell
completion). A ritual is a repeatable template: `fest ritual run` copies it into
`active/` with a counter suffix, and that copy is what you watch. The template is
never the live festival, so it is not a watch target. `dungeon/*` stays excluded
(terminal).

## Root cause

The picker preserves candidate-collection order for the unfiltered list
(`internal/tui/picker/picker.go` only re-sorts on a typed query). The watch
resolver passed `PreferredStatuses: id.WorkingStatusDirectories`, which is
lifecycle order `[planning, ready, active, ritual]`, so the picker showed
planning first with ritual templates mixed in. `fest go` was already active-first
(it uses `defaultCandidateStatuses()` built from `id.PrimaryStatusDirs`), so this
was scoped to `fest watch`.

## Changes

- **`internal/commands/watch`**: new `watchPickerStatuses = [active, ready, planning]`
  (priority order, ritual + dungeon excluded), reused by both the picker resolver
  and shell completion. The resolver now requests recency ordering.
- **`internal/commands/shared`**: new opt-in `OrderByStatusThenRecency` picker
  option and a `ModTime` candidate field. Ordering is a pure, fixture-testable
  comparator (status rank -> mtime desc -> name) over a filesystem step that takes
  each festival's most-recent file mtime. Festivals whose mtime cannot be read sort
  last in their group and are never dropped.

Only the interactive watch picker opts in. `fest go` and watch completion keep
their existing alphabetical order and do no per-candidate stat work.

## Behavior changes

- Watch picker order is now active -> ready -> planning, newest-first per group.
- `fest watch` no longer surfaces ritual festivals in the picker or in
  `fest watch <TAB>` completion.

## Testing

- `just build` / `golangci-lint run ./...` (0 issues) / full `go test ./...` green.
- New pure comparator tests (grouping, status-precedence-beats-recency,
  missing-mtime-sorts-last, name tie-break, unknown-status-last) and host-temp
  `latestModTime` tests.
- New watch tests: watchable statuses exclude ritual and are active-first; ritual
  cwd does not narrow; picker passes the ordering flag; completion excludes ritual.

## Design

Design package: `workflow/design/fest-watch-picker-ordering-2026-06-15/` in the
campaign repo. Methodology note:
`docs/tribal-knowledge/ritual-templates-excluded-from-watch.md`.

### DIFF
--- a/internal/commands/watch/resolver.go
+++ b/internal/commands/watch/resolver.go
@@ -210,14 +210,10 @@ func isFestivalNotFound(err error) bool {
 	return stderrors.As(err, &structured) && structured.Code == errors.ErrCodeNotFound
 }
 
-// watchPickerStatuses are the statuses surfaced by `fest watch`, in display
-// priority order. Ritual templates are excluded: `fest ritual run` copies a
-// ritual into active/ before there is anything to watch, so the template is
-// never itself a watch target. dungeon/* is terminal and likewise never watched.
+// watchPickerStatuses are watch targets, in display priority order. Ritual is a
+// template (run into active/ first) and dungeon is terminal, so neither is watched.
 var watchPickerStatuses = []string{"active", "ready", "planning"}
 
-// pickerStatuses narrows the watch picker to the watchable status dir the user is
-// inside, else all watchable statuses (active, ready, planning).
 func pickerStatuses(cwd, festivalsDir string) []string {
 	if narrowed := preferredPickerStatuses(cwd, festivalsDir); len(narrowed) > 0 {
 		return narrowed
--- a/internal/commands/watch/resolver.go
+++ b/internal/commands/watch/resolver.go
@@ -100,6 +100,7 @@ func (r targetResolver) resolve(ctx context.Context, selector string) (*show.Fes
 	path, err := r.pickFestival(ctx, festivalsDir, shared.FestivalPickerOptions{
 		IncludeStatusDirectories: false,
 		PreferredStatuses:        pickerStatuses(cwd, festivalsDir),
+		FallbackStatuses:         watchPickerStatuses,
 		OrderByStatusThenRecency: true,
 	})
 	if err != nil {
--- a/internal/commands/watch/resolver.go
+++ b/internal/commands/watch/resolver.go
@@ -100,6 +100,7 @@ func (r targetResolver) resolve(ctx context.Context, selector string) (*show.Fes
 	path, err := r.pickFestival(ctx, festivalsDir, shared.FestivalPickerOptions{
 		IncludeStatusDirectories: false,
 		PreferredStatuses:        pickerStatuses(cwd, festivalsDir),
+		OrderByStatusThenRecency: true,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


internal/commands/watch/resolver.go
```diff
@@ -100,6 +100,7 @@ func (r targetResolver) resolve(ctx context.Context, selector string) (*show.Fes
 	path, err := r.pickFestival(ctx, festivalsDir, shared.FestivalPickerOptions{
 		IncludeStatusDirectories: false,
 		PreferredStatuses:        pickerStatuses(cwd, festivalsDir),
```
obey-agent, line 102: Blocking: this still has a path where ritual/dungeon candidates can leak back into the watch picker. When `pickerStatuses` narrows to a single cwd status such as `[]string{"active"}` and that status has no candidates, `FestivalPickerItems` falls back by clearing `PreferredStatuses`; the second `CollectFestivalPickCandidates` call then uses `defaultCandidateStatuses()`, which includes `ritual` and `dungeon/*`. That violates the PR/design guarantee that watch only surfaces active/ready/planning. Please keep the fallback bounded to the watchable set, for example by adding a separate allowed-status cap or by having the watch picker fall back from the narrowed status to `watchPickerStatuses` instead of to the shared default set, with a regression where active is empty but ritual/dungeon contains festivals.

obey-agent (commented):
**Verdict: Request Changes**

The PR is well-scoped and the core comparator is clean: active/ready/planning priority, newest-first within groups, and completion staying alphabetical are all implemented in the right area. I would not merge yet because the watch picker can still fall back into the shared default status set and surface ritual/dungeon festivals in an empty narrowed-status edge case.

**Key Findings**

- `internal/commands/watch/resolver.go`: the picker passes a narrowed `PreferredStatuses` list, but `FestivalPickerItems` clears that preference when no candidates are found. That fallback re-enters `defaultCandidateStatuses()`, which includes `ritual` and `dungeon/*`, so the new watch-only status guarantee is not enforced in that path. I left an inline comment with the exact scenario and suggested fixes.

**What is Done Well**

- The ordering comparator is deterministic and separately tested for status priority, recency, zero mtimes, and name tie-breaks.
- The change keeps `fest go` and shell completion away from the filesystem mtime walk.
- The design correctly documents explicit selector resolution as out of scope, so I did not treat that as a blocker.

**Validation**

- `git diff --check origin/main...HEAD`: passed.
- `go test ./internal/commands/watch ./internal/commands/shared`: passed.
- `just build quick`: passed.
- `just lint all`: passed, 0 issues.
- `just test all-raw`: passed, including container integration tests.

**Staff Standard**

Not yet. I would be happy with the direction, but the picker needs the fallback status cap before this matches its own public behavior contract.

obey-agent (commented):
**Verdict: Approve**

Re-reviewed PR #198 at `9b76b7d17aba19073761c1f3afa4fbffb9d07285`.

The prior watch picker fallback issue is fixed: `watch` now passes an explicit fallback status set, so an empty narrowed result broadens only within watchable statuses instead of leaking `ritual` or `dungeon/*` festivals back into the picker. The new regression tests cover both the no-leak case and broadening from `active` into `ready` while still excluding non-watchable statuses.

Local verification passed:
- `go test ./internal/commands/watch ./internal/commands/shared`
- `just build quick`
- `just lint all`
- `just test all-raw`

No further changes requested.

obey-agent (approved):
Re-review passed at `9b76b7d17aba19073761c1f3afa4fbffb9d07285`. The previous fallback-status issue is fixed and local verification passed. No further changes requested.

PR #205, opened 2026-06-16
Intent-backlog cleanup — fest (6 intents) [WI-ee98f2]

lancekrogers:
# Intent-backlog cleanup — fest

Consolidated umbrella PR for the fest portion of the intent-backlog cleanup (`[WI-ee98f2]`). Each intent landed as its own gate-green PR into the `intent-cleanup` branch; this is the single review surface for what is headed to `main`. Nothing is on `main` yet.

Part of `workflow/design/intent-backlog-cleanup/` (`WI-ee98f2`).

## What's in here (6 intents)

| Per-intent PR | Intent | Summary |
|---|---|---|
| #200 | Workflow phase header | Render the real phase type instead of a generic "WORKFLOW" label in the workflow phase step header |
| #201 | Link-hijack guard | Block a planning-festival link from silently overwriting an active festival's project link |
| #202 | Ingest template | Add a GATHER step as the first ingest workflow step (copy existing requirements into `input_specs/`) |
| #203 | `fest walk` / `fest inspect` | New guided festival-orientation command |
| #204 | Picker cluster (2 intents) | Picker fallback for `fest status` / `fest progress` with no context **+** left/right festival cycling in `fest watch` |

## Intent detail

- **#200** — `navigator.FormatProgress` now uses `displayPhaseType()` rather than the raw `string(n.mode)`, so the header shows the actual phase type and cannot silently regress to "WORKFLOW".
- **#201** — `linkFestivalToProject` already guarded against hijacking an active festival's link; the project→festival direction did not (see "Review findings" below).
- **#203** — `fest walk` (alias `inspect`) prints what a festival is, where it is, status/progress, next task, blocked tasks, active gates, and warnings. It is read-mostly: it can transparently migrate a legacy progress file on first access (`progress.NewManager`), which the help text now states honestly.
- **#204** — When run outside any festival, `status`/`progress` fall back to a picker; `watch` gains left/right cycling across a festival cluster.

## Review findings addressed in this umbrella

Two commits in this umbrella fix issues that the per-intent reviews missed:

### `1f634bb` — complete the #201 hijack guard + two correctness fixes
- **Blocking:** `runGoLink` routes the in-a-project case to `linkProjectToFestival`, whose picker can surface planning festivals, but that function called `SetLinkWithPath` with **no** conflict guard. A project already linked to an active festival could be silently re-pointed at a planning festival. Added the same `ProjectConflict` + `isActiveFestivalPath` guard used by `linkFestivalToProject`, with `--force` escape semantics threaded through `runGoLink` → `linkProjectToFestival` and the `link.go` caller.
- **walk UTF-8:** goal-line truncation used a byte slice (`trimmed[:300]`) which can split a multibyte rune; now truncates on a rune slice.
- **`isActiveFestivalPath`:** matched *any* path component named `active` (so a workspace under an ancestor dir literally named `active` was misclassified); now checks the immediate parent dir only (`filepath.Base(filepath.Dir(path)) == "active"`), matching `walk.detectKind`.
- Tests: `TestLinkProjectToFestival_HijackGuardBlocks`, `TestIsActiveFestivalPath_ImmediateParentOnly`, `TestLoadGoal_MultibyteTruncation`.

### `eb7934b` — restore cwd location granularity regressed by #204 + polish
- **Blocking regression:** #204 changed `detectPath`/`targetPath` from `cwd` to the resolved festival **root** in `status/handlers.go` and `progress/progress.go`. Since `ResolveFestivalPath` returns the festival root, both commands always reported `loc.Type = "festival"` regardless of the real cwd, losing phase/sequence granularity. Restored `cwd` for the non-picker branch (the picker branch still uses the picked root, since cwd is outside any festival). Tests: `TestStatusLocationGranularity_SubdirUsescwd`, `TestProgressLocationGranularity_SubdirUsescwd`.
- **`watch` infinite-spin guard:** `runWatchCycle` now tracks a stale-since index and returns a validation error after one full rotation with no successful festival load, instead of spinning CPU when every cycle target goes stale.
- **Honesty/cleanup:** dropped the false "read-only" claim from `fest walk` help; removed an unreachable `ritual-run` branch in `detectKind` (ritual-status festivals live under `festivals/ritual/` and are caught by the `ritual-template` check; nothing consumed `ritual-run`).

## Verification (at head `eb7934b`)

- `go build ./...` — clean
- `just lint all` — **0 issues**; `just lint vet` — clean
- Changed-package tests green: `navigation`, `walk`, `status`, `progress`, `watch`, `guidance/workflow`, `navigation` (state) — including all new hijack-guard / granularity / UTF-8 / cycling tests.

## Non-blocking follow-ups (not gating)

- `fest walk` is sold as orientation but can migrate legacy progress formats on first access. The help text now says so; a true read-only mode in `progress.NewManager` would be cleaner long-term.
- The hijack-guard test exercises `ProjectConflict` + `isActiveFestivalPath` directly rather than the picker-driven command path (the picker needs a TTY). Adequate, but an extracted/seam-tested guard would be stronger.

### DIFF
--- a/internal/commands/navigation/go_link.go
+++ b/internal/commands/navigation/go_link.go
@@ -165,6 +165,15 @@ func linkFestivalToProject(ctx context.Context, cwd, targetPath string) error {
 	// Get the festival path for reverse navigation
 	festivalPath := loc.Festival.Path
 
+	if existing, existingFestivalPath, hasConflict := nav.ProjectConflict(festivalName, projectPath); hasConflict {
+		if isActiveFestivalPath(existingFestivalPath) {
+			return festErrors.Validation("project is already linked to an active festival").
+				WithField("existing_festival", existing).
+				WithField("project", projectPath).
+				WithField("hint", "use 'fest link --force' to override, or run 'fest unlink' from the active festival first")
+		}
+	}
+
 	// Set the bidirectional link with festival path
 	nav.SetLinkWithPath(festivalName, projectPath, festivalPath)
 
--- a/internal/commands/navigation/go_link.go
+++ b/internal/commands/navigation/go_link.go
@@ -29,6 +29,8 @@ var (
 
 // NewGoLinkCommand creates the context-aware link subcommand for fest go
 func NewGoLinkCommand() *cobra.Command {
+	var force bool
+
 	cmd := &cobra.Command{
 		Use:   "link [path]",
 		Short: "Link current festival to a project directory (or vice versa)",
@@ -59,14 +61,16 @@ Examples:
 			if len(args) > 0 {
 				path = args[0]
 			}
-			return runGoLink(cmd.Context(), path)
+			return runGoLink(cmd.Context(), path, force)
 		},
 	}
+
+	cmd.Flags().BoolVar(&force, "force", false, "overwrite an existing active festival link")
 
 	return cmd
 }
 
-func runGoLink(ctx context.Context, targetPath string) error {
+func runGoLink(ctx context.Context, targetPath string, force bool) error {
 	cwd, err := os.Getwd()
 	if err != nil {
 		return festErrors.IO("getting current directory", err)
@@ -78,7 +82,7 @@ func runGoLink(ctx context.Context, targetPath string) error {
 	}
 
 	// We're in a project directory, show festival picker
-	return linkProjectToFestival(cwd)
+	return linkProjectToFestival(cwd, force)
 }
 
 // isInsideFestival checks if the path is within a festivals/ directory
@@ -192,7 +196,7 @@ func linkFestivalToProject(ctx context.Context, cwd, targetPath string) error {
 }
 
 // linkProjectToFestival shows a picker to select a festival to link
-func linkProjectToFestival(cwd string) error {
+func linkProjectToFestival(cwd string, force bool) error {
 	absPath, err := filepath.Abs(cwd)
 	if err != nil {
 		return festErrors.Wrap(err, "resolving current directory")
@@ -265,6 +269,17 @@ func linkProjectToFestival(cwd string) error {
 	nav, err := navigation.LoadNavigation()
 	if err != nil {
 		return festErrors.Wrap(err, "loading navigation state")
+	}
+
+	if !force {
+		if existing, existingFestivalPath, hasConflict := nav.ProjectConflict(selectedFestival, absPath); hasConflict {
+			if isActiveFestivalPath(existingFestivalPath) {
+				return festErrors.Validation("project is already linked to an active festival").
+					WithField("existing_festival", existing).
+					WithField("project", absPath).
+					WithField("hint", "use 'fgo link --force' to override, or run 'fest unlink' from the active festival first")
+			}
+		}
 	}
 
 	// Set the bidirectional link with festival path
--- a/internal/commands/navigation/link.go
+++ b/internal/commands/navigation/link.go
@@ -21,6 +21,7 @@ import (
 type linkOptions struct {
 	showLink bool
 	json     bool
+	force    bool
 }
 
 // NewLinkCommand creates the link command
@@ -65,6 +66,7 @@ Use 'fest unlink' to remove the link for current festival.`,
 
 	cmd.Flags().BoolVar(&opts.showLink, "show", false, "show current link")
 	cmd.Flags().BoolVar(&opts.json, "json", false, "output in JSON format")
+	cmd.Flags().BoolVar(&opts.force, "force", false, "overwrite an existing active festival link")
 
 	return cmd
 }
@@ -144,6 +146,18 @@ func runLink(ctx context.Context, targetPath string, opts *linkOptions) error {
 	// Create link with festival path for reverse navigation
 	festivalName := loc.Festival.Name
 	festivalPath := loc.Festival.Path
+
+	if !opts.force {
+		if existing, existingFestivalPath, hasConflict := nav.ProjectConflict(festivalName, absPath); hasConflict {
+			if isActiveFestivalPath(existingFestivalPath) {
+				return errors.Validation("project is already linked to an active festival").
+					WithField("existing_festival", existing).
+					WithField("project", absPath).
+					WithField("hint", "use --force to override, or run 'fest unlink' from the active festival first")
+			}
+		}
+	}
+
 	nav.SetLinkWithPath(festivalName, absPath, festivalPath)
 
 	// Save navigation state
@@ -437,4 +451,20 @@ func findFestivalStatus(festivalsDir, festivalName string) string {
 		}
 	}
 	return ""
+}
+
+// isActiveFestivalPath reports whether the stored festival path is inside the active/ directory.
+// A festival path of "" means the link predates festival-path tracking; treat it as active
+// to be safe (prefer the existing link when uncertain).
+func isActiveFestivalPath(festivalPath string) bool {
+	if festivalPath == "" {
+		return true
+	}
+	parts := strings.Split(filepath.ToSlash(festivalPath), "/")
+	for _, part := range parts {
+		if part == "active" {
+			return true
+		}
+	}
+	return false
 }
--- a/internal/commands/navigation/link.go
+++ b/internal/commands/navigation/link.go
@@ -92,7 +92,7 @@ func runLink(ctx context.Context, targetPath string, opts *linkOptions) error {
 			}
 			projectDir = absPath
 		}
-		return linkProjectToFestival(projectDir)
+		return linkProjectToFestival(projectDir, opts.force)
 	}
 
 	// Inside a festival - detect location for festival metadata
@@ -460,11 +460,5 @@ func isActiveFestivalPath(festivalPath string) bool {
 	if festivalPath == "" {
 		return true
 	}
-	parts := strings.Split(filepath.ToSlash(festivalPath), "/")
-	for _, part := range parts {
-		if part == "active" {
-			return true
-		}
-	}
-	return false
+	return filepath.Base(filepath.Dir(festivalPath)) == "active"
 }
--- a/internal/commands/progress/progress.go
+++ b/internal/commands/progress/progress.go
@@ -137,7 +137,10 @@ func runProgress(ctx context.Context, opts *progressOptions) error {
 		resolvedFestivalPath = picked
 	}
 
-	targetPath := resolvedFestivalPath
+	targetPath := cwd
+	if resolveErr != nil {
+		targetPath = resolvedFestivalPath
+	}
 	if opts.taskPath != "" {
 		resolvedTaskPath, err := resolveTaskPath(opts.taskPath, resolvedFestivalPath, cwd)
 		if err != nil {
--- a/internal/commands/progress/progress.go
+++ b/internal/commands/progress/progress.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Obedience-Corp/fest/internal/lifecycle"
 	"github.com/Obedience-Corp/fest/internal/progress"
 	"github.com/Obedience-Corp/fest/internal/scope"
+	"github.com/Obedience-Corp/fest/internal/workspace"
 	"github.com/spf13/cobra"
 )
 
@@ -121,12 +122,22 @@ func runProgress(ctx context.Context, opts *progressOptions) error {
 	}
 
 	// Use shared helper to resolve festival path (supports linked festivals)
-	resolvedFestivalPath, err := shared.ResolveFestivalPath(cwd, festivalPath)
-	if err != nil {
-		return errors.Wrap(err, "detecting festival location")
+	resolvedFestivalPath, resolveErr := shared.ResolveFestivalPath(cwd, festivalPath)
+	if resolveErr != nil {
+		if opts.festival != "" || opts.taskID != "" || opts.taskPath != "" {
+			return errors.Wrap(resolveErr, "detecting festival location")
+		}
+		picked, pickErr := pickFestivalForProgress(ctx, cwd)
+		if pickErr != nil {
+			return pickErr
+		}
+		if picked == "" {
+			return nil
+		}
+		resolvedFestivalPath = picked
 	}
 
-	targetPath := cwd // Use current directory for location detection
+	targetPath := resolvedFestivalPath
 	if opts.taskPath != "" {
 		resolvedTaskPath, err := resolveTaskPath(opts.taskPath, resolvedFestivalPath, cwd)
 		if err != nil {
@@ -172,4 +183,20 @@ func runProgress(ctx context.Context, opts *progressOptions) error {
 
 	// Show progress overview
 	return showProgressOverview(ctx, mgr, loc, opts)
+}
+
+var progressPickerStatuses = []string{"active", "ready", "planning"}
+
+func pickFestivalForProgress(ctx context.Context, cwd string) (string, error) {
+	festivalsDir, err := workspace.FindFestivals(cwd)
+	if err != nil || festivalsDir == "" {
+		return "", errors.NotFound("festival").
+			WithHint("navigate to a festival directory or a campaign workspace")
+	}
+	return shared.PickFestivalPath(ctx, festivalsDir, shared.FestivalPickerOptions{
+		IncludeStatusDirectories: false,
+		PreferredStatuses:        progressPickerStatuses,
+		FallbackStatuses:         progressPickerStatuses,
+		OrderByStatusThenRecency: true,
+	})
 }
--- a/internal/commands/walk/walk.go
+++ b/internal/commands/walk/walk.go
@@ -1,0 +1,464 @@
+package walk
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/Obedience-Corp/fest/internal/commands/shared"
+	"github.com/Obedience-Corp/fest/internal/commands/show"
+	"github.com/Obedience-Corp/fest/internal/config"
+	"github.com/Obedience-Corp/fest/internal/errors"
+	gatescore "github.com/Obedience-Corp/fest/internal/gates"
+	"github.com/Obedience-Corp/fest/internal/progress"
+	"github.com/Obedience-Corp/fest/internal/scope"
+	tpl "github.com/Obedience-Corp/fest/internal/template"
+	"github.com/Obedience-Corp/fest/internal/ui"
+	"github.com/Obedience-Corp/fest/internal/workspace"
+	"github.com/spf13/cobra"
+)
+
+type walkOptions struct {
+	json bool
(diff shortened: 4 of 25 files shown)


internal/commands/walk/walk.go
```diff
@@ -0,0 +1,464 @@
...
+				}
+				if inFrontmatter {
+					continue
+				}
+			}
+			if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+				continue
+			}
+			if len(trimmed) > 300 {
+				trimmed = trimmed[:300] + "..."
```
obey-agent: Minor (UTF-8): byte-slice truncation can split a multibyte rune. A goal line >300 bytes with a rune straddling byte 300 yields a dangling lead byte (replacement char) in the output. Use rune-aware truncation — convert to `[]rune` and slice, or back up to a rune boundary with `utf8.RuneStart`. Same class as the ob `CapDiff` truncation note.

internal/commands/navigation/link.go
```diff
@@ -438,3 +452,19 @@ func findFestivalStatus(festivalsDir, festivalName string) string {
 	}
 	return ""
 }
+
+// isActiveFestivalPath reports whether the stored festival path is inside the active/ directory.
+// A festival path of "" means the link predates festival-path tracking; treat it as active
+// to be safe (prefer the existing link when uncertain).
+func isActiveFestivalPath(festivalPath string) bool {
```
obey-agent, line 459: Minor: `isActiveFestivalPath` matches *any* path component equal to `"active"`, whereas the convention elsewhere is the immediate parent dir (`festivals/active/{name}`; cf. `detectKind`'s parent-dir check in walk.go). A workspace living under an ancestor dir literally named `active` would misclassify planning festivals as active. The failure mode is conservative (over-blocks and suggests `--force`, never hijacks), so this is minor — prefer `filepath.Base(filepath.Dir(festivalPath)) == "active"`.

internal/commands/navigation/go_link.go
```diff
@@ -165,6 +165,15 @@ func linkFestivalToProject(ctx context.Context, cwd, targetPath string) error {
 	// Get the festival path for reverse navigation
 	festivalPath := loc.Festival.Path
 
+	if existing, existingFestivalPath, hasConflict := nav.ProjectConflict(festivalName, projectPath); hasConflict {
+		if isActiveFestivalPath(existingFestivalPath) {
```
obey-agent, line 173: **BLOCKER: the #201 hijack guard is missing on this path — the most likely hijack vector.** `runGoLink` routes the in-a-project case to `linkProjectToFestival`, which renders both active and planning festivals as picker options (lines 207, 225). This `SetLinkWithPath` call has no `ProjectConflict`/`isActiveFestivalPath` guard — the guard was only added to `linkFestivalToProject` (go_link.go:168-169) and `link.go:151-152`. `SetLinkWithPath` then silently reassigns: `state.go:177-179` does `delete(n.Links, oldFestival)` for the project's existing link before pointing it at the selected festival.

So a project already linked to an **active** festival can be silently re-pointed at a **planning** festival by running `fest link` from inside the project directory and choosing it from the picker — exactly the hijack #201 is meant to prevent. Add the same guard (and a `--force` escape) before this call.

Verified by tracing all three production `SetLinkWithPath` call sites, the `runGoLink` routing, and `SetLinkWithPath`'s delete-then-set semantics on the PR head.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** Umbrella PR for the fest portion of the intent cleanup (6 intents). I built the PR head, ran the changed-package tests (pass), `go vet` (clean), and exercised the binaries. Five of six intents are clean or near-clean; #201 has a real guard gap that blocks merge.

**Blocking (inline on `go_link.go:169`):** #201's hijack guard is wired into `linkFestivalToProject` (go_link.go:168) and `link.go:151`, but **not** into `linkProjectToFestival`, which is where `runGoLink` routes when you invoke `fest link` from inside a project directory. That picker renders planning festivals as selectable options (go_link.go:207, 225), and the unguarded `SetLinkWithPath` at :271 silently deletes the project's existing link (`state.go:177-179`) before reassigning it. So a project linked to an active festival can be silently re-pointed at a planning festival via the picker — the exact hijack #201 targets, and the likeliest vector. Add the same guard plus a `--force` escape before that call. (Verified by tracing all three call sites, the routing, and `SetLinkWithPath`'s delete-then-set semantics.)

**Minor (inline):**
- `walk.go:231`: byte-slice goal truncation can split a multibyte rune.
- `link.go:459`: `isActiveFestivalPath` matches `"active"` anywhere in the path rather than the immediate parent dir; conservative failure mode, but prefer the parent-dir check.

**Minor (not inline — target lines are unchanged):**
- #200 is incomplete: `displayPhaseType()` was applied to `formatStep` and `formatComplete` but not to `FormatProgress` (navigator.go:593, still `string(n.mode)`), so `fest progress` on a workflow phase still shows the generic "WORKFLOW" label this intent set out to fix. Fold the helper in there too.
- #204: `runProgress` lacks the `opts.json` short-circuit that `status` has, so `progress --json` with no context in a TTY would launch the interactive picker instead of emitting a JSON error. Piped/non-TTY `--json` is safe (PickFestivalPath short-circuits), so this is the TTY corner only.
- `internal/commands/walk/walk.go` is not gofmt-clean (struct-tag alignment). It does not fail the gate (golangci-lint defaults exclude gofmt and the justfile has no fmt-check), but `just fmt` will produce a diff — run `gofmt -w` before merge.

**What's done well:** #203 `fest walk` is genuinely read-only at its own layer (zero mutation calls; verified no filesystem changes across repeated runs on a current-format festival; the only write path is the inherited legacy-format migration in the shared `progress.Store.Load`, common to all read commands). #200/#202/#204 are otherwise correct, context is propagated, new production code uses the project error package (no `fmt.Errorf`), and no file over 500 lines is introduced.

**Staff standard:** Strong, well-decomposed cleanup, but I would not ship #201 while its primary hijack path is still open. Close the guard gap and re-request.

internal/commands/walk/walk.go
```diff
@@ -0,0 +1,464 @@
...
+	"github.com/spf13/cobra"
+)
+
+type walkOptions struct {
+	json bool
+	from string
+}
+
+// WalkView is the structured view returned by fest walk.
+type WalkView struct {
```
obey-agent, line 30: New internal/commands/walk (and walk_test) is a substantial addition (~780 lines) providing read-only festival orientation ( / ). Uses context, internal/errors, reuses show/progress/gatescope — good extension not parallel impl. The picker fallback tests and link guard (go_link/link.go) address real UX holes for no-context and hijack cases. All per-intent pieces were gate-green before umbrella. (Maintainability / Go standards)

obey-agent (commented):
**Verdict: Approve**

**Overview:** Consolidated umbrella for 6 intents in fest (workflow phase header, link-hijack guard, ingest GATHER template, new `fest walk`/`inspect`, picker fallbacks for status/progress, festival cycling in watch). All per-intent PRs were individually gate-green; this is the review surface before main. Heavy on new functionality (walk ~780 LOC + tests) + targeted UX fixes and hardening in navigation/status/watch. `build`+`vet` clean on umbrella.

**Key Findings:**
- Inline posted on internal/commands/walk/walk.go: good use of context, internal/errors package, reuse of existing show/progress/gates/scope rather than duplication. New read-only orientation command is a natural extension.
- Picker fallbacks (status/handlers, progress) and link guard (navigation/link.go + go_link) close real holes for no-active-festival and context hijack cases; tests added.
- The workflow template update (ingest/WORKFLOW.md) and navigator/parser_test updates are small and consistent.
- No issues with the watch cycling or root command wiring visible in targeted review.
- PR body correctly notes that full `just test` on umbrella is required before main merge (per-intent were green individually).

**What's Done Well:**
- Consistent pattern across the fest/obey/camp/festival-app waves: per-intent small PRs + umbrella for consolidated review.
- New walk command reuses rather than reinvents (DRY).
- Guard against link hijack and picker fallbacks improve the "first 5 minutes" and daily driver experience.
- Tests for the new functionality (picker, walk, cycling).

**Staff Standard:** Yes. This continues the high-quality incremental delivery of the intent backlog items. The new walk/inspect is a nice power-user and onboarding surface. Would be proud to submit after the full test run on the umbrella as noted.

Minor: the walk command help/aliases look solid from the snippet; ensure `fest walk --help` and JSON output are exercised in the full test.

lancekrogers (commented):
Independent review complete. I did not accept prior approval notes as authoritative. I reviewed the implementation and ran targeted tests on this branch; no blocking defects found.\n\nNo merge-blocking issue identified; if you want one follow-up hardening, consider a regression test for fragmented terminal key reads in watch-cycle input handling.

obey-agent (changes requested):
Final independent review (obey-agent).

Verdict: **REQUEST CHANGES**

I confirmed a merge-blocking issue in intent #201’s hijack protection:

- The guard is applied in `go_link.go` for the `linkFestivalToProject` path, but not for `linkProjectToFestival`.
- `fest link` executed from inside a project directory routes through `linkProjectToFestival` (and the picker used in that flow can now surface planning festivals).
- Inside that function, `SetLinkWithPath` is still called on selected entries without the planning-festival safety guard, which can silently overwrite an active project link.
- This can re-point an existing linked project to the wrong festival context and is not a safe failure mode.

Because intent #201 is the explicit blocker, this PR should not land until `linkProjectToFestival` has the same guard/`--force` semantics as `linkFestivalToProject`.

lancekrogers:
**Blocking findings addressed** in `eb7934b`, verified live:

**#201 hijack guard** — `linkProjectToFestival` (the in-a-project path the original fix missed) now takes a `force` param and applies the same `ProjectConflict`/`isActiveFestivalPath` guard before `SetLinkWithPath`, with a `--force` escape. All three `SetLinkWithPath` call sites are guarded. `TestLinkProjectToFestival_HijackGuardBlocks` passes.

**#204 status/progress granularity** — restored `cwd`-based location detection in the non-picker branch (the picker branch still uses the picked festival root). Reproduced live from `…/DC0001/001_INGEST/`: `fest status` → `Location phase · Phase 001_INGEST`, `fest progress` → phase-scoped step breakdown (was festival-level). New subdir granularity tests for both.

Polish also landed: walk.go rune-safe truncation, `isActiveFestivalPath` immediate-parent match, removed dead `ritual-run` branch, `FormatProgress` uses `displayPhaseType()`, cycling.go full-cycle stale guard, and the walk help text no longer claims read-only (it can run a one-time legacy migration).

just lint all: 0 issues · just test unit-raw: 91 packages pass.

obey-agent (approved):
**Verdict: Approve** (re-review after fixes; obey-agent)

**Overview:** fest portion of the intent-backlog cleanup (6 intents). Every blocking finding from the prior round is resolved, and I reproduced the key ones firsthand rather than trusting the tests: the #201 hijack guard now covers the in-a-project link path, and the #204 location-granularity regression is fixed. The remaining review items are all addressed.

**Verified firsthand (head eb7934b):**
- **#201 hijack guard** — `linkProjectToFestival` now takes `force` and applies the same `ProjectConflict`/`isActiveFestivalPath` guard before `SetLinkWithPath`, with a `--force` escape. All three `SetLinkWithPath` call sites are guarded. `TestLinkProjectToFestival_HijackGuardBlocks` passes.
- **#204 status/progress granularity** — `handlers.go` passes `cwd` to `DetectCurrentLocation` in the non-picker branch (picked path otherwise); `progress.go` defaults `targetPath = cwd`, uses the picked root only when the picker ran, and preserves the task-path branch. Reproduced live from `…/DC0001/001_INGEST/`: `fest status` -> `Location phase / Phase 001_INGEST`, `fest progress` -> phase-scoped step breakdown (was festival-level). New subdir granularity tests for both packages.
- **cycling.go** — the `staleSince` full-cycle guard bounds the all-stale spin to one rotation then returns a Validation error; the watch goroutine distinguishes a real watch error from cancellation.
- **loadGoal** — rune-based truncation (`[]rune`), no mid-rune split; test covers a multibyte line at the cut.
- **isActiveFestivalPath** — immediate-parent match (`filepath.Base(filepath.Dir(...)) == "active"`); test covers the ancestor-dir-named-active regression case.
- Dead `ritual-run` branch removed; `FormatProgress` uses `displayPhaseType()`; walk help text now discloses the one-time legacy migration (no false read-only claim).
- `errors` package throughout (no `fmt.Errorf`; the lint ratchet is clean), `ctx.Err()` checks, JSON output paths untouched.

**What's done well:** Reuses the show/progress/gatescope helpers instead of parallel implementations; the picker fallback and watch cycling handle the real edge cases (zero/one festival, non-TTY, MakeRaw failure, wrap-around); tests now include the subdir-granularity cases that previously let the #204 regression slip through.

**Non-blocking (not gating):** `handlers.go` (693) and `navigator.go` (749) exceed the 500-line guideline, but both pre-existed on main (this PR added only +28/-6 and +14/-5) — pre-existing tech debt, out of scope. `readCycleKey` leaks its stdin-read goroutine on ctx cancel — acceptable for a short-lived CLI.

**Gate:** `just lint all` 0 issues; `just test unit-raw` 91 packages pass.

**Staff Standard:** Yes. The blocking regressions are genuinely fixed (verified by reproducing them, not by trusting passing tests), the fixes are minimal and consistent with existing patterns, and the edge cases are covered. I would submit this as my own work.

internal/commands/navigation/go_link.go
```diff
@@ -258,6 +271,17 @@ func linkProjectToFestival(cwd string) error {
 		return festErrors.Wrap(err, "loading navigation state")
 	}
 
+	if !force {
```
obey-agent, line 274: This is where the #201 blocker is actually closed. The festival→project direction (`linkFestivalToProject`) already had this guard, but `runGoLink`'s in-a-project case routes here to `linkProjectToFestival`, whose picker can surface planning festivals — and the pre-fix code called `SetLinkWithPath` with no `ProjectConflict` check, so a project already bound to an *active* festival could be silently re-pointed at a planning one. The guard correctly only blocks when the existing link is active (`isActiveFestivalPath(existingFestivalPath)`), leaves planning→anything free, and offers `--force`. Verified `ProjectConflict` returns a conflict only for a *different* festival and conservatively treats a `nil`-link inconsistency as active (empty path → blocked). Correct and complete.

internal/commands/progress/progress.go
```diff
@@ -121,12 +122,25 @@ func runProgress(ctx context.Context, opts *progressOptions) error {
...
+			return pickErr
+		}
+		if picked == "" {
+			return nil
+		}
+		resolvedFestivalPath = picked
 	}
 
-	targetPath := cwd // Use current directory for location detection
+	targetPath := cwd
```
obey-agent, line 140: Good catch landing here (`eb7934b`). #204 had changed this to the resolved festival **root**, and since `ResolveFestivalPath` returns the root, `status`/`progress` always reported `loc.Type = "festival"` even from a phase/sequence subdir — a silent granularity regression that slipped through the per-intent gates. Restoring `cwd` for the non-picker branch (picker branch keeps the picked root, since cwd is outside any festival) is the right shape, and the new `*_LocationGranularity_SubdirUsescwd` tests pin it. One edge worth a future glance: when `--path`/`opts.path` resolves a festival from outside it, detection still keys off `cwd`; that matches pre-#204 behavior, so not a regression here.

obey-agent (commented):
Staff review (obey-agent) — re-reviewed at head `eb7934b`. I read the full diff, traced the link/status/progress/watch paths, verified `ProjectConflict` and the guard placement, and ran the gates.

Verdict: **Comment** (merge-ready; leaving the formal Approve for the maintainer).

**Overview:** Six intents (workflow phase header, link-hijack guard, ingest GATHER step, `fest walk`, picker fallback, watch cycling). Two umbrella-level commits fix issues the per-intent reviews missed, and both are correct.

**Blocking issues — both fixed in this umbrella:**
- **`1f634bb`** completes the #201 hijack guard on the project→festival path (`linkProjectToFestival` previously called `SetLinkWithPath` with no `ProjectConflict` check, so an active festival's link could be silently re-pointed). The guard is symmetric with `linkFestivalToProject`, blocks only when the existing link is active, and threads `--force` through. Also fixes the walk UTF-8 byte-slice truncation and the over-broad `isActiveFestivalPath` (any component named `active`).
- **`eb7934b`** restores cwd-based location granularity regressed by #204 — `status`/`progress` were reporting `loc.Type = "festival"` from every subdir because `detectPath`/`targetPath` had been switched to the festival root. Picker branch correctly keeps the picked root.

**What's done well:**
- `ProjectConflict` is a clean, correct seam (conflict only for a different festival; conservative on a `nil`-link inconsistency).
- The `watch` cycle now bails with a validation error after one full stale rotation instead of spinning CPU.
- Honesty fix: `fest walk`'s help no longer claims "read-only" since `progress.NewManager` can migrate on first access.
- Each fix lands with a targeted regression test (hijack repro, subdir granularity, multibyte truncation, over-broad active match).

**Non-blocking follow-ups (not gating):**
- `fest walk` can still migrate legacy progress on first access; a true read-only mode would be cleaner than just documenting it.
- The hijack-guard test exercises `ProjectConflict`/`isActiveFestivalPath` directly rather than the picker-driven command (TTY); adequate but a seam-tested guard would be stronger.
- The `staleSince` logic in `runWatchCycle` is duplicated across the not-found and `nil` branches; a tiny helper would DRY it.

**Gate status (head `eb7934b`):** `go build ./...` clean · `just lint all` 0 issues · `just lint vet` clean · changed-package tests green (navigation, walk, status, progress, watch, guidance/workflow, navigation-state) including all new tests.

**Staff standard:** Yes — I would submit this as my own work. The two missed defects are now closed with correct, tested fixes, and the regression that #204 introduced is caught and reverted with granularity tests that would prevent a recurrence.

internal/commands/navigation/go_link.go
obey-agent: Resolved in eb7934b. The #201 guard is now applied in linkProjectToFestival (the in-a-project path) at go_link.go:274-286: when not --force, ProjectConflict + isActiveFestivalPath block re-linking a project away from its active festival before SetLinkWithPath, returning a Validation error with a --force hint. All three SetLinkWithPath call sites are now guarded. Test: TestLinkProjectToFestival_HijackGuardBlocks (asserts the conflict is detected and the active link is not silently evicted).

internal/commands/walk/walk.go
obey-agent: Fixed in eb7934b: loadGoal truncates on a rune boundary now (rune slice, not byte slice), so no dangling lead byte reaches the output. Test: TestLoadGoal_MultibyteTruncation.

internal/commands/navigation/link.go
obey-agent: Fixed in eb7934b: isActiveFestivalPath now checks the immediate parent dir, not any path component. Test: TestIsActiveFestivalPath_ImmediateParentOnly covers the workspace-under-ancestor-dir-named-active case.

Issue #206, opened 2026-06-16
fest: non-blocking follow-ups from intent-cleanup review (PR #205)


obey-agent:
Non-blocking follow-ups surfaced during the staff review of the intent-cleanup umbrella PR #205. None blocked that merge; tracking them here so they aren't lost.

### 1. Two files exceed the 500-line guideline (pre-existing)
- `internal/commands/status/handlers.go` (693 lines)
- `internal/guidance/workflow/navigator.go` (749 lines)

Both pre-existed on `main` (PR #205 added only +28/-6 and +14/-5 respectively), so this is pre-existing debt, not a regression. Split each into focused files (e.g. status handlers by responsibility; navigator parse/render/state).

### 2. `readCycleKey` leaks its stdin-read goroutine on cancellation
`internal/commands/watch/cycling.go` — when `ctx.Done()` wins the select, the goroutine blocked on `os.Stdin.Read` is abandoned (raw mode sets no read deadline). Acceptable for a short-lived CLI, but a cancellable reader or a read deadline would make `fest watch` cycling leak-free.

### 3. `fest walk` is not truly read-only
`fest walk` / `fest inspect` advertises orientation-only, but `progress.NewManager` → `store.Load()` can run `migrateFromLegacy()` / `migrateWorkflowYAML()` and write/delete files under `.fest/` on first access to a legacy festival. PR #205 made the help text honest about this ("May transparently migrate legacy progress formats on first access"). A read-only store path that skips migration would let `walk`/`inspect` genuinely never mutate.

### 4. Error-hint command inconsistency
`internal/commands/navigation/go_link.go:280` (in `linkProjectToFestival`) hints `'fgo link --force'`, while the sibling guard in `linkFestivalToProject` hints `'fest link --force'`. Make the two hints consistent (and use whichever command name is correct).

### 5. (minor) Double `cancelWatch()` on quit
`runWatchCycle` calls `cancelWatch()` inside `readCycleKey` (on quit) and again right after it returns. Idempotent, harmless — drop the redundant call for clarity.

---
Source: review of PR #205 (`eb7934b`). Items 1-5 are all non-blocking.

lancekrogers:
PRs opened for each follow-up item (all branched off `main`, gates green: `just lint all` 0 issues + `just test unit-raw` 91 packages):

| Item | PR | Notes |
|------|----|-------|
| 4 — `fgo`/`fest` link hint inconsistency | #207 | extracted shared `linkForceHint` constant |
[...]

PR #207, opened 2026-06-16
fix(link): unify active-festival link --force hint (#206 item 4)

lancekrogers:
## Summary

Follow-up item **4 of #206** (non-blocking findings from the intent-cleanup review of #205).

`linkProjectToFestival` (`go_link.go:280`) hinted `'fgo link --force'` while the sibling guard in `linkFestivalToProject` (`go_link.go:177`) hinted `'fest link --force'`. The user-facing subcommand is `fest link`, so the `fgo` variant was wrong.

## Change

Extract the shared remediation hint into a single `linkForceHint` constant used by both guard sites, so the two can't drift again. No behavior change beyond the corrected hint text.

Scope note: line 172's guard intentionally differs from line 274's `if !force` wrapper — that's separate logic and is left untouched here.

## Gates

- `just lint all` -> 0 issues
- `just test unit-raw` -> all packages pass

Refs #206 (item 4).

### DIFF
--- a/internal/commands/navigation/go_link.go
+++ b/internal/commands/navigation/go_link.go
@@ -27,6 +27,8 @@ var (
 	pathStyle     = lipgloss.NewStyle().Foreground(ui.MetadataColor)
 )
 
+const linkForceHint = "use 'fest link --force' to override, or run 'fest unlink' from the active festival first"
+
 // NewGoLinkCommand creates the context-aware link subcommand for fest go
 func NewGoLinkCommand() *cobra.Command {
 	var force bool
@@ -174,7 +176,7 @@ func linkFestivalToProject(ctx context.Context, cwd, targetPath string) error {
 			return festErrors.Validation("project is already linked to an active festival").
 				WithField("existing_festival", existing).
 				WithField("project", projectPath).
-				WithField("hint", "use 'fest link --force' to override, or run 'fest unlink' from the active festival first")
+				WithField("hint", linkForceHint)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** A small, correct, well-scoped fix. It corrects a user-facing bug (the `fest go link` / `fgo link` guard hinted `'fgo link --force'` while the sibling guard hinted `'fest link --force'`) and unifies both on the canonical top-level command by extracting a single `linkForceHint` constant so the two can't drift again.

**Verification (traced against the head tree):**
- `fest link` is a real top-level command (`internal/commands/navigation/link.go` `NewLinkCommand`, mounted in `root.go:219`) and it defines `--force` (`link.go:69`). So the unified hint `fest link --force` points at a command that exists and accepts the flag; the prior `fgo link --force` was the less-canonical form.
- Both edited guards (`linkFestivalToProject` go_link.go:176, `linkProjectToFestival` go_link.go:279) now reference the shared constant. `linkProjectToFestival` is shared by both the `fest link` and `fest go link` entry points (`link.go:95` calls it), so one canonical hint is correct for both.
- Uses the project error framework (`festErrors.Validation(...).WithField(...)`), not `fmt.Errorf`. No behavior change beyond the corrected hint text.

**Key findings:** None blocking.

**What's done well:** Tight scope with an explicit scope note, the DRY-via-constant approach genuinely prevents the two hints from drifting again, and the fix targets real user-facing correctness. Gates reported green (`just lint all`, `just test unit-raw`).

**Staff standard:** Yes, I would submit this as my own work. A minimal, correct fix that improves user-facing accuracy and removes a duplication footgun.

PR #208, opened 2026-06-16
fix(watch): leak-free cycle key reader + single cancel ownership (#206 items 2,5)

lancekrogers:
## Summary

Follow-up items **2 and 5 of #206** (non-blocking findings from the intent-cleanup review of #205).

### Item 2 — `readCycleKey` no longer leaks a stdin-read goroutine
Previously `readCycleKey` spawned a goroutine blocked on `os.Stdin.Read`; when `ctx.Done()` won the select, that goroutine was abandoned (raw mode sets no read deadline). For deadline-capable readers — the real `os.Stdin` terminal is an `*os.File` — it now reads in the **calling** goroutine, re-arming a short read deadline (`cycleKeyPollInterval`) each iteration so context cancellation is observed between reads with **no leaked read**. Readers that can't honor deadlines (the test `bytes.Reader`s) fall back to the original buffered-goroutine path, which is safe because the channel is buffered and the goroutine exits on EOF.

### Item 5 — single cancel ownership
Cancellation now lives entirely in `runWatchCycle`, which already calls `cancelWatch()` unconditionally after each key read (needed for next/prev to stop the current watch). `readCycleKey` is now a side-effect-free key reader: it no longer takes or invokes `cancelWatch`, removing the redundant double-cancel on quit.

## Tests
- Existing arrow/ctrl-c/cancelled-context tests updated for the new signature (now exercise the blocking fallback via `bytes.Reader`).
- New: deadline path (times-out-then-reads), deadline path respects a cancelled context, deadline-unsupported reader falls back, and a table test for `classifyCycleKey`.
- All pass under `go test -race`.

## Gates
- `just lint all` -> 0 issues
- `just test unit-raw` -> all packages pass

Refs #206 (items 2, 5).


internal/commands/watch/cycling.go
```diff
@@ -114,7 +117,55 @@ func cycleWatchOptions(opts options) show.WatchOptions {
...
+			_ = r.SetReadDeadline(time.Time{})
+		}
+	}()
+
+	buf := make([]byte, 3)
+	for {
+		if ctx.Err() != nil {
+			return cycleQuit, true
+		}
+		if err := r.SetReadDeadline(time.Now().Add(cycleKeyPollInterval)); err != nil {
```
obey-agent: **Major — the "leak-free" fix is a no-op on macOS, a primary (code-signed) release platform.**

`fest watch` reads from the terminal, and on macOS a TTY character device does not support `SetReadDeadline`. Confirmed empirically on Darwin under a real PTY:

```
/dev/tty SetReadDeadline err: file type does not support deadline
```

So this call errors on the first poll, `readCycleKeyWithDeadline` returns `(0, false)` (line 148), and `readCycleKey` falls through to `readCycleKeyBlocking` (line 131) — the exact abandoned-stdin-goroutine path #206 item 2 set out to remove. Linux TTYs are epoll-pollable so the deadline path works there; macOS (darwin amd64+arm64, both shipped and code-signed per release.just) silently uses the blocking fallback.

Accurate nuance: the leak is currently *dormant* on every platform, because the watch loop's ctx is never cancelled in normal use — the terminal is in raw mode (`term.MakeRaw`, cycling.go:36) and there is no `signal.NotifyContext`/`signal.Notify` anywhere in the repo. So there is no live goroutine leak today, but this fix won't protect macOS once signal-driven cancellation is wired, and the "leak-free" framing is already inaccurate there.

Fix options: (a) make cancellation work cross-platform (a self-pipe the blocking read also selects on, or closing the fd to unblock the pending read), or (b) keep the bounded blocking fallback but correct the PR/commit framing to "leak-free on deadline-capable readers (Linux TTY); macOS uses the bounded blocking fallback," and add the regression test noted on the test file.

internal/commands/watch/cycling.go
```diff
@@ -125,31 +176,35 @@ func readCycleKey(ctx context.Context, r io.Reader, cancelWatch context.CancelFu
...
+	}
+	if len(b) >= 3 && b[0] == 0x1b && b[1] == '[' {
+		switch b[2] {
+		case 'C':
+			return cycleNext, true
+		case 'D':
+			return cyclePrev, true
+		}
+	}
+	return 0, false
```
obey-agent: **Minor (latent footgun) — the bare `0` "no value" sentinel aliases a real direction.** `cycleNext = iota` is `0` (cycling.go:17), and `readCycleKeyWithDeadline`/`readCycleKeyBlocking` return bare `0` as their "unsupported / cancelled / no key" value (lines 148, 196, 209). Safe today only because every caller checks the companion `bool` first — but a future caller that drops the bool would silently treat "no key" as "advance to the next festival." Add an explicit zero sentinel (e.g. `cycleNone cycleDirection = iota` first, shifting the real values) and return that named constant instead of bare `0`. Magic-number-as-sentinel is exactly what the const block exists to prevent.

internal/commands/watch/cycling_test.go
```diff
@@ -51,51 +54,117 @@ func TestRunWatchCycle_SinglePath_CallsWatch(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
+	r := &fakeDeadlineReader{timeouts: 100}
+	dir := readCycleKey(ctx, r)
+	if dir != cycleQuit {
+		t.Errorf("readCycleKey on cancelled ctx = %v, want cycleQuit", dir)
+	}
+}
+
+func TestReadCycleKey_DeadlineReader_FallsBackWhenUnsupported(t *testing.T) {
```
obey-agent: **No test proves the goroutine actually exits, so the leak this PR targets is unverified.** This fallback test feeds a reader that returns data immediately, so the blocking goroutine exits naturally regardless of any leak fix. No test exercises the precise leak condition: ctx cancelled while the blocking read is still pending. A concurrency fix without a test that fails on the pre-fix version isn't a verified fix, and since fest is in production this is the regression guard that matters.

Add a test with a reader that blocks forever (e.g. backed by a channel that never sends) under a cancellable ctx: sample `runtime.NumGoroutine()` before, cancel, then poll until it returns to baseline within a timeout. That both documents the intended cancellation semantics and locks them in (and would currently fail on macOS, surfacing the platform gap above).

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The single-cancel-ownership half (item 5) is genuinely good — `readCycleKey` is now a side-effect-free reader and the sole `cancelWatch()` lives in `runWatchCycle`, removing the prior double-cancel — and the extracted `classifyCycleKey` is clean and well-tested. But the headline deliverable, a leak-free cycle key reader (item 2), is not achieved on macOS and is unverified everywhere.

**Major (inline on cycling.go:147):** the deadline-based reader relies on `os.Stdin.SetReadDeadline`, which macOS TTYs do not support — confirmed empirically on Darwin under a real PTY (`/dev/tty SetReadDeadline → "file type does not support deadline"`). So on macOS (a shipped, code-signed platform) the code falls back to `readCycleKeyBlocking`, the exact abandoned-goroutine path this PR claims to remove. Honest nuance: the leak is currently *dormant* on all platforms because the watch ctx is never cancelled in normal use (raw mode via `term.MakeRaw`, and no `signal.NotifyContext`/`signal.Notify` anywhere in the repo) — so no live leak today, but the fix won't protect macOS once cancellation is wired, and "leak-free" is already inaccurate there.

**Also (inline):** no test asserts the goroutine exits on cancellation (the one "fallback" test uses an immediately-returning reader, so it can't catch the leak), and the bare `0` sentinel aliases `cycleNext` (iota 0) — safe today only because callers check the companion bool.

**Fix:** either implement cancellation that works cross-platform (self-pipe the read also selects on, or close the fd to unblock), or correct the "leak-free" framing to the Linux-only reality and document the macOS fallback — and in both cases add the goroutine-exit regression test. It's also worth deciding whether the leak is worth fixing at all before signal-driven cancellation exists.

**What's done well:** Single cancel ownership is correctly achieved and verified; `classifyCycleKey` is a clean pure function with a thorough table test; the deadline path's hygiene is correct where deadlines work (guarded deferred reset, `os.IsTimeout` re-arm, `ctx.Err()` before each read); no `fmt.Errorf` introduced; functions stay under 50 lines and the suite passes under `-race`.

**Staff standard:** Not yet. The refactor is careful and the cancel-ownership fix is solid, but a "leak-free" deliverable that no-ops on the maintainer's primary platform and has no test proving the goroutine exits isn't something I'd ship as-is. Make it cross-platform or correct the framing, add the regression test, and it's an Approve.

lancekrogers:
Addressed in b665953. Took the cross-platform option (real leak-free cancellation), not the framing-correction option.

- Replaced the `SetReadDeadline` reader with a poll-based one (`//go:build unix`, cycling_unix.go) that polls the terminal fd alongside a self-pipe closed on ctx cancellation. A cancelled context wakes `unix.Poll` immediately and leaves no reader parked, on Linux and macOS TTYs alike. A `!unix` shim (cycling_other.go) keeps cross-compilation building; fest ships only unix targets.
- `readCycleKey` now takes `*os.File` (call site in `runWatchCycle` unchanged: still `os.Stdin`).
- Added `cycleNone` as the zero-value direction, removing the `0`-aliases-`cycleNext` sentinel trap; `classifyCycleKey` returns `(cycleNone, false)`.
- Added `TestReadCycleKey_CancelReturnsQuitNoLeak` (the goroutine-exit regression test). Verified it FAILS when forced down the blocking-fallback path (NumGoroutine stays above baseline) and passes on the poll path. Added real-fd arrow/Ctrl-C/EOF/already-cancelled tests.
- `golang.org/x/sys` promoted to a direct dependency.

Gates: `just lint all` clean, `just test unit-raw` green, watch package green under `-race`, `GOOS=windows go vet` of the fallback clean.

internal/commands/watch/cycling_unix.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"context"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// readCycleKey blocks until the user presses a recognized cycle key
+// (→ next, ← previous, q/Ctrl-C quit) or ctx is cancelled, whichever comes
```
obey-agent: This doc comment says `q/Ctrl-C quit`, but `q` is not handled: `classifyCycleKey` maps only `0x03` (Ctrl-C) and the arrow sequences, and the watch footer (`show/watch.go:186`) advertises only "Ctrl+C to exit • ← → cycle festivals". So the comment is inaccurate. More broadly, the watch package is comment-sparse by convention (resolver.go/command.go: ~2 comment-lines each, no per-func doc comments) and the repo default is zero explanatory comments. Please drop the multi-line doc comments added across cycling.go / cycling_unix.go / cycling_other.go and remove this inaccurate one. The commit message already carries the "why" well, which is where it belongs.

internal/commands/watch/cycling_unix_test.go
```diff
@@ -0,0 +1,120 @@
...
+	defer func() { _ = w.Close() }()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	if dir := readCycleKey(ctx, r); dir != cycleQuit {
+		t.Errorf("readCycleKey on cancelled ctx = %v, want cycleQuit", dir)
+	}
+}
+
+// TestReadCycleKey_CancelReturnsQuitNoLeak is the regression test for the
```
obey-agent: Per the repo convention, tests carry no rationale preamble (no "regression for X" explanation). This 5-line block should come out: the name `TestReadCycleKey_CancelReturnsQuitNoLeak` already states intent and the why lives in the commit/PR. The test logic itself is exactly right: NumGoroutine baseline, cancel mid-poll, assert prompt `cycleQuit` and return to baseline under `-race`.

obey-agent (changes requested):
Staff re-review (obey-agent) at head `b665953`. I read the full diff, traced the cancellation path, and ran the gate under `-race`.

Verdict: **Request Changes** (the engineering is complete and correct; the only remaining item is comment cleanup).

## All three prior findings are resolved

- **Leak-free reader (item 2), the headline deliverable, is now genuinely cross-platform.** `cycling_unix.go` polls the terminal fd alongside a self-pipe that is closed when ctx is cancelled (`unix.Poll`), so cancellation wakes the poll immediately and no goroutine is left parked on the terminal. `unix.Poll` works on macOS TTYs (unlike `SetReadDeadline`), so the macOS no-op the prior review flagged is gone. The helper goroutine is guaranteed to exit via `defer close(stop)` (LIFO, before the `readEnd` close). `go.mod` cleanly promotes `x/sys` from indirect to direct at the same version (already transitive via `x/term`), no new dependency.
- **Sentinel (minor) fixed.** `cycleNone = iota` is now the explicit zero value and `classifyCycleKey` returns it; the 0-aliases-`cycleNext` trap is gone.
- **Regression test added and correct.** `TestReadCycleKey_CancelReturnsQuitNoLeak` exercises the exact leak condition (ctx cancelled while the read is pending) and asserts the goroutine count returns to baseline. Passes under `-race -count=1`; the full watch package is green under `-race`, build + vet clean.

## Blocking: comment cleanup

The new files add multi-line doc/explanatory comments throughout, which diverges from the watch package's comment-sparse convention (`resolver.go`: 2 comment-lines/267, no per-func doc comments) and the repo's no-comments default. Two inline comments mark the most important:

- `cycling_unix.go:13` is **factually wrong**: it says `q/Ctrl-C quit`, but `q` is not handled and the footer advertises only Ctrl+C + arrows. A misleading comment is worse than none.
- `cycling_unix_test.go:83` is a test-rationale preamble.

Strip the added doc comments across `cycling.go` / `cycling_unix.go` / `cycling_other.go` / `cycling_unix_test.go` (that removes the inaccurate one too). The commit message already carries the "why".

## What's done well

The self-pipe + `unix.Poll` design is the right cross-platform fix, the build-tag split is clean, the fallback path is honestly scoped to fd-exhaustion, and the regression test pins the exact condition that was unverified before. No `fmt.Errorf`; functions stay small.

## Staff standard

The engineering, yes, I would ship this design. Strip the comments to match the package and remove the inaccurate one, and it is a clean Approve.

obey-agent (commented):
**Verdict: Approve**

Overview: This PR fixes the watch-cycle cancellation ownership cleanly and replaces the stdin-read goroutine with a Unix poll/self-pipe path, which addresses the leak without leaving a reader blocked on the terminal. The `cycleNone` split also removes the old zero-value ambiguity around unrecognized input.

Key Findings: No blocking findings. I specifically reviewed the cancellation path, fd/poll behavior, fallback path, and key classification behavior.

What's Done Well: The leak regression is covered with a cancellation test that waits for the reader to return, and the input classifier is isolated enough to make the direction mapping easy to reason about.

Staff Standard: Yes, I would be comfortable submitting this as my own work. Focused verification run: `go test ./internal/commands/watch` on PR head `b6659535730db7bb20d1524f0a520ab67769a4c6` passed.

lancekrogers:
Addressed in 8be0cff (comment cleanup, the only remaining blocker).

Stripped the doc/inline comments added in this PR across `cycling.go`, `cycling_unix.go`, `cycling_other.go`, and `cycling_unix_test.go` to match the watch package's comment-sparse convention and the repo's no-comments default. That removes the factually-wrong one at `cycling_unix.go:13` that listed `q` as a quit key (only Ctrl-C and arrows are handled). No behavior change.

Gates re-run at head 8be0cff: `just lint all` clean, watch package green under `-race`, host + `GOOS=windows` builds clean.

obey-agent (commented):
**Verdict: Approve**

Overview: Re-reviewed the current PR head, which now lands on the Unix poll/self-pipe implementation rather than the earlier deadline-based approach in the patch history. The code removes cancellation side effects from `readCycleKey`, keeps cancellation ownership in `runWatchCycle`, and avoids leaving a goroutine parked on stdin during normal Unix cancellation.

Key Findings: No blocking findings in the current code. One non-blocking note: parts of the PR description still describe the earlier SetReadDeadline approach, while the current implementation is poll/self-pipe; updating the PR body would make future archaeology less confusing.

What's Done Well: The implementation is tightly scoped to cycle-key reading, keeps the non-Unix fallback isolated, and adds a direct cancellation/no-leak regression test.

Staff Standard: Yes, I would be comfortable submitting this as my own work. Focused verification run on PR head `8be0cfffc5fe69541be565f57d86f75edc51b443`: `go test ./internal/commands/watch` passed.

PR #209, opened 2026-06-16
fix(walk): make fest walk/inspect truly read-only (#206 item 3)

lancekrogers:
## Summary

Follow-up item **3 of #206** (non-blocking findings from the intent-cleanup review of #205).

`fest walk` / `fest inspect` advertises read-only orientation, but on first access to a **legacy** festival it mutated disk: it converted `.fest/progress.yaml` to JSONL and consumed `.fest/workflow_state.yaml`. PR #205 had only made the help text honest about this; this PR makes the command genuinely read-only.

## Root cause (found via a real-binary smoke, not just unit tests)

Walk's own progress loads were not the whole story. The mutation came from the **shared detection path**:

```
walk -> show.DetectCurrentFestival -> parseFestivalInfo -> CalculateFestivalStats -> progress.NewManager -> Store.Load (migrates)
```

`CalculateFestivalStats` is display-only but used the migrating `Load`, so *any* festival detection rewrote a legacy festival's `.fest/`. Unit tests that called the new read-only path directly passed while the binary still mutated — caught by a real `fest walk` smoke.

## Change

- **`Store.LoadReadOnly`** — materializes task/time/workflow state in memory from the same synthetic-event stream `Load` produces *after* migration, but never writes JSONL or removes YAML. Factored `parseEventsFile`/`materializeFrom` out of `loadFromEvents`, and `parseWorkflowYAML` out of `migrateWorkflowYAML`, to share materialization (no behavior change to `Load`).
- **`NewManagerReadOnly`** wraps it.
- **`walk`** uses `NewManagerReadOnly` for its own loads; help text now states it never mutates.
- **`CalculateFestivalStats`** loads read-only — computing a festival's stats must never rewrite its `.fest/`. This also makes the detection path read-only for `show`/`status`/`list` (a strict improvement; genuine migration still happens when a mutation command loads progress in its own body).

## Verification

Real binary against a legacy festival (`progress.yaml` + `workflow_state.yaml`, no JSONL):
- `fest walk` and `fest walk --json` -> both YAMLs unchanged (checksums equal), **no** `progress_events.jsonl` written, progress still reported correctly (`1/1 tasks (100%)`).

Tests:
- `TestLoadReadOnly_DoesNotMutateLegacyFiles` / `TestLoadReadOnly_MatchesMigratingLoad` — read-only view equals the migrating `Load`'s, while the migrating `Load` consumes the YAML and read-only preserves it.
- `TestCalculateFestivalStats_DoesNotMutate` — the precise regression guard; **fails** against the old migrating load, passes now.

## Gates
- `just lint all` -> 0 issues
- `just test unit-raw` -> 91 packages pass

Refs #206 (item 3).

### DIFF
--- a/internal/commands/show/stats.go
+++ b/internal/commands/show/stats.go
@@ -73,8 +73,9 @@ func CalculateFestivalStats(ctx context.Context, festivalDir string) (*FestivalS
 	festivalRoot := resolveFestivalRoot(festivalDir)
 
 	// Manager is the single source of truth for progress numbers.
-	// Keep a reference to delegate task totals after the structural walk.
-	mgr, mgrErr := progress.NewManager(ctx, festivalRoot)
+	// Stats are display-only, so load read-only: computing a festival's stats
+	// must never migrate or rewrite its .fest/ files as a side effect.
+	mgr, mgrErr := progress.NewManagerReadOnly(ctx, festivalRoot)
 	var store *progress.Store
 	if mgrErr == nil {
 		store = mgr.Store()
--- a/internal/commands/show/stats_readonly_test.go
+++ b/internal/commands/show/stats_readonly_test.go
@@ -1,0 +1,59 @@
+package show
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/Obedience-Corp/fest/internal/progress"
+)
+
+func TestCalculateFestivalStats_DoesNotMutate(t *testing.T) {
+	festivalDir := t.TempDir()
+
+	if err := os.WriteFile(filepath.Join(festivalDir, "fest.yaml"), []byte("name: stats-ro\nmetadata:\n  id: SR0001\n"), 0o644); err != nil {
+		t.Fatalf("write fest.yaml: %v", err)
+	}
+	taskDir := filepath.Join(festivalDir, "001_PLAN", "01_seq")
+	if err := os.MkdirAll(taskDir, 0o755); err != nil {
+		t.Fatalf("mkdir task dir: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(taskDir, "01_task.md"), []byte("# Task\n"), 0o644); err != nil {
+		t.Fatalf("write task: %v", err)
+	}
+
+	festDir := filepath.Join(festivalDir, progress.ProgressDir)
+	if err := os.MkdirAll(festDir, 0o755); err != nil {
+		t.Fatalf("mkdir .fest: %v", err)
+	}
+	legacyYAML := "festival: stats-ro\nupdated_at: 2026-01-15T10:00:00Z\ntasks:\n" +
+		"  001_PLAN/01_seq/01_task.md:\n" +
+		"    task_id: 001_PLAN/01_seq/01_task.md\n" +
+		"    status: completed\n" +
+		"    progress: 100\n" +
+		"    started_at: 2026-01-15T09:00:00Z\n" +
+		"    completed_at: 2026-01-15T09:30:00Z\n"
+	legacyPath := filepath.Join(festDir, progress.ProgressFileName)
+	if err := os.WriteFile(legacyPath, []byte(legacyYAML), 0o644); err != nil {
+		t.Fatalf("write legacy progress: %v", err)
+	}
+
+	stats, err := CalculateFestivalStats(context.Background(), festivalDir)
+	if err != nil {
+		t.Fatalf("CalculateFestivalStats: %v", err)
+	}
+	if stats == nil {
+		t.Fatal("stats is nil")
+	}
+
+	if _, err := os.Stat(legacyPath); err != nil {
+		t.Errorf("legacy progress.yaml was removed by stats calculation: %v", err)
+	}
+	if _, err := os.Stat(filepath.Join(festDir, progress.ProgressEventsFile)); !os.IsNotExist(err) {
+		t.Error("stats calculation wrote progress_events.jsonl; it must not mutate disk")
+	}
+	if stats.Tasks.Total < 1 {
+		t.Errorf("stats.Tasks.Total = %d, want >= 1 (read-only load must still report progress)", stats.Tasks.Total)
+	}
+}
--- a/internal/commands/show/stats_readonly_test.go
+++ b/internal/commands/show/stats_readonly_test.go
@@ -5,8 +5,11 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
+	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
 	"github.com/Obedience-Corp/fest/internal/progress"
+	"gopkg.in/yaml.v3"
 )
 
 func TestCalculateFestivalStats_DoesNotMutate(t *testing.T) {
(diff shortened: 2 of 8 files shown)


internal/commands/show/stats.go
```diff
@@ -73,8 +73,9 @@ func CalculateFestivalStats(ctx context.Context, festivalDir string) (*FestivalS
 	festivalRoot := resolveFestivalRoot(festivalDir)
 
 	// Manager is the single source of truth for progress numbers.
-	// Keep a reference to delegate task totals after the structural walk.
-	mgr, mgrErr := progress.NewManager(ctx, festivalRoot)
+	// Stats are display-only, so load read-only: computing a festival's stats
+	// must never migrate or rewrite its .fest/ files as a side effect.
+	mgr, mgrErr := progress.NewManagerReadOnly(ctx, festivalRoot)
```
obey-agent, line 78: **Blocker — the read-only conversion is incomplete; `fest walk`/`inspect` still mutate `.fest/` for any festival with a `WORKFLOW.md` phase.**

This makes the *top-level* stats load read-only, but `CalculateFestivalStats` → `calculatePhaseStats` (stats.go:162, `HasWorkflowFile`) → `calculateWorkflowPhaseStats` (stats.go:329) → `shared.LoadWorkflowStepsForPhase` (workflow_render.go:113) still calls the **migrating** `store.Load` (workflow_render.go:135). That runs `migrateFromLegacy`/`migrateWorkflowYAML`, which `os.Remove` `progress.yaml`/`workflow_state.yaml` (store.go:266, :666) and write `progress_events.jsonl`.

`fest walk` (alias `inspect`) reaches this unconditionally: `runWalk` → `resolveFestival` → `show.DetectCurrentFestival` → `parseFestivalInfo` → `CalculateFestivalStats` (detect.go:453). So on the common legacy festival whose INGEST/PLAN/RESEARCH phase ships a `WORKFLOW.md`, a "read-only" walk rewrites on-disk state. (The same hole makes `fest status` mutate via status/collectors.go:61, and `fest show` via the same detect path.)

Verified end-to-end against head `b7cdb63` (the full call chain plus the `os.Remove`/write sites), so this is the common case, not an edge.

Fix: convert the two `store.Load` calls in `LoadWorkflowStepsForPhase` (workflow_render.go:135) and the sibling gates loader (workflow_render.go:230) to `store.LoadReadOnly(ctx)` — every caller of `CalculateFestivalStats` is a display surface, so a read-only load is correct there.

internal/commands/show/stats_readonly_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/Obedience-Corp/fest/internal/progress"
+)
+
+func TestCalculateFestivalStats_DoesNotMutate(t *testing.T) {
+	festivalDir := t.TempDir()
+
+	if err := os.WriteFile(filepath.Join(festivalDir, "fest.yaml"), []byte("name: stats-ro\nmetadata:\n  id: SR0001\n"), 0o644); err != nil {
+		t.Fatalf("write fest.yaml: %v", err)
+	}
+	taskDir := filepath.Join(festivalDir, "001_PLAN", "01_seq")
```
obey-agent, line 21: **The regression guard doesn't cover the path that still mutates.** This test sets up a *task-based* phase (`001_PLAN/01_seq/01_task.md`), so it exercises `calculatePhaseStats`'s task branch — not the `WORKFLOW.md` branch that goes through `calculateWorkflowPhaseStats` → `LoadWorkflowStepsForPhase` → the migrating `store.Load`. As written it passes green even though `walk`/`inspect` still mutate workflow-phase festivals.

Add a case with a phase directory containing a `WORKFLOW.md` plus an on-disk `workflow_state.yaml` (and/or legacy `progress.yaml`), then assert both files still exist after `CalculateFestivalStats`. That is the test that actually pins the "truly read-only" claim and would have caught this.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** A clean, well-factored partial fix — the direct `walk` manager loads and `CalculateFestivalStats`'s own manager load are now genuinely read-only, and the new `LoadReadOnly` + `TestLoadReadOnly_MatchesMigratingLoad` equivalence test is excellent. But the PR's headline guarantee ("walk/inspect never mutate festival state") is not met for the common case of a festival containing a `WORKFLOW.md` phase.

**Blocker (inline on stats.go:78):** `fest walk`/`inspect` still mutate `.fest/`. `CalculateFestivalStats` → `calculatePhaseStats` (HasWorkflowFile, stats.go:162) → `calculateWorkflowPhaseStats` (stats.go:329) → `shared.LoadWorkflowStepsForPhase` (workflow_render.go:113) calls the *migrating* `store.Load` (workflow_render.go:135), which `os.Remove`s `progress.yaml`/`workflow_state.yaml` (store.go:266/666) and writes JSONL. `fest walk` reaches it unconditionally via detect.go:453. I verified this end-to-end against head `b7cdb63` (call chain + the mutation sites); every standard festival has at least one `WORKFLOW.md` phase, so this is the common case. The same hole also makes `fest status` (status/collectors.go:61) and `fest show` silently mutate — worth fixing together.

**Why it slipped:** the PR's smoke evidently used a festival with no `WORKFLOW.md` phase, and the new regression test (stats_readonly_test.go) uses a task-based phase, so neither covers the migrating branch (inline on the test file).

**Fix:** point `LoadWorkflowStepsForPhase` (workflow_render.go:135) and the sibling gates loader (:230) at `store.LoadReadOnly(ctx)`. Then extend stats_readonly_test.go with a `WORKFLOW.md`-phase + `workflow_state.yaml` case asserting no mutation. Optional hardening: make read-only mode structurally unable to write (a `readOnly` flag that no-ops `Save`/append), so the contract can't silently regress — today `GetFestivalProgress`'s `LazyPopulateTimeData`→`Save` (aggregate.go:129) is safe only because `Save` no-ops on empty pending events (store.go:314), not by construction.

**What's done well:** The Load/LoadReadOnly refactor is clean — extracting the shared materialization so the read-only view provably matches the migrating view, with a real equivalence test. Context/cancellation are threaded correctly through the new paths, new code uses `internal/errors`, and the root-cause framing was right; it just stopped one frame short of the workflow-phase sub-path.

**Staff standard:** Not yet. Solid engineering, but shipping a "truly read-only" guarantee that still rewrites `.fest/` for most real festivals is worse than the honest help text it replaces, because users will now trust it. Close the workflow-phase path and add the covering test, and this is an Approve.

lancekrogers:
Addressed in abb3317.

The review correctly caught that fixing CalculateFestivalStats's own load was insufficient: `calculateWorkflowPhaseStats` -> `shared.LoadWorkflowStepsForPhase` (and the sibling `LoadGateStepsForPhase`) still called the migrating `store.Load`, so a WORKFLOW.md/GATES.md phase was migrated on display.

- `LoadWorkflowStepsForPhase` (workflow_render.go:135) and `LoadGateStepsForPhase` (:231) now call `store.LoadReadOnly`. All callers are display-only (walk, show stats, show tree, status presenters, progress display).
- Added `TestCalculateFestivalStats_WorkflowPhaseDoesNotMutate` covering the WORKFLOW.md-phase path; verified it FAILS against the migrating load (workflow_state.yaml consumed, progress_events.jsonl written) and passes after the fix.
- Real-binary smoke: `fest walk` against a festival with a WORKFLOW.md phase + workflow_state.yaml preserves the YAML and writes no JSONL.

Gates: `just lint all` clean, `just test unit-raw` green.

obey-agent (commented):
Staff re-review (obey-agent) at head `abb3317`. I read the full diff, traced every display caller of the affected loaders, and ran the gate under `-race`.

Verdict: **Comment, merge-ready** (leaving the formal Approve to the maintainer).

## Both prior blocking findings are resolved

- **The read-only conversion is now complete.** `LoadWorkflowStepsForPhase` (`workflow_render.go:136`) and `LoadGateStepsForPhase` (`:231`) both call `store.LoadReadOnly(ctx)`, so rendering a `WORKFLOW.md`/`GATES.md` phase no longer migrates or rewrites `.fest/`. I traced every display caller of those loaders, `show/tree`, `show/stats`, `progress/display`, `status/collectors` + `presenters`, and `walk` via `detect` → `CalculateFestivalStats`, and all now route through the read-only load. The top-level stats load uses `NewManagerReadOnly`.
- **The regression guard now covers the path that mutated.** `TestCalculateFestivalStats_WorkflowPhaseDoesNotMutate` sets up a `WORKFLOW.md` phase with an on-disk `workflow_state.yaml` and asserts both that the yaml survives and that no `progress_events.jsonl` is written. It fails on the pre-fix migrating load and passes now. The `LoadReadOnly` equivalence tests still pass.

Verified end-to-end at head `abb3317`: build clean, and the progress / show / walk / status / shared packages are all green under `-race`.

## What's done well

The `Load` / `LoadReadOnly` split with a real equivalence test (`TestLoadReadOnly_MatchesMigratingLoad`) is exactly the right shape: the read-only view provably matches the migrating view, so the contract can't silently drift. Context/cancellation are threaded through the new paths, new code uses `internal/errors`, and the fix closed the workflow-phase sub-path that the prior pass stopped one frame short of.

## Staff standard

Yes, I would ship this. Clean root-cause fix, complete across the display surfaces, with a test that actually fails on the pre-fix code.

internal/progress/store.go
```diff
@@ -157,6 +157,54 @@ func (s *Store) Load(ctx context.Context) error {
...
+		parsed, err := s.parseEventsFile(ctx)
+		if err != nil {
+			return err
+		}
+		events = parsed
+	case fileExists(s.legacyFilePath()):
+		if err := s.loadLegacyYAML(ctx); err != nil {
+			return errors.Wrap(err, "loading legacy progress format")
+		}
+		events = generateEventsFromState(s.data.Tasks)
```
obey-agent: This read-only path drops valid legacy task state when the old `progress.yaml` entry has no timestamps. Existing backward-compatibility coverage on `main` accepts legacy YAML like `status: completed` + `progress: 100` without `started_at`/`completed_at`; for that shape, `generateEventsFromState` emits no events and `materializeFrom` overwrites the YAML-loaded `s.data`, so `LoadReadOnly` returns an empty task map. That makes read-only commands under-report progress for the legacy festivals this PR is meant to preserve. Keep the YAML-loaded task state for the legacy read-only case, or make the synthetic conversion lossless for timestamp-free completed/in-progress/blocked tasks, and add a regression test for legacy progress without timestamps.

obey-agent (commented):
**Verdict: Request Changes**

Overview: The goal is correct: orientation/display paths should not migrate or delete legacy `.fest/` files as a side effect. The implementation covers the obvious timestamp-rich legacy fixtures and moves important display callers to read-only loading, but the new `LoadReadOnly` path is lossy for valid legacy `progress.yaml` entries that do not include timestamps.

Key Findings:
- Inline comment on `internal/progress/store.go`: `LoadReadOnly` loads legacy YAML, converts it through `generateEventsFromState`, then overwrites `s.data` via `materializeFrom`. Existing backward-compatible legacy YAML can contain `status: completed` + `progress: 100` without `started_at`/`completed_at`; that shape produces no synthetic events, so the task disappears from read-only progress views.

What's Done Well: The PR found the right integration point by making `CalculateFestivalStats`, workflow rendering, gate rendering, and `walk` use read-only loading. The non-mutation assertions for `progress.yaml`, `workflow_state.yaml`, and `progress_events.jsonl` are the right regression target.

Staff Standard: Not yet. I would want the legacy read-only load to preserve all accepted legacy task state, plus a regression test for timestamp-free legacy progress, before merging. Focused verification on PR head `abb331760f673a5cc7d4250f6c5d1564100e5892` passed for the submitted tests (`go test ./internal/progress ./internal/commands/show ./internal/commands/walk ./internal/commands/shared`), and an added scratch regression for timestamp-free legacy progress failed as described.

lancekrogers:
Addressed in acf89f0.

Good catch: `LoadReadOnly` round-tripped legacy `progress.yaml` through `generateEventsFromState` then `materializeFrom`, and that path emits no events for accepted legacy entries that omit `started_at`/`completed_at` (e.g. `status: completed` + `progress: 100`), so those tasks vanished. The migrating `Load` keeps them in memory because `migrateFromLegacy` sets `s.data` directly via `loadLegacyYAML` and never re-materializes.

`LoadReadOnly` now loads legacy task state directly via `loadLegacyYAML` and parses `workflow_state.yaml` read-only, matching `Load`'s in-memory result without the lossy round trip. JSONL stays authoritative and is materialized from events as before.

Added `TestLoadReadOnly_PreservesTimestampFreeLegacyTasks`; verified it FAILS against the round-trip path (`ok=false task=<nil>`) and passes after the fix. `TestLoadReadOnly_MatchesMigratingLoad` still passes (parity preserved).

Gates at head acf89f0: `just lint all` clean, `just test unit-raw` green.

obey-agent (commented):
**Verdict: Approve**

Overview: Re-reviewed the current head after the previous legacy-read-only blocker. `LoadReadOnly` now keeps legacy `progress.yaml` task state directly instead of forcing it through synthetic events, which preserves accepted timestamp-free legacy entries while still avoiding JSONL creation or YAML deletion.

Key Findings: The prior blocker is resolved. I did not find a remaining blocking issue in the current read-only load path, stats path, workflow/gate rendering path, or `walk` integration.

What's Done Well: The new `TestLoadReadOnly_PreservesTimestampFreeLegacyTasks` covers the exact lossy shape from the previous review, and the display callers now consistently use read-only loading where migration would be surprising.

Staff Standard: Yes, I would be comfortable submitting this as my own work. Focused verification run on PR head `acf89f0ea1dc313d73bcd8211d24a57849ca3243`: `go test ./internal/progress ./internal/commands/show ./internal/commands/walk ./internal/commands/shared` passed.

PR #211, opened 2026-06-16
fix(watch): keep cursor column-aligned in raw-mode festival cycling

lancekrogers:
## Problem

After merging the recent fest changes, `fest watch` renders as a garbled diagonal staircase that wraps into unreadable text (screenshot of the symptom: the progress tree cascades down-right and wraps mid-word).

## Root cause

The festival-cycling feature added in **#204** puts the terminal into **raw mode** (`term.MakeRaw` in `internal/commands/watch/cycling.go`) so it can read the `← →` arrow keys for cycling. `MakeRaw` clears `OPOST`/`ONLCR`, which is the terminal flag that maps a bare `\n` to `\r\n` on output. With it off, every `\n` the renderer emits moves the cursor **down but not back to column 0**, so each line starts where the previous one ended. That is exactly the staircase.

It only reproduces with **2+ watchable festivals** in a real terminal — the single-festival path (`watchFestivalAtPath`) and the non-TTY path both skip raw mode, which is why it renders fine when piped or with one festival. This is a regression from #204 (before it, `fest watch` never entered raw mode).

It is **not** fixed by either open PR: #208 fixes a goroutine leak in the key reader (`cycling.go` only), #209 is `fest walk` read-only. Neither touches the render, and the #206 follow-up issue doesn't list it.

## Fix

Route the watch render output through a small writer that translates a bare `\n` → `\r\n` while in cycle (raw) mode, so each line returns to column 0. `clearScreen`, `renderFestivalView`, and `printWatchFooter` now take an `io.Writer`; the cycle path gets the translating writer (`watchWriter(true)`), every other path (including standalone watch, which never enters raw mode) gets plain `os.Stdout`.

This keeps the fix portable (no platform-specific termios manipulation, no new dependencies) and confined to the `show` package. The cycle key reader and raw-mode setup in `cycling.go` are untouched.

## Changes

- `internal/commands/show/watch.go` — add `crlfWriter` + `watchWriter`; thread an `io.Writer` through `runWatchMode`, `runPollingMode`, `renderFestivalView`, `clearScreen`, `printWatchFooter`.
- `internal/commands/show/standalone_watch.go` — pass `os.Stdout` to the now-parameterized helpers (standalone watch never enters raw mode).
- `internal/commands/show/watch_test.go` (new) — CRLF translation, no-double-CRLF, escape-sequence passthrough, `watchWriter` mode selection, and that cycle-mode footer output contains no bare `\n`.

## Verification

- `go build ./...` — clean
- `just lint all` — **0 issues**; `just lint vet` — clean
- `go test ./internal/commands/show/... ./internal/commands/watch/...` — green, incl. the new CRLF tests
- The render path is unchanged except for the output writer, so `fest show`, non-cycle `fest watch`, and standalone watch are unaffected.

**Please confirm in your terminal** with a multi-festival campaign (`fest watch`, then `← →` to cycle) — that's the only path that exercises raw mode, and it can't be fully reproduced in a non-PTY test harness. Immediate workaround if needed before merge: `fest watch --collapsed` renders a shallow tree that fits any width.

### DIFF
--- a/internal/commands/show/watch.go
+++ b/internal/commands/show/watch.go
@@ -32,19 +32,50 @@ func (c crlfWriter) Write(p []byte) (int, error) {
 		}
 		b.WriteByte(p[i])
 	}
-	if _, err := c.w.Write(b.Bytes()); err != nil {
-		return 0, err
+	translated := b.Bytes()
+	written, err := c.w.Write(translated)
+	if written == len(translated) && err == nil {
+		return len(p), nil
 	}
-	return len(p), nil
+
+	n, remaining := 0, written
+	for i := 0; i < len(p); i++ {
+		cost := 1
+		if p[i] == '\n' && (i == 0 || p[i-1] != '\r') {
+			cost = 2
+		}
+		if remaining < cost {
+			break
+		}
+		remaining -= cost
+		n++
+	}
+	if err == nil && n < len(p) {
+		err = io.ErrShortWrite
+	}
+	return n, err
 }
 
-// watchWriter returns the writer for a watch frame. In cycle mode the terminal
-// is in raw mode, so newlines must be translated to keep lines column-aligned.
+// watchWriter returns the stdout writer for a watch frame. In cycle mode the
+// terminal is in raw mode, so newlines must be translated to keep lines
+// column-aligned.
 func watchWriter(cycleHint bool) io.Writer {
+	return cycleOutput(cycleHint, os.Stdout)
+}
+
+// watchErrWriter returns the writer for warnings and errors emitted during a
+// watch frame. term.MakeRaw stays active for the entire cycle session, so
+// stderr output must also translate newlines or it produces the same staircase
+// as the rendered tree.
+func watchErrWriter(cycleHint bool) io.Writer {
+	return cycleOutput(cycleHint, os.Stderr)
+}
+
+func cycleOutput(cycleHint bool, w io.Writer) io.Writer {
 	if cycleHint {
-		return crlfWriter{w: os.Stdout}
+		return crlfWriter{w: w}
 	}
-	return os.Stdout
+	return w
 }
 
 // ProgressBarWidth defines the number of characters in the progress bar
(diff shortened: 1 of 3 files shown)


internal/commands/show/watch.go
```diff
@@ -14,6 +16,37 @@ import (
...
+// which clears OPOST/ONLCR so the terminal no longer maps "\n" to "\r\n" on
+// output. Without this, every rendered line starts where the previous one ended
+// (a diagonal staircase). It reports the original input length per io.Writer.
+type crlfWriter struct{ w io.Writer }
+
+func (c crlfWriter) Write(p []byte) (int, error) {
+	var b bytes.Buffer
+	b.Grow(len(p) + bytes.Count(p, []byte{'\n'}))
+	for i := 0; i < len(p); i++ {
+		if p[i] == '\n' && (i == 0 || p[i-1] != '\r') {
```
obey-agent, line 30: Minor cleanup suggestion: `crlfWriter.Write` ignores the underlying writer byte count on error and returns `0, err`. This is not the blocking issue; the requested-changes review is about raw-mode `os.Stderr` output still using bare newlines. Preserving partial-write counts would still make this wrapper stricter if reused later.

obey-agent (commented):
Superseded by the later `REQUEST_CHANGES` review after the `gpt-5.5` pass. The material issue is that raw-mode stderr output still uses bare newlines, so the PR does not fully cover terminal output generated during cycle mode.

obey-agent (changes requested):
Verdict: Request Changes

The stdout render path fix is pointed in the right direction, but the PR still leaves raw-mode terminal output paths using bare newlines on `os.Stderr`. Because `term.MakeRaw` is active for the whole cycle session, those warning/error paths can still produce the same cursor-column misalignment this PR is meant to eliminate.

Required change:
- Route raw-mode watch warnings/errors through the same newline-normalizing output path, or otherwise restore CRLF behavior for stderr while cycle mode is active. The affected paths include file watcher errors, refresh warnings, and the file-watch-unavailable fallback message in `internal/commands/show/watch.go`.

Non-blocking cleanup:
- `crlfWriter.Write` could preserve partial-write counts more strictly, but that is secondary to the stderr raw-mode issue.

The implementation should be mergeable once all user-visible terminal output produced during cycle raw mode is column-aligned, not just stdout rendering.

obey-agent (commented):
**Verdict: Approve**

Overview: This is a focused fix for the raw-mode watch rendering regression. The patch routes the festival watch render path, clear-screen output, footer output, polling fallback, and cycle-mode stderr warnings through a writer that restores CRLF behavior while raw mode has terminal output processing disabled.

Key Findings: No blocking findings. I reviewed the changed render call sites and the fallback path; the translation stays confined to cycle/raw mode and standalone watch continues to use normal stdout.

What's Done Well: The `crlfWriter` preserves existing CRLF sequences, keeps escape sequences untouched, and has tests for partial-write accounting. The new tests directly cover the behavior that prevents bare `\n` from causing the terminal staircase.

Staff Standard: Yes, I would be comfortable submitting this as my own work. Focused verification run on PR head `4556cbfdf7a46681e1526e1bb81c74d43573ef5e`: `go test ./internal/commands/show ./internal/commands/watch` passed.

PR #212, opened 2026-06-16
feat(promote): pick a festival from anywhere in a campaign

lancekrogers:
## Summary

`fest promote` was Festival-scoped: it hard-errored with "not inside a festival" unless run from inside (or linked to) a specific festival directory. You could only promote a festival while physically standing in it.

This makes promotion easy from anywhere in a campaign, mirroring how `fest go` lets you jump to any festival:

- **Scope relaxed `Festival` → `Workspace`.** The festivals root now resolves from the campaign root, a sibling festival, or a project submodule (via `workspace.FindWorkspace`, which honors `.campaign/` and `CAMP_ROOT`). `runPromote` already resolved the festival itself, so it never relied on the injected festival context.
- **Positional selector + tab completion.** `fest promote <festival>` resolves by ID, exact name, or fuzzy match (reuses `shared.ResolveFestivalSelector`); `fest promote <TAB>` completes promotable festival names.
- **Interactive picker fallback.** When no festival context is available and the terminal is interactive, `fest promote` opens the same Bubble Tea fuzzy picker `fest go` uses (`shared.PickFestivalPath`).
- **Confirmation on the picker path.** A picker selection prompts `Promote <name>: <from> → <to>? [y/N]` (default No) before the directory move and auto-commit. Running inside a festival stays immediate, so existing muscle memory is unchanged.
- **Promotable-only list.** The picker and completion show only festivals in `planning`/`ready`/`active` (festivals that have a valid lifecycle transition); ritual and dungeon are excluded.
- **JSON mode** returns a structured `{ "success": false, "error": ... }` instead of launching a TUI.

## Resolution order

1. Explicit selector argument (`fest promote <festival>`)
2. Current-directory festival context (today's behavior)
3. Interactive picker (TTY only)
4. Otherwise, an actionable `NotFound` error

## Testing

- `just build quick`, `just test unit` (2293/2293 pass), `just lint all` (0 issues).
- New unit tests cover selector resolution, cwd-context resolution, the non-interactive no-context error, completion scoping (excludes ritual), and an end-to-end `runPromote` selector promotion (planning → ready).
- Manually verified against a throwaway campaign: `fest promote <TAB>` completion, `fest promote <name>` from the campaign root, cwd-context promote, and the `--json` error path.

The interactive picker and confirmation prompt require a real TTY, so they are exercised manually rather than in unit tests; the surrounding plumbing is unit-tested.

### DIFF
--- a/internal/commands/promote/promote.go
+++ b/internal/commands/promote/promote.go
@@ -2,6 +2,7 @@
 package promote
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
@@ -42,7 +43,7 @@ type promoteOptions struct {
 func NewPromoteCommand() *cobra.Command {
 	opts := &promoteOptions{}
 	cmd := &cobra.Command{
-		Use:   "promote",
+		Use:   "promote [festival]",
 		Short: "Promote a festival to the next lifecycle status",
 		Long: `Promote moves a festival through the lifecycle: planning → ready → active → completed.
 
@@ -51,15 +52,26 @@ Each transition validates readiness:
   ready → active:      Festival is ready to begin execution
   active → completed:  All tasks must be completed
 
+By default, promotes the festival you are currently inside. From elsewhere in a
+campaign, pass a festival name or run promote interactively to pick one:
+  fest promote my-feature       Promote a festival by name (tab completion)
+  fest promote                  Pick a festival from a fuzzy picker (in a terminal)
+
 Use --dungeon to send a festival directly to a dungeon status:
   fest promote --dungeon someday     Shelve for later
   fest promote --dungeon archived    Archive the festival
   fest promote --dungeon completed   Mark as completed (skips task validation)`,
 		Annotations: map[string]string{
-			"scope": string(scope.Festival),
+			"scope": string(scope.Workspace),
 		},
+		Args:              cobra.MaximumNArgs(1),
+		ValidArgsFunction: completePromoteTarget,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runPromote(cmd.Context(), opts)
+			selector := ""
+			if len(args) > 0 {
+				selector = args[0]
+			}
+			return runPromote(cmd.Context(), opts, selector)
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


internal/commands/promote/promote.go
```diff
@@ -81,16 +167,21 @@ func runPromote(ctx context.Context, opts *promoteOptions) error {
...
-	}
-	if loc.Festival == nil {
-		return errors.NotFound("festival")
+	festivalsDir := ""
+	if ws, ok := scope.WorkspaceFrom(ctx); ok && ws != nil {
+		festivalsDir = ws.FestivalsPath
 	}
 
-	festival := loc.Festival
+	festival, fromPicker, err := resolveFestivalForPromote(ctx, festivalsDir, cwd, selector, !opts.json)
```
obey-agent, line 175: This new resolution path can return a festival whose path is unrelated to the process cwd, but the later chain dependency gate still resolves the festivals root from `os.Getwd()` inside `checkChainDependencies`. When cwd is the campaign root or a project submodule, `tpl.FindFestivalsRoot(cwd)` fails and the helper silently returns no block, so selector-based `ready -> active` promotions can bypass hard upstream dependencies.

I reproduced this with a ready festival `FA0010` chained behind an active upstream `FA0009`: running `fest promote onboarding-parity-FA0010 --no-commit` from the campaign root promoted it to `active`; running `fest promote --no-commit` from inside the festival correctly blocked. Please anchor the gate on `festival.Path` or the resolved workspace festivals root, and add a regression test for selector promotion from outside `festivals/`.

obey-agent (commented):
Verdict: Request Changes

Overview: This is a well-scoped improvement to make `fest promote` usable from the campaign root and project submodules, and it reuses the existing selector and picker plumbing instead of creating a parallel implementation. The command wiring, completion behavior, JSON non-interactive path, and picker confirmation are directionally sound.

Key Findings:
- Hard chain dependencies can be bypassed when promoting `ready -> active` by selector from outside `festivals/`. I left an inline comment on `internal/commands/promote/promote.go`: `checkChainDependencies` still anchors on process cwd, so the new out-of-festival resolution path can silently skip the gate.

What's Done Well:
- The PR keeps the resolution order clear: explicit selector, current festival, picker fallback, then actionable error.
- It reuses `shared.ResolveFestivalSelector`, `CompleteFestivalPickSelectors`, and `PickFestivalPath`, which keeps behavior aligned with existing navigation surfaces.
- The new tests cover selector resolution, cwd fallback, completion scoping, and non-interactive failure behavior.

Verification:
- `go test ./internal/commands/promote` passed.
- `just build quick` passed.
- `just test unit` passed: 90 packages, 2293/2293 tests.
- `just lint all` passed: 0 issues.
- Manual repro confirmed the regression: a ready festival chained behind an active upstream promoted from the campaign root with `fest promote <selector> --no-commit`, while the same setup correctly blocked when run from inside the festival.

Staff Standard: I would be proud of the feature shape and reuse, but I would not merge it until the dependency gate is anchored on the selected festival/workspace rather than cwd and covered by a regression test.

lancekrogers:
Addressed in c6c7a90.

The chain dependency gate now anchors the festivals root on the festival being promoted (`tpl.FindFestivalsRoot(festival.Path)`) instead of `os.Getwd()`, so the hard-upstream check holds whether `promote` is invoked from inside the festival, the campaign root, or a project submodule. `AutoCommitStatusChange` already anchors on the workspace from context (injected by the now-Workspace scope), so it was unaffected; the chain gate was the only cwd-anchored correctness path.

Regression coverage:
- `TestCheckChainDependencies_AnchorsOnFestivalNotCwd` — gate unit boundary with cwd moved outside `festivals/`.
- `TestRunPromote_SelectorReadyToActiveBlockedByChainFromCampaignRoot` — end-to-end selector promotion from the campaign root.

I verified both tests fail on the old cwd-anchored code and pass with the fix. Re-ran the gate at this head: `just build quick`, `just test unit` (2295/2295), `just lint all` (0 issues). Manually reproduced your scenario: promoting `FA0010` (ready -> active, hard upstream `FA0009` active) from the campaign root now prints the chain dependency gate block and leaves `FA0010` in `ready`.

obey-agent (commented):
**Verdict: Approve**

Re-reviewed current head `c6c7a90b1232ef2d4e9f5cd7947db4f973ad5343`. The prior blocker is fixed: the chain dependency gate now anchors on the selected festival path instead of process cwd, so selector-based `ready -> active` promotion from the campaign root no longer bypasses hard upstream dependencies.

**Key Findings**

- No remaining blocking findings. The new regression coverage includes both the unit boundary (`checkChainDependencies` from outside `festivals/`) and an end-to-end selector promotion from the campaign root that must remain blocked while the hard upstream is incomplete.

**What's Done Well**

- The PR keeps the scope expansion small and reuses existing shared selector/picker behavior instead of adding a parallel resolver.
- JSON/non-interactive behavior remains explicit, and picker confirmation only applies to the picker path, preserving current in-festival behavior.
- The chain gate fix is anchored at the right domain object: the festival being promoted.

**Verification**

- `go test ./internal/commands/promote` passed.
- `go test ./internal/scope ./internal/commands/shared ./internal/commands/promote` passed.
- `just build quick` passed.
- `just test unit` passed: 90 packages, 2295/2295 tests.
- `just lint all` passed: 0 issues.

**Staff Standard**

Yes. I would be comfortable submitting this as my own work: the original correctness gap is closed, covered by targeted regressions, and the command behavior stays aligned with existing fest navigation surfaces.

PR #213, opened 2026-06-17
fix(watch): open the festival detected from cwd, not the first in the cycle

lancekrogers:
## Summary

`fest watch`, run from inside a festival, opened the wrong festival. When the workspace had more than one watchable festival, `runWatch` built the watch cycle and always started it at index 0 (the newest active festival, then ready, then planning by `OrderByStatusThenRecency`), ignoring the festival detected from the current directory.

Now `fest watch` opens the festival in context:

- `runWatch` detects the current festival from cwd via `deps.detectFestival` (`show.DetectCurrentFestival`, which walks up for festival markers and also follows project links) and starts the cycle at that festival's index in the cycle list.
- Matching is by canonical, symlink-resolved path (`filepath.EvalSymlinks` with `Abs`/`Clean` fallback), so the detected festival root and the cycle entry resolve to the same on-disk directory regardless of how each path was constructed.
- Falls back to index 0 when cwd is not inside a watchable festival, preserving today's behavior. The cycle navigation (arrow keys to browse other festivals) is unchanged; only the starting position is fixed.

Root cause was a single line in `runWatch` (`internal/commands/watch/command.go`): `runWatchCycle(ctx, paths, 0, ...)`.

## Testing

- `just build quick`, `just test unit` (2297/2297 pass), `just lint all` (0 issues).
- New unit tests:
  - `TestCurrentFestivalIndex` (table): detected festival selects its index; not-in-a-festival, detected-festival-outside-the-cycle, and missing-detector all fall back to 0.
  - `TestWatchCommandCycleStartsAtCurrentFestival`: end-to-end through `runWatch` (non-TTY), asserting the watch delegate receives the current festival, not the first in the cycle. Verified this test fails on the old `startIndex=0` code and passes with the fix.
- Manual repro with the built binary against a temp campaign: two active festivals where `older-feature-FS0001` is index 1 by recency. Running `fest watch` from inside `older-feature-FS0001` now opens `older-feature-FS0001` (confirmed in rendered output), where before it opened the newest festival.

### DIFF
--- a/internal/commands/watch/command.go
+++ b/internal/commands/watch/command.go
@@ -114,7 +114,7 @@ func runWatch(ctx context.Context, args []string, opts *options, deps commandDep
 		if err == nil {
 			paths, listErr := deps.listCycleTargets(ctx, cwd)
 			if listErr == nil && len(paths) > 1 {
-				return runWatchCycle(ctx, paths, 0, *opts, deps)
+				return runWatchCycle(ctx, paths, currentFestivalIndex(ctx, cwd, paths, deps), *opts, deps)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
Verdict: Approve

Overview: This is a tightly scoped fix for `fest watch` cycle startup. The implementation keeps the existing cycle behavior intact, but chooses the initial cycle index from the festival detected from the current working directory, with symlink-aware canonical path comparison and safe fallback to index 0 when context cannot be resolved.

Key Findings: No blocking or follow-up issues found. I did not leave inline comments because there were no correctness, maintainability, or production-readiness concerns that met the bar.

What's Done Well: The change extends the existing watch resolver/cycle path instead of introducing a parallel implementation. The tests cover direct index selection, fallback behavior, missing detector behavior, and an end-to-end `runWatch` delegate assertion that catches the original regression.

Validation:
- `go test ./internal/commands/watch`
- `go test -run 'Test(CurrentFestivalIndex|WatchCommandCycleStartsAtCurrentFestival|RunWatchCycle)' ./internal/commands/watch`
- `just build quick`
- `just test unit` (2297/2297 passed)
- `just lint all`
- `git diff --check refs/remotes/origin/main...HEAD`

Staff Standard: Yes, I would be comfortable submitting this as my own work. It is small, integrated with the existing abstractions, and covered by focused regression tests plus the repo gates.

PR #214, opened 2026-06-17
feat(picker): show festival progress bars in the interactive picker

lancekrogers:
## Summary

The shared interactive festival picker now shows a progress bar and percentage after each festival, so you can see status while selecting. The picker is shared, so this lands in every festival picker: `fest go`, `fest promote`, `fest watch`, `fest status`, and `fest progress`.

```
─ 4/4 ──────────────────────────────────────────────────
> Type to filter...
▶ [active] daemon-chat-DC0001    ██████████ 100%
  [ready] launch-prep-LP0003     █████░░░░░ 50%
  [planning] my-feature-FE0007   ░░░░░░░░░░ 0%
  [active]/
```

## How it works

- `picker.Item` gains a display-only `Detail` field, rendered in an aligned trailing column. It is **not** used for fuzzy matching, so typing still filters on the festival name only. When no item has a `Detail`, rendering is byte-identical to before, so non-festival pickers are unaffected.
- `PickFestivalPath` fills each item's `Detail` with a compact bar via `progress.NewManagerReadOnly` + `GetFestivalProgress` (reusing `ui.RenderProgressBar`, the same helper `fest list` uses). This is the same orientation-only read pattern `fest stats` already uses.
- Items with no tasks (status directories like `[active]/`, empty festivals) report `Total == 0` and get no bar, so the column stays clean without needing per-item type flags.
- Progress is computed **concurrently** (bounded worker pool) so picker-open stays snappy even in large workspaces, and only after the TTY check, so non-interactive callers (scripts, agents) pay nothing.

Read-only note: for modern festivals this performs no disk writes. The only possible write is the same one-time legacy time-data backfill that `fest list`/`fest stats` already trigger via `GetFestivalProgress`.

## Testing

- `just build quick`, `just test unit` (2301/2301 pass), `just lint all` (0 issues). The new concurrent path is clean under `go test -race`.
- New unit tests:
  - `shared`: `TestFestivalProgressDetail` (real 4-task/2-complete fixture marked via the progress Manager → bar contains `50%`; empty festival → no bar) and `TestAttachProgressDetails` (festival item gets a bar, path-less item does not).
  - `picker`: `TestPadRight` and `TestViewRendersDetailColumn` (View renders the Detail column and the names).
- Rendered the real composed frame (shown above) through the actual `attachProgressDetails` + `picker.View` code paths against real progress fixtures.

### DIFF
--- a/internal/commands/shared/festival_picker.go
+++ b/internal/commands/shared/festival_picker.go
@@ -4,12 +4,17 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/navigation"
+	"github.com/Obedience-Corp/fest/internal/progress"
 	"github.com/Obedience-Corp/fest/internal/tui/picker"
+	"github.com/Obedience-Corp/fest/internal/ui"
 	"golang.org/x/term"
 )
+
+const progressDetailWorkers = 8
 
 // PickFestivalPath opens the shared festival picker and returns the selected path.
 func PickFestivalPath(ctx context.Context, festivalsDir string, opts FestivalPickerOptions) (string, error) {
@@ -28,6 +33,7 @@ func PickFestivalPath(ctx context.Context, festivalsDir string, opts FestivalPic
 	if len(items) == 0 {
 		return "", nil
 	}
+	attachProgressDetails(ctx, items)
 
 	selected, err := picker.Run(items, navigation.Score)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


obey-agent (commented):
Verdict: Approve

Overview: This is a well-scoped enhancement to the shared festival picker. The implementation keeps `Detail` display-only, preserves name-only fuzzy matching, and reuses the existing read-only progress manager plus existing UI progress bar renderer instead of adding a parallel progress path.

Key Findings: No blocking or follow-up issues found. I did not leave inline comments because there were no correctness, maintainability, or production-readiness concerns that met the bar.

What's Done Well: The change is integrated at the shared picker boundary, so `fest go`, `fest promote`, `fest watch`, `fest status`, and `fest progress` benefit consistently. The bounded concurrent progress lookup path is covered by targeted tests and was clean under the race detector. The PTY smoke rendered the real picker frame with aligned progress details, and status-directory rows remained visually clean.

Validation:
- `go test ./internal/commands/shared ./internal/tui/picker`
- `go test -race ./internal/commands/shared ./internal/tui/picker`
- `just build quick`
- `just test unit` (2301/2301 passed)
- `just lint all`
- `git diff --check refs/remotes/origin/main...HEAD`
- PTY smoke: ran `/tmp/fest-pr-214-review/bin/fest go` from the campaign root and verified the interactive picker rendered progress bars/percentages in the detail column before cancelling.

Staff Standard: Yes, I would be comfortable submitting this as my own work. It is cohesive, uses existing project primitives, and has the right regression coverage for the UI and concurrent detail population paths.

PR #215, opened 2026-06-17
fix(sync): dev channel resolves to the latest tag, never behind stable

lancekrogers:
## Summary

`fest system sync` on a dev build was syncing months-old templates. The dev channel only ever matched `-dev` tags (`internal/github/tags.go`), so once dev tagging lapsed (the last dev tag, `v0.1.0-dev.2`, is from April) it silently froze there while stable advanced to `v0.4.x`. Dev and festival-app users got stale templates with no signal that anything was wrong.

This was a release-process gap, not a parse bug, but the resolver had no fallback, so it stranded dev users by design. This makes the dev channel self-healing.

## Change

- **Dev channel now resolves to the newest tag overall.** Its candidate set is dev pre-releases **plus** stable releases, so it can never fall behind stable. A newer dev pre-release still wins when one exists; once a stable release of the same base ships, dev follows it.
- **Comparator fix.** `semverGreater` now ranks a release above its own pre-release (`v1.0.0 > v1.0.0-dev.5`), which is required now that dev and stable tags are compared together. This is backward-compatible for the stable-only and dev-only paths.
- **Removed a parallel implementation.** The test file previously reimplemented the resolver (`resolveFromOutput`) and had already diverged in spirit. Extracted pure `selectLatestTag` and `parseLsRemoteTags` helpers so production and tests share one code path.

`ResolveLatestTag` has a single production caller (`fest system sync`), so the behavior change is scoped to template sync; nothing else (e.g. self-update) depends on it.

Note: this is the durable fix. Cutting fresh dev tags is still fine and will make dev lead again, but it's no longer required to keep dev users current.

## Testing

- `just build quick`, `just test unit` (2301/2301 pass), `just lint all` (0 issues).
- New/updated unit tests in `internal/github/tags_test.go`:
  - `TestSelectLatestTag_DevNeverBehindStable`: dev falls back to latest stable when dev tags are stale (the exact bug: stale `v0.1.0-dev.2` + stable up to `v0.4.3` → `v0.4.3`); a newer dev pre-release leads stable; a release outranks its own pre-releases; dev-only sets still pick the highest dev.
  - `TestSemverGreater`: added release-vs-pre-release and cross-base cases.
  - `TestParseLsRemoteTags`: tag extraction and annotated-pointer (`^{}`) filtering.
- Verified against the real repo with the built binary: `fest system sync --channel dev --dry-run` now prints `Resolved dev channel to tag v0.4.4` (was `v0.1.0-dev.2`), matching the stable channel.

### DIFF
--- a/internal/github/tags.go
+++ b/internal/github/tags.go
@@ -51,11 +51,14 @@ func semverGreater(a, b semver) bool {
 	if a.patch != b.patch {
 		return a.patch > b.patch
 	}
-	// For pre-release comparison (dev channel), compare numerically by pre suffix.
-	// e.g. "dev.10" > "dev.9"
-	aNum := preReleaseNum(a.pre)
-	bNum := preReleaseNum(b.pre)
-	return aNum > bNum
+	// Same base version: a release (no pre-release) outranks any pre-release of
+	// that version, e.g. v1.0.0 > v1.0.0-dev.5.
+	if (a.pre == "") != (b.pre == "") {
+		return a.pre == ""
+	}
+	// Two pre-releases of the same base: compare numerically by suffix, e.g.
+	// "dev.10" > "dev.9".
+	return preReleaseNum(a.pre) > preReleaseNum(b.pre)
 }
 
 // preReleaseNum extracts the numeric part from a pre-release string like "dev.3".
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
**Verdict: Approve**

Reviewed current head `6a34fda91a352a77cede57908bcdd8e09a0a7d8a`. This is a focused fix for the dev-channel sync freeze: the production resolver now shares pure parsing/selection helpers with the tests, and the dev channel compares dev prereleases against stable releases so it no longer strands users on stale `-dev` tags.

**Key Findings**

- No blocking findings. The comparator now correctly ranks a stable release above its own prerelease, and the dev channel behavior is covered for stale dev tags, newer dev prereleases, stable-over-same-base prereleases, and dev-only tag sets.

**What's Done Well**

- The duplicate test-only resolver was removed, which reduces the chance of future test drift.
- `parseLsRemoteTags` keeps annotated tag pointer filtering explicit and independently tested.
- The production caller remains scoped to `fest system sync` channel resolution, so this does not broaden into unrelated update paths.

**Verification**

- `go test ./internal/github` passed.
- `gofmt -l internal/github/tags.go internal/github/tags_test.go` was clean.
- `just build quick` passed.
- `just test unit` passed: 90 packages, 2301/2301 tests.
- `just lint all` passed: 0 issues.

**Staff Standard**

Yes. I would be comfortable submitting this: the behavior is intentional, the comparison semantics are now correct for mixed stable/dev candidate sets, and the tests exercise the release-process failure mode directly.

PR #217, opened 2026-06-17
feat(workflow): add 'fest workflow create' alias for 'fest create workflow'

lancekrogers:
## Problem

Creating a standalone `WORKFLOW.md` outside a festival is supported via `fest create workflow <name>` (it writes the doc, initializes `.workflow/`, and starts a tracked run so `fest next` works immediately). But the entrypoint is hard to discover: agents and humans look under the `workflow` noun first, and `fest workflow ...` only exposes `init`, `start`, `show`, `advance`, etc. Those operate on a WORKFLOW.md that already exists (`fest workflow init` literally requires one). There was no command under `fest workflow` to scaffold a new one, so the scaffolding command was effectively invisible from that namespace.

This actually tripped up an agent session: asked to "create a WORKFLOW.md with the fest CLI," it went down the `fest workflow init`/`start` path and stalled instead of finding `fest create workflow`.

## Change

Add `fest workflow create` as a discoverability alias for `fest create workflow`:

- It delegates to the same `festival.RunCreateWorkflow`, so both surfaces behave identically.
- The flag set is shared via a new exported `BindCreateWorkflowFlags`, so the two commands cannot drift.
- The inherited `--phase` persistent flag (from the parent `workflow` command, used by the navigation subcommands) is rejected explicitly rather than silently ignored. This matches the existing `rejectFestivalOnlyFlags` contract, which the codebase added specifically to avoid confusing silent flag drops.
- Help text now points to the scaffolding command from both the `fest workflow` group description and `fest workflow init`.

No behavior change to `fest create workflow` itself; the help text was extracted to a shared const but is byte-identical.

## Testing

- Unit (`internal/commands/workflow/create_test.go`): alias is registered, flag parity with `create workflow`, `--phase` is rejected, and an end-to-end standalone scaffold produces `WORKFLOW.md` + `.workflow/workflow.yaml`.
- Integration (`tests/integration/create_workflow_standalone_test.go`): `fest workflow create` produces the same artifacts as `fest create workflow` and `fest next` works immediately after; `--phase` is rejected.
- Gates: `just lint all` (0 issues), `just lint vet`, `go vet -tags integration`, and `just test unit` (2305/2305 passing).

## Notes for reviewers

The alias is intentionally an alias, not a relocation: `fest create workflow` stays canonical alongside its `create festival/phase/sequence/task` siblings, and `fest workflow create` is the discoverability shortcut under the noun agents reach for first.

This is the CLI half of a broader follow-up captured as a campaign intent (skill docs for standalone WORKFLOW.md creation, then propagation to the festival plugin skill copies for claude/codex/opencode/cursor). Those docs intentionally follow once this command lands, so they describe shipped behavior.

### DIFF
--- a/internal/commands/festival/create_workflow.go
+++ b/internal/commands/festival/create_workflow.go
@@ -56,13 +56,9 @@ type createWorkflowResult struct {
 	Suggestions []string         `json:"suggestions,omitempty"`
 }
 
-// NewCreateWorkflowCommand creates the 'create workflow' subcommand.
-func NewCreateWorkflowCommand() *cobra.Command {
-	opts := &CreateWorkflowOptions{}
-	cmd := &cobra.Command{
-		Use:   "workflow",
-		Short: "Create a standalone or phase WORKFLOW.md from structured step definitions",
-		Long: `Generate a parseable WORKFLOW.md from prompts, inline JSON, or a steps file.
+// createWorkflowLong is the shared help text for `fest create workflow` and its
+// `fest workflow create` alias so the two never drift.
+const createWorkflowLong = `Generate a parseable WORKFLOW.md from prompts, inline JSON, or a steps file.
 
 Outside a festival phase, this creates WORKFLOW.md in the current directory,
 initializes .workflow/ runtime state, and starts a tracked run so fest next works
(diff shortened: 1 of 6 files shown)


obey-agent (commented):
**Verdict:** Approve

**Overview:** This is a tight discoverability fix. The new `fest workflow create` surface delegates to the existing `festival.RunCreateWorkflow` implementation and shares the flag binding, which keeps behavior aligned with `fest create workflow` instead of creating a second implementation.

**Key Findings:** None blocking.

**What's Done Well:** The implementation handles the inherited `--phase` flag explicitly, keeps command scope global so standalone creation is not blocked by festival resolution, and adds both unit and integration coverage for registration, flag parity, scaffolding, and `fest next` after create.

**Staff Standard:** Yes. I would be comfortable submitting this as my own work.

Verification: `go test ./internal/commands/workflow ./internal/commands/festival`

PR #221, opened 2026-06-18
Extend fest walk to standalone workflows

obey-agent:
## Summary

- Extend `fest walk` / `fest inspect` to resolve standalone `WORKFLOW.md` contexts in addition to festivals.
- Add workflow-shaped JSON output and text orientation for standalone workflows, including mode, run status, progress, current step, blocked state, and stale-document warnings.
- Reuse the existing standalone workflow resolver/display model so festival context still wins when both signals are present.
- Add walk command tests for anonymous standalone workflow JSON and text output.

## Validation

- `go test ./internal/commands/walk`
- `just test unit`
- Manual: `go run ./cmd/fest walk --json <tmp-workflow-dir>`

### DIFF
--- a/internal/commands/walk/walk.go
+++ b/internal/commands/walk/walk.go
@@ -28,16 +28,17 @@ type walkOptions struct {
 
 // WalkView is the structured view returned by fest walk.
 type WalkView struct {
-	Kind     string `json:"kind"`
-	Name     string `json:"name"`
-	Status   string `json:"status"`
-	Path     string `json:"path"`
-	Goal     string `json:"goal,omitempty"`
+	Kind     string        `json:"kind"`
+	Name     string        `json:"name"`
+	Status   string        `json:"status"`
+	Path     string        `json:"path"`
+	Goal     string        `json:"goal,omitempty"`
 	Progress *walkProgress `json:"progress,omitempty"`
-	Next     string `json:"next,omitempty"`
+	Next     string        `json:"next,omitempty"`
 	Blocked  []walkBlocker `json:"blocked,omitempty"`
-	Gates    []string `json:"gates,omitempty"`
-	Warnings []string `json:"warnings,omitempty"`
+	Gates    []string      `json:"gates,omitempty"`
+	Warnings []string      `json:"warnings,omitempty"`
+	Workflow *walkWorkflow `json:"workflow,omitempty"`
 }
 
 type walkProgress struct {
@@ -47,8 +48,31 @@ type walkProgress struct {
 }
 
 type walkBlocker struct {
-	Task    string `json:"task"`
-	Reason  string `json:"reason"`
+	Task   string `json:"task"`
+	Reason string `json:"reason"`
+}
+
+type walkWorkflow struct {
+	Mode           string             `json:"mode"`
+	WorkflowDoc    string             `json:"workflow_doc"`
+	RuntimeDir     string             `json:"runtime_dir,omitempty"`
+	RunID          string             `json:"run_id,omitempty"`
+	RunStatus      string             `json:"run_status"`
+	CurrentStep    int                `json:"current_step"`
+	TotalSteps     int                `json:"total_steps"`
+	CompletedSteps int                `json:"completed_steps"`
+	Blocked        bool               `json:"blocked,omitempty"`
+	DocHashChanged bool               `json:"doc_hash_changed,omitempty"`
+	Steps          []walkWorkflowStep `json:"steps"`
+}
+
+type walkWorkflowStep struct {
+	Number        int    `json:"number"`
+	Name          string `json:"name"`
+	Status        string `json:"status"`
+	Current       bool   `json:"current,omitempty"`
+	HasCheckpoint bool   `json:"has_checkpoint,omitempty"`
+	Goal          string `json:"goal,omitempty"`
 }
 
 // NewWalkCommand creates the walk/inspect command.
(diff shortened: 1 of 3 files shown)


internal/commands/walk/walk.go
```diff
@@ -47,8 +48,31 @@ type walkProgress struct {
 }
 
 type walkBlocker struct {
-	Task    string `json:"task"`
-	Reason  string `json:"reason"`
+	Task   string `json:"task"`
+	Reason string `json:"reason"`
+}
+
+type walkWorkflow struct {
```
obey-agent: **Parallel implementation of an existing JSON contract.** `walkWorkflow` (and `walkWorkflowStep` just below) are field-for-field, tag-for-tag identical to the unexported `standaloneWorkflowJSON` / `standaloneWorkflowStep` in `internal/commands/show/standalone.go`, and the mapping loop in `buildStandaloneWalkView` re-derives the same `StandaloneWorkflowInfo` -> JSON projection that `emitStandaloneWorkflowJSON` already performs.

I verified the two emit byte-identical output: `fest walk --json <dir>`'s `.workflow` object equals the entire body of `fest show --json` for the same WORKFLOW.md.

Two independent definitions of the same wire contract will drift. fest is in production and festival-app deserializes these shapes, so a field add/rename/tag change on one side silently breaks the consumer reading the other. The staff standard here is "extend existing systems, no parallel implementations."

Suggested fix: export the view type + builder from the `show` package (e.g. `show.StandaloneWorkflowJSON` plus `show.NewStandaloneWorkflowJSON(info)`), make `WalkView.Workflow` reference that exported type, and have both `emitStandaloneWorkflowJSON` and `buildStandaloneWalkView` construct through the one builder. A single source of truth keeps the two surfaces in lockstep.

internal/commands/walk/walk.go
```diff
@@ -177,6 +217,71 @@ func resolveFestival(ctx context.Context, startDir, campaignRoot string) (*show.
...
+			Number:        step.Number,
+			Name:          step.Name,
+			Status:        string(step.Status),
+			Current:       step.IsCurrent,
+			HasCheckpoint: step.HasCheckpoint,
+			Goal:          step.Goal,
+		})
+	}
+
+	if info.Blocked && view.Next != "" {
```
obey-agent, line 218: **The interesting branches are untested.** This blocked -> `walkBlocker` translation and the `DocHashChanged` -> warning branch just below are the most behaviorally meaningful parts of the standalone view, but both new tests (`TestRunWalk_StandaloneWorkflowJSONOutput` / `...TextOutput`) exercise only anonymous mode: no run, never blocked, never stale. These two branches have zero coverage.

Please add a tracked-mode test (`.workflow/workflow.yaml` plus an active run state with `Blocked: true` and `DocHashChanged: true`) asserting the blocker is surfaced in `view.Blocked` and the "WORKFLOW.md has changed since this run started" warning appears. That is where a regression would actually bite.

obey-agent (commented):
## Staff Review: Request Changes

**Verdict: Request Changes**

**Overview:** A clean, well-scoped extension that teaches `fest walk` / `fest inspect` to orient inside a standalone `WORKFLOW.md`, reusing the existing `show.ResolveStandaloneWorkflow` resolver so festival context still wins. Behavior is correct, the standalone-first ordering matches the existing `fest show` flow, and the gate is green at head (gofmt clean, build, vet, and `go test ./internal/commands/walk` all pass; manual `walk --json` verified). One maintainability issue blocks merge.

**Key findings:**
- (blocker) Parallel JSON model. `walkWorkflow` / `walkWorkflowStep` are byte-for-byte identical to the unexported `standaloneWorkflowJSON` / `standaloneWorkflowStep` in `internal/commands/show/standalone.go`, and the projection loop in `buildStandaloneWalkView` duplicates `emitStandaloneWorkflowJSON`. I confirmed `fest walk --json`'s `.workflow` object is byte-identical to `fest show --json`'s body. Two definitions of one wire contract will drift, and festival-app deserializes both. Export a shared view + builder from the `show` package and have both call sites use it. (inline at walk.go:55)
- (should fix) Test gap on the meaningful branches. Both new tests cover anonymous mode only; the blocked-run -> `walkBlocker` translation and the stale-document (`DocHashChanged`) warning branch have zero coverage. Add a tracked-mode test. (inline at walk.go:260)

**What's done well:**
- Correct precedence reuse: `ResolveStandaloneWorkflow` returns nil for festival contexts, so festival handling is preserved and the ordering matches `fest show` (show.go:197).
- Proper context propagation and project error package usage throughout (`errors.Wrap`, ctx threaded into every resolver call, `ctx.Err()` checked up front).
- Incidentally gofmt-aligned the pre-existing `WalkView` struct tags.
- Walk's standalone text output stays in the command's own H1/H2 house style instead of copying show's renderer, which is the right call for a distinct surface.

**Staff standard:** Would I submit this as my best work? Not yet. The logic is sound and the tests that exist are good, but duplicating a JSON contract that an external consumer already deserializes is exactly the "parallel implementation" the bar warns against. Collapse the two workflow-JSON definitions into one exported source of truth and cover the blocked/stale branches, and this becomes a confident approve.

lancekrogers:
Addressed both review findings in 3719213:

- **Parallel JSON model:** exported `show.StandaloneWorkflowJSON` + `show.NewStandaloneWorkflowJSON(info)` as the single source of truth for the standalone WORKFLOW.md wire shape, and pointed `WalkView.Workflow` at it. Removed the duplicate `walkWorkflow` / `walkWorkflowStep` types and the duplicate projection loop. Verified `fest walk --json`'s `.workflow` object is still byte-identical to `fest show --json`, now from one definition.
- **Untested branches:** added `TestBuildStandaloneWalkView_BlockedAndStale`, covering the blocked-run -> `walkBlocker` translation and the `DocHashChanged` -> warning branch.

Gate at the new head: gofmt clean, `go build ./...`, `go vet`, `golangci-lint` (0 issues), and `go test ./internal/commands/walk ./internal/commands/show` all green.

obey-agent (commented):
**Verdict: Approve**

**Overview:** I re-reviewed the updated head (`3719213`) for PR #221. The previous blocker is addressed: `fest walk --json` now embeds the shared `show.StandaloneWorkflowJSON` projection instead of maintaining a duplicate wire model, and the blocked/stale workflow projection has focused test coverage.

**Key Findings:**
- No blocking findings in the updated diff.
- The standalone workflow path still preserves festival precedence by resolving through `show.ResolveStandaloneWorkflow`, which returns nil for festival contexts.
- The JSON contract is now centralized behind `show.NewStandaloneWorkflowJSON`, so `show --json` and `walk --json.workflow` have one source of truth.

**What's Done Well:**
- Clean reuse of the existing standalone resolver/display model.
- Good command behavior for both text and JSON standalone workflow output.
- Coverage now includes anonymous workflow output plus blocked/stale projection.
- Error wrapping and context propagation remain consistent with the repo style.

**Validation:**
- `go test ./internal/commands/walk ./internal/commands/show ./internal/workflow/standalone`
- `just test unit` - 90 packages, 2311/2311 tests passed
- `just build stable` - build/vet across 107 packages passed
- Manual probe: built a local `fest` binary and verified `walk --json <standalone-workflow-dir>` includes a `workflow` object matching the standalone `show --json` body.

**Staff Standard:** Yes, I would be comfortable merging this version.

PR #224, opened 2026-06-22
feat(watch): show festival status and add p promote shortcut

lancekrogers:
## Summary

Two changes to `fest watch`, both scoped to the festival watch view:

1. **Show lifecycle status.** The default tree watch view now renders the festival's lifecycle status (`planning` / `ready` / `active`) above the progress bar. Previously only `--summary` showed it; the tree view did not.
2. **Promote from watch with `p`.** In an interactive watch session, pressing `p` promotes the watched festival through its next lifecycle transition and refreshes in place at the festival's new on-disk location.

## Design

`workflow/design/fest-watch-status-and-promote-2026-06-22/` (design workitem `WI-831ebc`).

## Key decision: reuse the `fest promote` flow

This is a TUI/UX change, **not** a new promotion implementation. The `p` key triggers the same flow `fest promote` runs. The transition table, readiness + chain-dependency gates, directory move, frontmatter / `fest.yaml` / nav-link updates, and auto-commit all stay in `internal/commands/promote` and are reused via a new `PromoteResolved(ctx, festival) (newPath, err)` seam:

- `runPromote`'s post-resolution body was extracted into `promoteCore`; `runPromote` and `PromoteResolved` both call it. `fest promote` behavior (including `--json`, `--force`, `--dungeon`, picker confirm) is unchanged.
- The watch loop leaves raw mode before promoting so the shared confirmation and output behave exactly as on the CLI, then re-enters raw mode for the next frame. No new result/error taxonomy, no separate commit policy.

## Behavior

- **Every promote confirms** (`y/N`); `p` sits next to the `←`/`→` cycle keys and the action mutates on-disk state.
- A held readiness or chain gate prints the same message as `fest promote` and stays in watch. Forcing remains a deliberate `fest promote --force`, never a keystroke.
- After a successful promote the directory moves; the loop re-resolves to the new path and keeps watching.
- Single-festival watch now enters the raw-mode loop in a TTY so the shortcut works without cycling. The footer shows `p promote`, and `← → cycle` only when there is more than one target.
- Non-TTY / piped watch and standalone `WORKFLOW.md` watch are unchanged (no raw mode, no promote).

## Tests

- `promote`: `PromoteResolved` confirmed (moves planning→ready, shares the CLI flow) and declined (no move); existing `runPromote` tests unchanged.
- `watch`: `classifyCycleKey` maps `p`/`P` to the promote action; `cycleWatchOptions` propagates the cycling flag.
- `show`: tree watch view renders the lifecycle status line; footer advertises `p promote` (and `← → cycle` only with multiple targets).
- Gates green at HEAD: `just build quick-stable` + `quick-dev`, `just test unit` (2317 passed), `just lint all` (0 issues), `go vet`, `gofmt`.

## Note on coverage

The interactive raw-mode keystroke path (`promoteFromWatch` + the TTY loop wiring) is covered at the component level (key classification, footer, `PromoteResolved` core) plus manual acceptance; a full PTY-driven integration test in the container harness is a reasonable follow-up.

### DIFF
--- a/internal/commands/promote/promote.go
+++ b/internal/commands/promote/promote.go
@@ -182,6 +182,18 @@ func runPromote(ctx context.Context, opts *promoteOptions, selector string) erro
 		}
 		return err
 	}
+
+	_, err = promoteCore(ctx, festival, fromPicker, opts)
+	return err
+}
+
+// PromoteResolved promotes an already-resolved festival through the fest promote
+// flow, asking for confirmation, and returns the post-move path (empty if nothing moved).
+func PromoteResolved(ctx context.Context, festival *show.FestivalInfo) (string, error) {
+	return promoteCore(ctx, festival, true, &promoteOptions{})
+}
+
+func promoteCore(ctx context.Context, festival *show.FestivalInfo, confirm bool, opts *promoteOptions) (string, error) {
 	currentStatus := festival.Status
 
 	var nextStatus string
@@ -190,7 +202,7 @@ func runPromote(ctx context.Context, opts *promoteOptions, selector string) erro
 		// Dungeon override: resolve the dungeon status
 		resolved := id.ResolveStatusPath(opts.dungeon)
 		if !strings.HasPrefix(resolved, "dungeon/") {
-			return errors.Validation("invalid dungeon status").
+			return "", errors.Validation("invalid dungeon status").
 				WithField("value", opts.dungeon).
 				WithField("hint", "valid values: completed, archived, someday")
 		}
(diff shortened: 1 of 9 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** This PR keeps the promotion behavior centralized by extracting `promoteCore` and having the watch shortcut call the same flow as `fest promote`, which is the right integration shape for a mutating lifecycle action. The watch UI changes are narrowly scoped: lifecycle status appears in the tree view, raw-mode hints distinguish single-target promote from multi-target cycling, and non-TTY/standalone watch paths stay on the existing plain watch behavior.

**Key Findings:** No blocking findings. I did not find a correctness, error-handling, or maintainability issue that warrants an inline comment.

**What's Done Well:** The PR avoids a parallel promotion implementation, preserves the existing readiness/chain/commit policy, and adds focused tests around the new reuse seam, footer behavior, status rendering, and key classification. The comments around raw terminal mode and CRLF handling are useful because that code is easy to break accidentally.

**Verification:** I reviewed every changed file at head `26cb09fd82b4255193e1baf3c8e211c3eb73a316` and ran:

- `go test ./internal/commands/promote ./internal/commands/show ./internal/commands/watch`
- `just test unit` (90 packages, 2317/2317 tests passed)
- `just build quick-stable`
- `just lint vet`

**Staff Standard:** Yes, I would be comfortable submitting this as my own work. The only residual risk is the intentionally manual/TTY portion of the raw-mode promote loop, but the component coverage and reuse of the existing promotion path are appropriate for this PR.

PR #225, opened 2026-06-25
feat(extensions): load installer-managed marketplace extension path (feature-gated)

lancekrogers:
Part of festival **OI0006** (obey-installer extensions support). Adds fest-side support for loading extensions that **obey-installer** activates into `~/.obey/fest/marketplace/extensions/`, behind the `marketplace_extension_source_v1` capability.

## What

- **`internal/features`** (new leaf package) — fest's capability set: the `marketplace_extension_source_v1` constant, `Supported()` (what `version --json` advertises so the installer's host-feature contract can read it), and `Enabled()`.
- **`internal/extensions/loader.go`** — `MarketplaceExtensionsDir()` (= `~/.obey/fest/marketplace/extensions`, mirroring `config.ConfigDir()`), plus a feature-gated load inserted **above built-in and below user/project**. Because `loadFromDirectory` is last-wins, a `marketplace` extension overrides `built-in` and is overridden by `user`/`project` — matching the spec's precedence. Doc comment updated to the four tiers.
- **`internal/version/version.go`** — additive `features` field in `version --json`.
- **`internal/extensions/types.go`** — `marketplace` added to the `Source` enumeration comment.

## Production safety

- **Flag off** (`FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=0`) → byte-for-byte the current behavior (path not scanned).
- **Flag on (default)** → a missing `~/.obey/fest/marketplace/extensions` is a harmless no-op, so users who never run obey-installer see **zero behavior change**. Users who do get installer-activated extensions loaded automatically (no init friction).
- No change to camp/fest plugin discovery — only this new extension tier.

## Verification

`just lint all` → 0 issues; `just test unit-raw` → 91 packages pass, no failures. `fest version --json` now includes `"features": ["marketplace_extension_source_v1"]`.

Tests (host temp dirs, `FEST_CONFIG_DIR`): flag-on loads with `Source==marketplace`; flag-off ignores; a same-named user-config extension overrides the marketplace one; a missing dir is benign.

## Default-on rationale (call out for review)

I defaulted the capability **on** (opt-out via env) rather than off, because the gated path is empty/missing for everyone who hasn't used obey-installer, so there's no behavior change for them, and it avoids requiring users to flip a flag to make installer-activated extensions load. Happy to flip to default-off if you'd rather gate it explicitly.

### DIFF
--- a/internal/extensions/loader.go
+++ b/internal/extensions/loader.go
@@ -5,8 +5,16 @@ import (
 	"path/filepath"
 
 	"github.com/Obedience-Corp/fest/internal/config"
+	"github.com/Obedience-Corp/fest/internal/features"
 	tpl "github.com/Obedience-Corp/fest/internal/template"
 )
+
+// MarketplaceExtensionsDir returns the obey-installer-managed extension path
+// (~/.obey/fest/marketplace/extensions). Extensions the installer activates
+// there are loaded when the marketplace_extension_source_v1 feature is on.
+func MarketplaceExtensionsDir() string {
+	return filepath.Join(config.ConfigDir(), "marketplace", ExtensionsDirName)
+}
 
 // ExtensionLoader handles loading extensions from multiple sources
 type ExtensionLoader struct {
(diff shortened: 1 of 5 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** This is a narrow, well-contained fest-side capability addition. The loader precedence is implemented in the right place, the marketplace path is feature-gated, and the version JSON advertises the new capability without disturbing existing fields.

**Key Findings:** No blocking issues found. I did not leave inline comments.

**What's Done Well:** The tests cover flag-on, flag-off, missing directory, and user-over-marketplace precedence, which are the important behavioral edges for this change.

**Staff Standard:** Yes, I would be comfortable submitting this change as my own work.

PR #226, opened 2026-07-01
feat(commit): use campaign name in commit tags

lancekrogers:
**Draft — blocked on [Obedience-Corp/camp#360](https://github.com/Obedience-Corp/camp/pull/360).** Marks the last step of the campaign-name commit-tag work (camp #358/#359 already merged).

## What

`fest` commits carried the hardcoded `OBEY-CAMPAIGN` brand instead of the campaign name. This switches all of `fest`'s tag sites to commitkit's name-style formatters:

- `internal/commands/commit/commit.go` — replaces the hand-rolled `fmt.Sprintf("[OBEY-CAMPAIGN-%s-%s]", ...)` festival tag **and** the `FormatCampaignTag` call with `FormatContextTagsFullNamed`.
- `internal/commands/status/handlers_festival.go` — `PrependCampaignTag` → `PrependContextTagsFullNamed`.
- `internal/commands/ritual/run.go` — `PrependCampaignTag` → `PrependContextTagsFullNamed`.

Result: `[obey-campaign:8deed8b4] ...`, `[obey-campaign:8deed8b4-FE-CW0003] ...`, etc. When the campaign name can't be resolved, the formatter falls back to the legacy `[OBEY-CAMPAIGN-...]` form, so behavior degrades gracefully exactly as before.

The festival case strips `OBEY-FE-` from `festRef` and passes the bare id, letting `FormatContextTagsFullNamed` re-add the `FE-` marker (matching camp's tag grammar).

## Dependency

Needs `commitkit.DetectCampaignName` / `LoadCampaignName` from camp#360. `go.mod` is pinned provisionally to that commit (`v0.3.0-rc.1.0...df9bc95`). **Before merge:** re-pin camp to a released/main commit once #360 lands. The obey-shared + sahilm/fuzzy go.mod changes are transitive requirements of the newer camp.

## Tests / gates

`go build ./...`, `go test ./...`, and `just lint all` (0 issues) green. End-to-end coverage lives in camp's currently-skipped `TestIntegration_CommitTags_FestCommit`, which can be un-skipped once camp releases a tag containing #360 and this lands.


obey-agent:
Ready for review. The camp dep is now pinned to the stable **v0.2.15** release (was a provisional pin to the #360 branch commit while it was open). `go build ./...`, `go test ./...`, and `just lint all` (0 issues) are green against the stable tag.

internal/commands/ritual/run.go
```diff
@@ -197,7 +197,8 @@ func autoCommitRitualRun(ctx context.Context, ritualName, ritualID, ritualPath,
 	if ws.Type == scope.WorkspaceTypeCampaign {
 		campaignID, err := commitkit.LoadCampaignID(ctx, ws.Root)
 		if err == nil && campaignID != "" {
-			message = commitkit.PrependCampaignTag(campaignID, message)
+			name, _ := commitkit.LoadCampaignName(ctx, ws.Root)
```
obey-agent, line 200: Requesting changes: this behavior change needs the existing ritual integration test updated. `tests/integration/ritual_run_test.go` still expects `[OBEY-CAMPAIGN-12345678]`, so the integration suite fails now that this path emits the name-style tag. I confirmed with `just build test-binary && DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock" TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run TestRitualRun_AutoCommit -timeout 5m ./tests/integration`: actual subject is `[testcampaign:12345678] ...`. Please update that expected prefix and its nearby comment mentioning `PrependCampaignTag`.

obey-agent (commented):
**Verdict: Request Changes**

The production change is directionally correct: `fest` now delegates tag formatting to `commitkit` and the dependency is pinned to the stable camp release containing the needed accessors. I found one blocking test drift: an existing ritual integration test still asserts the legacy campaign tag even though this PR changes that path to name-style tags.

Key finding:
- `TestRitualRun_AutoCommit/CampaignWorkspace` fails because it still expects `[OBEY-CAMPAIGN-12345678]`; actual output is `[testcampaign:12345678]`. See the inline comment on `internal/commands/ritual/run.go`.

What is done well:
- The implementation uses `commitkit` rather than duplicating campaign tag grammar.
- The camp dependency is now on stable `v0.2.15`, removing the provisional PR-branch pin.

Verification:
- `go test ./internal/commands/commit ./internal/commands/ritual ./internal/commands/status` passed.
- `go test ./...` passed.
- `just build test-binary && DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock" TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run TestRitualRun_AutoCommit -timeout 5m ./tests/integration` failed on the stale expected prefix.

Staff standard: I would merge after the integration expectation and comment are updated.

obey-agent (commented):
**Verdict: Comment — merge-ready** (leaving the Approve click for Lance)

Re-reviewed at head `b002c9c` after the agent's fix.

**Original finding reconciled:** The agent's fix to `TestRitualRun_AutoCommit/CampaignWorkspace` is correct. The test seeds `name: TestCampaign` / `id: 12345678-1234-...`, which slugifies to `testcampaign` and truncates to `12345678` (hex, so the name-style head emits rather than the legacy fallback), so `[testcampaign:12345678]` is exactly what the ritual path now produces. Verified against the setup, not just accepted.

**Completeness audit (beyond the one test the agent ran):** I grepped every fest test for stale `OBEY-CAMPAIGN` assertions. Only `tests/integration/ritual_run_test.go` sets up a campaign in the integration suite, so no other integration test exercised a campaign tag and none were missed. All unit tests pass.

**One additional fix applied in `b002c9c`:** the `commitWithCampaignSupport` doc comment in `internal/commands/commit/commit.go` still described the tag as `[OBEY-CAMPAIGN-{cid}-FE-{fid}]` — directly contradicting this PR. Corrected it to the name-style `[{campaign-name}:{cid}-FE-{fid}]`.

**What's done well:**
- Delegates to `commitkit` formatters instead of duplicating tag grammar; drops the hand-rolled `fmt.Sprintf` literal.
- Pinned to the stable `camp v0.2.15` release, not a PR-branch pseudo-version.
- Graceful legacy fallback preserved (empty/unresolvable name falls back to the old form).

**Coverage note (non-blocking):** the `fest commit` festival path (the `OBEY-FE-` -> bare-id stripping) has no fest-side integration test; its end-to-end coverage is camp's currently-skipped `TestIntegration_CommitTags_FestCommit`, which is now un-skippable since a released camp tag exists.

**Staff standard:** Yes — I would submit this. `go build ./...`, `go test ./...`, and `just lint all` (0 issues) are green at head.

PR #227, opened 2026-07-02
fix(status): crash-safe, atomic festival status moves (STATE-01/05)

lancekrogers:
Closes the two data-integrity findings from the Fable 2026-07-01 review (`workflow/code_reviews/fable-2026-07-01-code-review/fest/`), both adversarially re-verified (CONFIRMED, refute-first).

## STATE-01 (critical) — data-loss race in festival completion
`copyAndDelete` ran `os.RemoveAll(destPath)` on a destination it did not create. On two concurrent completions of the same festival: A renames source→dest and wins; B's rename fails because the source is gone; B falls into `copyAndDelete`; B's `Walk` errors on the missing source; the cleanup `RemoveAll(destPath)` deletes A's festival **and its `.fest` history**.

Fix:
- `MoveToDateDirectory` now falls back to copy only on a genuine cross-device (`EXDEV`) rename error. Any other failure (vanished source, permissions) surfaces instead of triggering a recursive copy.
- `copyAndDelete` stages the copy into a sibling `.partial` directory it owns and renames that into place only after a complete copy, so a failure never removes a foreign destination. It also refuses an already-existing destination.

## STATE-05 — non-atomic history writes
`status_history.json` was written with a truncate-in-place `os.WriteFile`, so a crash mid-write could corrupt it, and a corrupt file then made every later `RecordStatusChange` fail forever.

Fix:
- Writes go through a temp file + `rename` (mirroring `internal/chain/write.go`).
- A corrupt history file is quarantined to `.corrupt` and recording continues from empty instead of failing.

## Tests
New deterministic regressions (host `t.TempDir`, per fest convention): concurrent-completion preserves the festival, `copyAndDelete` preserves a foreign destination, corrupt history is quarantined and recording recovers.

## Gates (verbatim, at branch head)
- `just lint all` → `0 issues.` exit 0
- `just lint vet` → exit 0
- `just test unit` → `ALL 2324 TESTS PASSED` exit 0

## Scope note
This is the data-integrity slice of the fest-data-safety festival (FD0001). The remaining review items (STATE-04 single status-truth seam / MOD-02, NEXT-01 gate ordering, STATE-07 scaffold fest.yaml, CON-01 dungeon depth resolution, CON-02/03 JSON exit codes) follow as separate focused PRs so each stays reviewable.

### DIFF
--- a/internal/commands/status/date_directory.go
+++ b/internal/commands/status/date_directory.go
@@ -75,18 +75,22 @@ func GetCompletedPath(festivalsRoot, festivalName, dateDir string) string {
 }
 
 // copyAndDelete performs a copy+delete move across filesystems. It stages the
-// copy into a sibling ".partial" directory it owns and renames that into place
-// only after a complete copy, so a failure never removes a destination this
-// call did not create.
+// copy into a process-unique sibling directory it owns and renames that into
+// place only after a complete copy, so concurrent moves of the same festival
+// never delete each other's in-progress stage, and a failure never removes a
+// destination this call did not create.
 func copyAndDelete(sourcePath, destPath string) (string, error) {
-	stagePath := destPath + ".partial"
-	_ = os.RemoveAll(stagePath)
-
-	if err := os.MkdirAll(stagePath, 0755); err != nil {
+	destParent := filepath.Dir(destPath)
+	if err := os.MkdirAll(destParent, 0755); err != nil {
 		return "", errors.IO("creating staging directory", err)
 	}
 
-	err := filepath.Walk(sourcePath, func(path string, info os.FileInfo, err error) error {
+	stagePath, err := os.MkdirTemp(destParent, filepath.Base(destPath)+".partial-*")
+	if err != nil {
+		return "", errors.IO("creating staging directory", err)
+	}
+
+	err = filepath.Walk(sourcePath, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
--- a/internal/commands/status/date_directory.go
+++ b/internal/commands/status/date_directory.go
@@ -2,9 +2,11 @@
 package status
 
 import (
+	stderrors "errors"
 	"io"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/Obedience-Corp/fest/internal/errors"
@@ -45,11 +47,25 @@ func MoveToDateDirectory(sourcePath, completedDir, dateDir string) (string, erro
 
 	// Attempt atomic rename first
 	if err := os.Rename(sourcePath, destPath); err != nil {
-		// Fallback to copy+delete for cross-filesystem moves
+		if !isCrossDevice(err) {
+			return "", errors.IO("moving festival to date directory", err).
+				WithField("source", sourcePath).
+				WithField("destination", destPath)
+		}
 		return copyAndDelete(sourcePath, destPath)
 	}
 
 	return destPath, nil
+}
+
+// isCrossDevice reports whether a rename failed solely because source and
+// destination live on different filesystems.
+func isCrossDevice(err error) bool {
+	var linkErr *os.LinkError
+	if stderrors.As(err, &linkErr) {
+		return stderrors.Is(linkErr.Err, syscall.EXDEV)
+	}
+	return stderrors.Is(err, syscall.EXDEV)
 }
 
 // GetCompletedPath returns the full path for a completed festival with date organization.
@@ -58,16 +74,18 @@ func GetCompletedPath(festivalsRoot, festivalName, dateDir string) string {
 	return filepath.Join(festivalsRoot, "dungeon", "completed", dateDir, festivalName)
 }
 
-// copyAndDelete performs a copy+delete operation for cross-filesystem moves.
-// It copies all files recursively, then deletes the source.
-// If copy fails, no deletion occurs (atomic behavior).
+// copyAndDelete performs a copy+delete move across filesystems. It stages the
+// copy into a sibling ".partial" directory it owns and renames that into place
+// only after a complete copy, so a failure never removes a destination this
+// call did not create.
 func copyAndDelete(sourcePath, destPath string) (string, error) {
-	// Create destination directory
-	if err := os.MkdirAll(destPath, 0755); err != nil {
-		return "", errors.IO("creating destination directory", err)
+	stagePath := destPath + ".partial"
+	_ = os.RemoveAll(stagePath)
+
+	if err := os.MkdirAll(stagePath, 0755); err != nil {
+		return "", errors.IO("creating staging directory", err)
 	}
 
-	// Copy all files recursively
 	err := filepath.Walk(sourcePath, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
@@ -78,22 +96,31 @@ func copyAndDelete(sourcePath, destPath string) (string, error) {
 			return err
 		}
 
-		targetPath := filepath.Join(destPath, relPath)
+		targetPath := filepath.Join(stagePath, relPath)
 
 		if info.IsDir() {
 			return os.MkdirAll(targetPath, info.Mode())
 		}
 
-		// Copy file
 		return copyFile(path, targetPath, info.Mode())
 	})
 	if err != nil {
-		// Cleanup on failure - remove partial copy
-		_ = os.RemoveAll(destPath)
+		_ = os.RemoveAll(stagePath)
 		return "", errors.Wrap(err, "copying festival files")
 	}
 
-	// Delete source only after successful copy
+	if _, err := os.Stat(destPath); err == nil {
+		_ = os.RemoveAll(stagePath)
+		return "", errors.Validation("destination already exists").
+			WithField("destination", destPath)
+	}
+
+	if err := os.Rename(stagePath, destPath); err != nil {
+		_ = os.RemoveAll(stagePath)
+		return "", errors.IO("finalizing festival move", err).
+			WithField("destination", destPath)
+	}
+
 	if err := os.RemoveAll(sourcePath); err != nil {
 		return "", errors.IO("removing source after copy", err)
 	}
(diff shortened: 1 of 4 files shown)


internal/commands/status/date_directory.go
```diff
@@ -45,29 +47,45 @@ func MoveToDateDirectory(sourcePath, completedDir, dateDir string) (string, erro
...
+// copyAndDelete performs a copy+delete move across filesystems. It stages the
+// copy into a sibling ".partial" directory it owns and renames that into place
+// only after a complete copy, so a failure never removes a destination this
+// call did not create.
 func copyAndDelete(sourcePath, destPath string) (string, error) {
-	// Create destination directory
-	if err := os.MkdirAll(destPath, 0755); err != nil {
-		return "", errors.IO("creating destination directory", err)
+	stagePath := destPath + ".partial"
+	_ = os.RemoveAll(stagePath)
```
obey-agent: The staging path `destPath + ".partial"` is deterministic, not process-unique, and this unconditional `os.RemoveAll(stagePath)` runs at entry. Two concurrent cross-device completions of the same festival will delete each other's in-progress staging dir, causing spurious copy failures under exactly the concurrency scenario STATE-01 hardens. No data loss occurs (source is removed only after the atomic rename-into-place, and the destination-exists guard holds), but consider `os.MkdirTemp(filepath.Dir(destPath), festivalName+".partial-*")` for a collision-free stage. The added concurrency test only covers the same-FS rename path, so this EXDEV path is untested.

internal/commands/status/history.go
```diff
@@ -90,7 +128,11 @@ func LoadStatusHistory(ctx context.Context, festivalDir string) ([]StatusHistory
 
 	var history []StatusHistoryEntry
 	if err := json.Unmarshal(data, &history); err != nil {
-		return nil, errors.Wrap(err, "parsing status history")
+		quarantinePath := historyPath + ".corrupt"
+		if renameErr := os.Rename(historyPath, quarantinePath); renameErr != nil {
```
obey-agent: `LoadStatusHistory` now performs a filesystem mutation (rename to `.corrupt`) inside what is documented as a pure read ("loads the status history... Returns an empty slice if no history file exists"). Two issues: (a) the doc comment should note the quarantine side effect, or the quarantine should move into the record/write path; (b) `os.Rename` overwrites any existing `.corrupt`, so a second corruption event silently destroys the earlier forensic copy — prefer a timestamped quarantine name.

internal/commands/status/history.go
```diff
@@ -90,7 +128,11 @@ func LoadStatusHistory(ctx context.Context, festivalDir string) ([]StatusHistory
...
+			return nil, errors.Wrap(err, "parsing status history")
+		}
+		return []StatusHistoryEntry{}, nil
```
obey-agent, line 139: Recovering from corrupt history by resetting to empty is the right call, but it happens silently — the user is never told their status history was discarded (and `atomic.go` swallows `RecordStatusChange` errors, so there's no other signal). Emit a one-line stderr warning naming the `.corrupt` path so the reset is observable. Doesn't need to change control flow.

internal/commands/status/history.go
```diff
@@ -62,13 +62,51 @@ func RecordStatusChange(ctx context.Context, festivalDir, fromStatus, toStatus,
...
+	}
+	if err := tmp.Close(); err != nil {
+		_ = os.Remove(tmpPath)
+		return errors.IO("closing temp status history", err).WithField("path", tmpPath)
+	}
+	if err := os.Chmod(tmpPath, 0644); err != nil {
+		_ = os.Remove(tmpPath)
+		return errors.IO("setting status history mode", err).WithField("path", tmpPath)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
```
obey-agent, line 104: The temp file is `Sync()`'d, but the containing directory is not fsync'd after this rename, so the latest history write isn't durable against power loss (the anti-corruption invariant still holds — a crash reverts to the prior valid file, never a torn one). This matches the `internal/chain/write.go` precedent the PR cites, so acceptable; flagging only so the durability-vs-atomicity boundary of the "crash-safe" claim is explicit.

obey-agent (commented):

**Overview**
Closes two data-integrity findings. STATE-01: `MoveToDateDirectory` previously fell back to `copyAndDelete` on *any* rename error, and `copyAndDelete` ran `os.RemoveAll(destPath)` on a destination it did not own — so a concurrent completion (loser's rename fails ENOENT because the source already moved) could recursively delete the winner's festival and its `.fest` history. STATE-05: `status_history.json` was truncate-in-place `os.WriteFile`, so a crash mid-write corrupted it permanently. I traced both fixes against the real callers.

Both central claims hold:
- `isCrossDevice` properly unwraps `*os.LinkError` / `syscall.EXDEV`, so the destructive copy fallback fires *only* on genuine cross-filesystem moves; a vanished-source rename now surfaces as an error. This is the crux of STATE-01 and it's right.
- Same-filesystem (the common case) is a single atomic `os.Rename`; the concurrency test asserts exactly-one-success with content preserved.
- `copyAndDelete` stages into a `.partial` dir it owns, refuses an existing dest, renames stage->dest, removes source strictly last — no data-loss even in the racy cross-device case.
- `writeStatusHistoryAtomic` does temp-write + `Sync()` + `rename`; corrupt history is quarantined to `.corrupt` and recording recovers from empty.

**Key Findings** (all minor — see inline)
1. EXDEV staging path `destPath + ".partial"` is not process-unique + unconditional `RemoveAll` at entry -> two concurrent cross-device completions stomp each other (spurious failures, no data loss). `os.MkdirTemp` would make it collision-free.
2. `LoadStatusHistory` now mutates the FS (rename to `.corrupt`) inside a documented pure read, and `os.Rename` overwrites any prior `.corrupt` (loses forensics).
3. Silent history reset — user gets no warning their history was quarantined (and `atomic.go` swallows `RecordStatusChange` errors).
4. No parent-dir fsync after rename — latest write isn't durable vs power loss (anti-corruption invariant still holds); matches the `chain/write.go` precedent, noting to make the durability boundary explicit.

**What's Done Well**
EXDEV-only guard is the correct minimal fix (no fuzzy heuristics); source removed strictly after atomic rename-into-place; temp `Sync()`'d + explicit mode; project `errors` package throughout (no `fmt.Errorf`); genuine regression tests (two-goroutine concurrency + quarantine recovery), host `t.TempDir` per fest convention with the `resolvePath` symlink helper.

**Staff Standard**
Meets the bar for a production data-safety fix. Core atomicity/crash-safety verified for the common path and logically sound for the rare cross-device fallback. Merges are safe; the one I'd land soon is a process-unique staging dir (finding 1). Leaving the Approve for Lance.

internal/commands/status/date_directory.go
lancekrogers: Fixed in 317af4c. copyAndDelete now stages into a process-unique directory via os.MkdirTemp(filepath.Dir(destPath), filepath.Base(destPath)+".partial-*") instead of the deterministic destPath+".partial", so two concurrent EXDEV completions of the same festival can no longer collide on their staging dirs (and can no longer delete each other's in-progress copy).

Also had to restore an implicit MkdirAll(filepath.Dir(destPath)) before the MkdirTemp call — os.MkdirTemp requires the parent to already exist, unlike the old MkdirAll(stagePath, ...), which broke two pre-existing direct-call tests (TestCopyAndDelete, TestCrossFilesystemFallback) that relied on copyAndDelete creating nested dest dirs itself.

Added two new tests: TestCopyAndDelete_ConcurrentCompletionPreservesFestival exercises the EXDEV copy path with two real concurrent goroutines racing to complete the same festival (mirrors the existing rename-race test). I also added TestCopyAndDelete_DoesNotTouchLegacyDeterministicStagePath, which deterministically pins the exact defect: it seeds a foreign directory at the legacy destPath+".partial" path before calling copyAndDelete and asserts it survives untouched. I verified this second test fails against the pre-fix code and passes post-fix (the goroutine-based test alone didn't reliably reproduce the bug — the workload is too small for the race window to matter under Go's scheduler).

internal/commands/status/history.go
lancekrogers: Fixed in 317af4c. Went with option (a): updated LoadStatusHistory's doc comment to explicitly note the quarantine side effect, rather than moving the quarantine into the write path (RecordStatusChange is the only writer, but LoadStatusHistory is called directly by other callers like GetLastStatusChange, so the quarantine needs to stay where corruption is actually detected).

Also fixed the overwrite issue: added a quarantineCorruptHistory helper that picks the first free name in the sequence historyPath+".corrupt", ".corrupt.1", ".corrupt.2", ... so a second corruption event no longer destroys an earlier forensic copy. Added TestLoadStatusHistory_QuarantineDoesNotOverwriteEarlierCorrupt to cover it.

internal/commands/status/history.go
lancekrogers: Fixed in 317af4c. LoadStatusHistory now emits a one-line stderr warning naming the quarantine path when it discards corrupt history:

fest: discarded corrupt status history, quarantined at <path>.corrupt

No control flow change. Covered by TestLoadStatusHistory_WarnsOnQuarantine, which redirects os.Stderr through a pipe and asserts the message names the quarantine path.

internal/commands/status/history.go
lancekrogers: Acknowledged, no code change needed here — this matches the internal/chain/write.go precedent as you noted, and the anti-corruption invariant (crash reverts to the prior valid file, never a torn one) holds without a parent-dir fsync. Leaving as-is.

obey-agent (commented):

The current head `317af4c` addresses the prior hardening notes on the data-safety fix. `copyAndDelete` now uses a process-unique sibling staging directory via `os.MkdirTemp`, so concurrent EXDEV fallback moves do not share or delete a deterministic `.partial` path. Corrupt status history quarantine now uses unique `.corrupt` filenames and prints a stderr warning with the quarantine path.

Key findings:
- No blocking findings.
- The original STATE-01/STATE-05 fixes still hold: non-EXDEV rename failures no longer enter copy fallback, destination ownership is preserved, and history writes are temp-write/sync/rename.
- The follow-up tests cover concurrent copy fallback, legacy deterministic stage preservation, unique corrupt quarantines, and visible quarantine warning.

Verification:
- Reviewed changed files in a clean detached worktree at `317af4c9a4d89edb36a9b1adbe52af324ba95c1d`.
- `go test ./internal/commands/status` passed.

Staff standard: yes. I would be comfortable landing this data-safety change.

PR #228, opened 2026-07-02
fix(next,scaffold): earlier-gate ordering before cwd shortcut + scaffold writes fest.yaml (NEXT-01/STATE-07)

lancekrogers:
Implements the fest core-loop fixes from the Fable 2026-07-01 code review (FD0001 festival). All findings independently confirmed in the review's adversarial second pass.

## Findings closed

- **NEXT-01 (HIGH)** — `fest next` cwd shortcut bypassed earlier-phase gates. When cwd is inside a phase with a `position=before` WORKFLOW.md, `next.go` routed straight into that phase's workflow before the earlier-incomplete gate/after-workflow checks (which only ran on the selector fallback path). Running from inside a later phase could walk past an earlier phase's unsatisfied `GATES.md`. Fixed by hoisting the ordering check (new `routeEarlierIncompleteWork`) above the cwd shortcut so an earlier incomplete gate/workflow always routes first.
- **NEXT-02 / NEXT-03 (MEDIUM)** — the ordering checks failed open on a read error (`ewErr == nil &&` / `egErr == nil &&` silently skipped), and `FindFirstIncompletePhase`'s error was swallowed. Both now fail closed and propagate, matching the deliberately fail-closed remediation-gate path.
- **STATE-07 (HIGH)** — `fest scaffold from-plan` never wrote `fest.yaml`, so a scaffolded festival failed `fest next` in place with `festival root (no fest.yaml found) not found` (fest.yaml is the `ResolveFestivalPath` marker and the `EnforcePreActive` requirement). The scaffold command now writes `fest.yaml` via `config.SaveFestivalConfig` with festival metadata after the structure is generated.

## Tests

- `TestRouteEarlierIncompleteWork_HandlesEarlierGateFromLaterPhase` — an incomplete earlier gate routes before a later phase (would not route on the pre-hoist path).
- `TestRouteEarlierIncompleteWork_NotHandledWhenEarlierComplete` — no misfire when the earlier phase is complete.
- `TestRouteEarlierIncompleteWork_FailsClosedOnUnreadableFestival` — a read error returns handled+error (would return handled=false/nil under the old fail-open pattern).
- `TestWriteScaffoldFestYaml_ProducesRunnableMarker` — scaffold writes a loadable fest.yaml with correct metadata (fails without the fix, which wrote no fest.yaml).

## Gates (verbatim, branch head)

- `just lint all` → `0 issues.` exit 0
- `just lint vet` → exit 0
- `just test unit` → `ALL 2325 TESTS PASSED` exit 0

### DIFF
--- a/internal/commands/scaffold/from_plan.go
+++ b/internal/commands/scaffold/from_plan.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Obedience-Corp/fest/internal/config"
 	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/id"
+	"github.com/Obedience-Corp/fest/internal/progress"
 	sc "github.com/Obedience-Corp/fest/internal/scaffold"
 	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/ui"
@@ -158,17 +159,17 @@ func runFromPlan(ctx context.Context, opts *fromPlanOptions) error {
 
 	if !opts.DryRun {
 		campaignRoot := filepath.Dir(festivalsRoot)
-		if werr := writeScaffoldFestYaml(festivalDir, campaignRoot, festivalID, destCategory, plan); werr != nil {
+		if werr := writeScaffoldFestYaml(ctx, festivalDir, campaignRoot, festivalID, destCategory, plan); werr != nil {
 			return emitError(opts, werr)
 		}
-		result.FilesCreated = append(result.FilesCreated, filepath.Join(festivalDir, config.FestivalConfigFileName))
 	}
+	result.FilesCreated = append(result.FilesCreated, filepath.Join(festivalDir, config.FestivalConfigFileName))
 
 	// Emit result
 	return emitResult(opts, result, festivalID)
 }
 
-func writeScaffoldFestYaml(festivalDir, campaignRoot, festivalID, destCategory string, plan *sc.ParsedPlan) error {
+func writeScaffoldFestYaml(ctx context.Context, festivalDir, campaignRoot, festivalID, destCategory string, plan *sc.ParsedPlan) error {
 	now := time.Now().UTC()
 	cfg := config.DefaultFestivalConfig()
 	cfg.Metadata = config.FestivalMetadata{
@@ -186,6 +187,9 @@ func writeScaffoldFestYaml(festivalDir, campaignRoot, festivalID, destCategory s
 				Notes:     "Festival scaffolded from plan",
 			},
 		},
+	}
+	if initialSize, sizeErr := progress.ComputeDirectorySize(ctx, festivalDir); sizeErr == nil && initialSize > 0 {
+		cfg.Metadata.InitialSizeBytes = initialSize
 	}
 	if err := config.SaveFestivalConfig(festivalDir, campaignRoot, cfg); err != nil {
 		return errors.Wrap(err, "writing fest.yaml").WithField("path", filepath.Join(festivalDir, config.FestivalConfigFileName))
--- a/internal/commands/scaffold/from_plan.go
+++ b/internal/commands/scaffold/from_plan.go
@@ -7,12 +7,16 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/Obedience-Corp/fest/internal/config"
+	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/id"
 	sc "github.com/Obedience-Corp/fest/internal/scaffold"
 	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/ui"
 	"github.com/Obedience-Corp/fest/internal/workspace"
+	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


internal/commands/scaffold/from_plan.go
```diff
@@ -152,10 +156,43 @@ func runFromPlan(ctx context.Context, opts *fromPlanOptions) error {
...
+	}
+
 	// Emit result
 	return emitResult(opts, result, festivalID)
 }
 
+func writeScaffoldFestYaml(festivalDir, campaignRoot, festivalID, destCategory string, plan *sc.ParsedPlan) error {
+	now := time.Now().UTC()
+	cfg := config.DefaultFestivalConfig()
+	cfg.Metadata = config.FestivalMetadata{
```
obey-agent, line 175: Parity gap with `create_festival.go:writeFestYaml`: the scaffolded metadata omits `InitialSizeBytes`. `create festival` sets it via a follow-up `SaveFestivalConfig` so the token-proxy/content-delta baseline is the festival's initial `.md` size. Here the baseline stays 0, so the first delta computation over-counts the entire scaffolded corpus as "added." Consider setting `InitialSizeBytes` after the structure is generated, matching the create path, so scaffolded festivals report accurate deltas.

internal/commands/scaffold/from_plan.go
```diff
@@ -152,10 +156,43 @@ func runFromPlan(ctx context.Context, opts *fromPlanOptions) error {
 		return emitError(opts, err)
 	}
 
+	if !opts.DryRun {
```
obey-agent, line 160: Dry-run preview under-reports. Both the write and the `result.FilesCreated` append are gated on `!opts.DryRun`, so `fest scaffold from-plan --dry-run` will not list `fest.yaml` among files-to-be-created even though a real run always creates it. For an accurate preview, append the `fest.yaml` path to `result.FilesCreated` unconditionally (the append is cheap and does not write) and keep only the actual `writeScaffoldFestYaml` call gated behind `!opts.DryRun`.

obey-agent (commented):

**Overview**
Two production core-loop fixes plus fail-closed hardening. NEXT-01: `fest next` hoists `routeEarlierIncompleteWork` above the cwd `position=before` WORKFLOW.md shortcut, so an earlier phase's unsatisfied after-workflow/GATES.md always routes before a later phase you happen to be `cd`'d into. NEXT-02/03: `FindFirstIncompletePhase` and the earlier-after-workflow/gate checks now propagate read errors instead of silently failing open. STATE-07: `fest scaffold from-plan` now writes `fest.yaml`, which `FindFestivalRoot` requires as the root marker.

**Key Findings** (both minor — see inline)
- NEXT-01 ordering change is correct and regression-safe: `routeEarlierIncompleteWork` only scans phases numerically *before* the passed phase (`phaseName >= currentPhaseName` breaks), so for the earliest phase it returns `(false,nil)` and the cwd shortcut behaves exactly as before. Only diverts when an earlier after-workflow/gate is genuinely incomplete.
- Fail-closed changes are safe: the only error source is the top-level `os.ReadDir(festivalPath)`, and by the time `runNext` reaches these `FindFestivalRoot` already confirmed the root exists, so the new `return err` paths are effectively unreachable for a valid festival.
- STATE-07 `fest.yaml` is complete, not a stub: `writeScaffoldFestYaml` mirrors `create_festival.go:writeFestYaml` (DefaultFestivalConfig + full metadata + StatusHistory). No clobber risk (`runFromPlan` hard-errors if the dir exists).
- Minor parity: scaffolded metadata omits `InitialSizeBytes` (create festival sets it) -> first content-delta over-counts the whole corpus. And dry-run under-reports `fest.yaml` in `FilesCreated`.

**What's Done Well**
Tight single-responsibility extraction unifying three previously-inline fail-open checks into one fail-closed helper reused by both cwd and selector paths; behavioral tests reuse the established fixtures and correct `gate:<phase>` progress-key convention, asserting the pre-hoist bug, the no-misfire case, and the fail-closed case; fest.yaml kept consistent with the canonical create path.

**Staff Standard**
Meets it — clean, correct, well-tested fix for a real production core-loop defect with genuinely fail-safe error handling. Findings are minor parity/UX nits only. Leaving the Approve for Lance.

internal/commands/scaffold/from_plan.go
lancekrogers: Fixed in b843dd0. Moved the `campaign.Metadata.InitialSizeBytes` assignment into `writeScaffoldFestYaml` — since `runner.Run()` already generates the full phase/sequence/task structure before this function writes fest.yaml, computing `progress.ComputeDirectorySize(ctx, festivalDir)` at that point captures the same post-structure baseline `create_festival.go`'s two-phase `recordInitialSize` achieves, in a single write. Added `TestWriteScaffoldFestYaml_SetsInitialSizeBytes` asserting the baseline matches the scaffolded .md content size.

internal/commands/scaffold/from_plan.go
lancekrogers: Fixed in b843dd0. The `result.FilesCreated` append is now unconditional; only the actual `writeScaffoldFestYaml` call (the write) stays gated behind `!opts.DryRun`. Added `TestRunFromPlan_DryRun_ReportsFestYamlInFilesCreated`, which runs `runFromPlan` end-to-end with `--dry-run --json` and asserts fest.yaml appears in `files_created` while confirming no file is actually written to disk.

obey-agent (commented):

The current head `b843dd0` keeps the core-loop ordering fix intact and addresses the prior scaffold parity notes. `from-plan` now sets `InitialSizeBytes` from the generated scaffold content and includes `fest.yaml` in `FilesCreated` even for dry-run output without actually writing it during dry-run.

Key findings:
- No blocking findings.
- The NEXT-01/NEXT-02/NEXT-03 route remains fail-closed and still routes earlier incomplete work before later cwd shortcuts.
- The STATE-07 scaffold marker fix is now closer to the canonical create path: it writes loadable `fest.yaml`, preserves status history metadata, records initial size, and previews `fest.yaml` in dry-run.

Verification:
- Reviewed changed files in a clean detached worktree at `b843dd01c04f0aa0dc83f4039f7f511e63cde553`.
- `go test ./internal/commands/next ./internal/commands/scaffold` passed.

Staff standard: yes. This is a clean production core-loop fix with focused regression coverage.

PR #229, opened 2026-07-02
fix(show,promote,dungeon): non-zero JSON exit codes + depth-tolerant dungeon resolution (CON-01/02/03)

lancekrogers:
Implements three findings from the Fable 2026-07-01 code review (fest area), all adversarially confirmed in `second-pass-fable.md`.

## CON-01 (HIGH) — depth-tolerant dungeon resolution
Festival resolution recursed exactly one level into dungeon buckets whose name matched `^\d{4}-\d{2}(-\d{2})?$`, so a festival under `festivals/dungeon/someday/pre-2026-03-01/` was unresolvable (`pre-` fails the pattern, recursion skipped). This is the fest-side cause of fest-gif's silent empty GIF and of `fest show --festival <nested> --json` returning `not found`.

Fix: collapse `detect.go`'s four date-gated recursions into one bounded-depth walker (`festivalDirsUnderStatus`) that descends any organizational subfolder up to a depth cap, then resolves the festival marker. Covers `FindFestivalByName` (the `fest show --festival` path), `findLinkedFestivalPath` (linked detection), `ListFestivalsByStatus`, and `ListFestivalsByStatusLight` (`fest show all`).

## CON-02 / CON-03 (HIGH) — non-zero exit on JSON error
`fest show --json` and `fest promote --json` emitted an error body but returned `nil`, so the process exited 0 on error and any exit-code-checking consumer saw false success. Both now return the existing `ErrAlreadyPrinted` sentinel, so the process exits 1 while still printing the JSON body (and `main.go` skips the stderr double-print).

## Scope note (explicit tradeoff)
Three other date-pattern sites (`status/helpers.go`, `shared/festival_selector.go`, `navigation/completions.go`) retain date-bucket assumptions and are a documented follow-up. `detect.go` is the path behind the reported bug; `status/helpers.go` in particular feeds the data-mutating status-move path being changed concurrently in the FD0001 data-safety work, so it is left untouched here to avoid conflict.

## Tests / gates
New regression tests: nested `pre-2026-03-01` festival resolves via `FindFestivalByName` and lists via `ListFestivalsByStatus`; `emitShowErrorJSON` returns the exit-1 sentinel. Gates at branch head: `just lint all` 0 issues, `just lint vet` clean, `just test unit` ALL 2324 TESTS PASSED.

### DIFF
--- a/internal/commands/show/detect.go
+++ b/internal/commands/show/detect.go
@@ -93,7 +93,8 @@ func DetectCurrentFestival(ctx context.Context, startDir, campaignRoot string) (
 // directories, including festivals nested under organizational buckets.
 func findLinkedFestivalPath(festivalsRoot, name string) string {
 	for _, status := range id.StatusDirectories {
-		for _, entry := range festivalDirsUnderStatus(filepath.Join(festivalsRoot, status)) {
+		deep := strings.HasPrefix(status, "dungeon/")
+		for _, entry := range festivalDirsUnderStatus(filepath.Join(festivalsRoot, status), deep) {
 			if filepath.Base(entry.path) == name {
 				return entry.path
 			}
@@ -103,22 +104,35 @@ func findLinkedFestivalPath(festivalsRoot, name string) string {
 }
 
 // maxDungeonNestDepth bounds how deep festival resolution recurses under a
-// status directory when festivals are filed inside organizational subfolders
-// (date buckets like 2026-03-01, or free-form buckets like pre-2026-03-01).
+// dungeon status directory when festivals are filed inside organizational
+// subfolders (date buckets like 2026-03-01, or free-form buckets like
+// pre-2026-03-01).
 const maxDungeonNestDepth = 3
 
-// festivalDir is a resolved festival root plus the immediate bucket it was
-// nested under, if any.
+// maxWorkingStatusNestDepth bounds recursion under a working status
+// directory (active/ready/planning/ritual) to a single date-bucket level,
+// matching the shallow lookup those statuses have always used.
+const maxWorkingStatusNestDepth = 1
+
+// festivalDir is a resolved festival root plus the outermost organizational
+// bucket it was nested under below the status dir, if any.
 type festivalDir struct {
 	path   string
 	bucket string
 }
 
-// festivalDirsUnderStatus returns every festival root at or below statusDir,
-// recursing through non-festival subdirectories up to maxDungeonNestDepth so a
-// festival filed under any organizational bucket resolves, not only ones whose
-// bucket name matches a date pattern.
-func festivalDirsUnderStatus(statusDir string) []festivalDir {
+// festivalDirsUnderStatus returns every festival root at or below statusDir.
+// When deep is true (dungeon statuses), it recurses through any non-festival
+// subdirectory up to maxDungeonNestDepth so a festival filed under any
+// organizational bucket resolves. When deep is false (working statuses), it
+// only recurses one level into date-bucket-shaped subdirectories, preserving
+// the prior shallow lookup for active/ready/planning/ritual.
+func festivalDirsUnderStatus(statusDir string, deep bool) []festivalDir {
+	maxDepth := maxWorkingStatusNestDepth
+	if deep {
+		maxDepth = maxDungeonNestDepth
+	}
+
 	var out []festivalDir
 	var walk func(dir, bucket string, depth int)
 	walk = func(dir, bucket string, depth int) {
@@ -135,13 +149,17 @@ func festivalDirsUnderStatus(statusDir string) []festivalDir {
 				out = append(out, festivalDir{path: child, bucket: bucket})
 				continue
 			}
-			if depth < maxDungeonNestDepth {
-				nextBucket := bucket
-				if nextBucket == "" {
-					nextBucket = entry.Name()
-				}
-				walk(child, nextBucket, depth+1)
+			if depth >= maxDepth {
+				continue
 			}
+			if !deep && !LooksLikeDateDir(entry.Name()) {
+				continue
+			}
+			nextBucket := bucket
+			if nextBucket == "" {
+				nextBucket = entry.Name()
+			}
+			walk(child, nextBucket, depth+1)
 		}
 	}
 	walk(statusDir, "", 0)
@@ -176,19 +194,39 @@ func FindFestivalByName(ctx context.Context, festivalsDir, name, campaignRoot st
 	if ctx == nil {
 		ctx = context.Background()
 	}
+
+	var loose *festivalDir
+	var looseStatus string
+
 	for _, status := range id.StatusDirectories {
-		for _, entry := range festivalDirsUnderStatus(filepath.Join(festivalsDir, status)) {
+		deep := strings.HasPrefix(status, "dungeon/")
+		for _, entry := range festivalDirsUnderStatus(filepath.Join(festivalsDir, status), deep) {
 			base := filepath.Base(entry.path)
-			if base != name && !strings.HasPrefix(base, name+"_") && !strings.Contains(base, name) {
-				continue
+			switch {
+			case base == name || strings.HasPrefix(base, name+"_"):
+				info, err := parseFestivalInfo(ctx, entry.path, campaignRoot)
+				if err != nil {
+					continue
+				}
+				info.Status = status
+				if entry.bucket != "" {
+					info.StatusDate = entry.bucket
+				}
+				return info, nil
+			case loose == nil && strings.Contains(base, name):
+				match := entry
+				loose = &match
+				looseStatus = status
 			}
-			info, err := parseFestivalInfo(ctx, entry.path, campaignRoot)
-			if err != nil {
-				continue
-			}
-			info.Status = status
-			if entry.bucket != "" {
-				info.StatusDate = entry.bucket
+		}
+	}
+
+	if loose != nil {
+		info, err := parseFestivalInfo(ctx, loose.path, campaignRoot)
+		if err == nil {
+			info.Status = looseStatus
+			if loose.bucket != "" {
+				info.StatusDate = loose.bucket
 			}
 			return info, nil
 		}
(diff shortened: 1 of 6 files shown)


internal/commands/show/detect.go
```diff
@@ -89,38 +89,63 @@ func DetectCurrentFestival(ctx context.Context, startDir, campaignRoot string) (
...
-		if strings.HasPrefix(status, "dungeon/") {
-			statusDir := filepath.Join(festivalsRoot, status)
-			entries, err := os.ReadDir(statusDir)
-			if err != nil {
+// maxDungeonNestDepth bounds how deep festival resolution recurses under a
+// status directory when festivals are filed inside organizational subfolders
+// (date buckets like 2026-03-01, or free-form buckets like pre-2026-03-01).
+const maxDungeonNestDepth = 3
+
+// festivalDir is a resolved festival root plus the immediate bucket it was
```
obey-agent: Comment says "the immediate bucket it was nested under," but `walk` fixes `bucket` at the first descent level and never updates it, so for `status/A/B/fest` it records the outermost bucket `A`, not the immediate parent `B`. In the real `dungeon/<sub>/<date>/fest` layout these coincide, so the value is correct — but fix the comment to say "outermost organizational bucket below the status dir" to avoid future confusion.

internal/commands/show/detect.go
```diff
@@ -152,55 +177,20 @@ func FindFestivalByName(ctx context.Context, festivalsDir, name, campaignRoot st
...
-		entries, err := os.ReadDir(statusDir)
-		if err != nil {
-			continue // Skip inaccessible directories
-		}
-
-		for _, entry := range entries {
-			if !entry.IsDir() {
+		for _, entry := range festivalDirsUnderStatus(filepath.Join(festivalsDir, status)) {
+			base := filepath.Base(entry.path)
+			if base != name && !strings.HasPrefix(base, name+"_") && !strings.Contains(base, name) {
```
obey-agent: The preserved loose match `strings.Contains(base, name)` now runs against a larger, deeper candidate set (every festival up to depth 3 under every status). This widens an existing false-positive surface: `fest show FA0001` can match `FA00010`, and a deeply nested dungeon festival whose name merely contains the query can be returned (first in DFS order) ahead of the intended one. Pre-existing, but since you're refactoring this call site, consider tightening to exact + `name+"_"` prefix and dropping the substring branch, or ordering exact matches first.

internal/commands/show/detect.go
```diff
@@ -89,38 +89,63 @@ func DetectCurrentFestival(ctx context.Context, startDir, campaignRoot string) (
...
 			}
-			for _, entry := range entries {
-				if entry.IsDir() && LooksLikeDateDir(entry.Name()) {
-					datePath := filepath.Join(statusDir, entry.Name(), name)
-					if info, err := os.Stat(datePath); err == nil && info.IsDir() {
-						if isValidFestival(datePath) {
-							return datePath
-						}
-					}
+			if depth < maxDungeonNestDepth {
```
obey-agent: The depth-3 walk now applies to every status, including `active`/`ready`/`planning`/`ritual`, which previously only surfaced direct children (or one-level date buckets). A stray non-festival wrapper dir in `active/` (e.g. `active/backup/old-fest/`) would now surface `old-fest` as an active festival with `bucket="backup"`. Festivals are normally direct children of working statuses, so risk is low, but the broadening is silent — worth a comment noting the recursion is intentional for all statuses, and/or restricting deep recursion to `dungeon/*` where the nesting actually occurs.

obey-agent (commented):
## Staff Review — Verdict: REQUEST CHANGES

**Overview**
Three things: (1) collapses four date-gated one-level recursions in `show/detect.go` into a single bounded-depth walker (cap 3) so festivals filed under non-date buckets resolve; (2) makes `show.emitShowErrorJSON` return `ErrAlreadyPrinted` so `fest show --json` exits 1 on error while still printing the JSON body; (3) makes three `promoteCore` `--json` error branches return the same sentinel. I traced the exit-code path end-to-end: `main.go` skips the stderr double-print only for `ErrAlreadyPrinted` and always `os.Exit(1)`; `root.go` sets `SilenceErrors`/`SilenceUsage`. That mechanism is correct, and the dungeon walker is symlink-safe (`entry.IsDir()` is false for symlinks) and stops descending on a valid festival.

**The core mechanism is right but the promote `--json` contract was left half-fixed — the exact class of bug the PR set out to close.**

**Blocking findings**
- **(major) `runPromote` resolve-failure `--json` path still exits 0 — `internal/commands/promote/promote.go` ~line 178.** When festival resolution fails (`fest promote <nonexistent> --json`, or promote run outside a festival — the picker is disabled because `allowPicker = !opts.json`), the handler emits `{"success":false,...}` but `return shared.EncodeJSON(...)` returns `nil`, so the process exits 0. `fest promote bad --json; echo $?` prints the false-success body but exits 0. This directly contradicts CON-02/03 and is inconsistent with the three `promoteCore` paths the PR just fixed. Encode the body, then `return "", errors.ErrAlreadyPrinted`. (This line is in `runPromote`, unchanged by the PR, so it couldn't be flagged inline — but it's the headline defect.)
- **(major) Invalid `--dungeon` value emits no JSON body in `--json` mode — `internal/commands/promote/promote.go` ~line 205.** `fest promote --dungeon bogus --json` returns a plain `errors.Validation` (exit 1, correct) but writes text to stderr and nothing to stdout, so `... --json | jq` gets empty input and fails to parse. For consistency it should encode a `success:false` JSON body like the other promote error paths. (Also on an unchanged line — see this summary.)
- **(test gap)** No promote `--json` exit-code test — precisely why the :178 regression slipped through. Add a behavioral test asserting non-zero exit + parseable JSON for the resolve-failure and invalid-transition paths.

**Minor (inline)**
- `detect.go:110` — bucket comment inaccurate (records outermost bucket, not immediate parent).
- `detect.go:182` — loose `strings.Contains(base, name)` now runs over a deeper candidate set, widening the false-positive surface (`FA0001` matches `FA00010`).
- `detect.go:138` — depth-3 walk now applies to `active`/`ready`/`planning`/`ritual` too; a stray wrapper dir in `active/` could surface a bogus festival. Consider restricting deep recursion to `dungeon/*`.

**What's Done Well**
Genuine simplification (143 deletions removing four near-duplicate blocks); symlink-safe by construction; the `show --json` side is fully consistent (every error path routes through `emitShowErrorJSON`); regression tests assert both `Status` and `StatusDate` for the nested case.

**Staff Standard**
Not yet. The mechanism is sound and the show/dungeon work is solid, but the headline promise — consistent non-zero JSON exit across show *and promote* — is not met: one promote `--json` path exits 0 on error and another emits no JSON body. Because fest is in production and agent/festival-app consumers key off exit code + `--json`, these are contract defects, not polish. Also worth a release note: `show/promote --json` on error now exits 1 where it previously exited 0 — confirm festival-app treats exit 1 + stdout JSON as a parseable error, not a hard crash.

internal/commands/show/detect.go
lancekrogers: Fixed in 3be0d57. Updated the festivalDir struct comment to: "festivalDir is a resolved festival root plus the outermost organizational bucket it was nested under below the status dir, if any."

internal/commands/show/detect.go
lancekrogers: Fixed in 3be0d57. `festivalDirsUnderStatus` now takes a `deep bool`; dungeon/* statuses still get the depth-3 walk into any non-festival subdirectory, but active/ready/planning/ritual now cap recursion at 1 level and only descend into subdirectories that look like date buckets (`LooksLikeDateDir`), matching the prior shallow lookup. A wrapper dir like `active/backup/old-fest/` no longer surfaces. Threaded the flag through all 4 callers (findLinkedFestivalPath, FindFestivalByName, ListFestivalsByStatus, ListFestivalsByStatusLight) via `strings.HasPrefix(status, "dungeon/")`. Added TestListFestivalsByStatus_ActiveWrapperDirNotSurfaced covering this.

internal/commands/show/detect.go
lancekrogers: Fixed in 3be0d57. FindFestivalByName now scans all statuses for an exact match (`base == name`) or prefix match (`name+"_"`) first and returns immediately on either; the loose `strings.Contains` match is only recorded as a fallback (first one seen) and is only used if no exact/prefix match is found anywhere. So `fest show FA0001` can no longer resolve to `FA00010` when an exact `FA0001` exists, regardless of directory walk order. Added TestFindFestivalByName_ExactMatchPreferredOverSubstring covering this.

lancekrogers:
Both promote `--json` exit-code blockers from the review are now fixed in 1afe4f6 (the three `detect.go` minors were already handled in 3be0d57):

- **`runPromote` resolve-failure `--json` path (~promote.go:177)** returned `shared.EncodeJSON(...)` (nil on success), so `fest promote <nonexistent> --json` printed `success:false` but exited 0. It now encodes the body then returns `errors.ErrAlreadyPrinted`, so it exits 1 with a parseable body.
- **`promoteCore` invalid `--dungeon` value (~promote.go:204)** wrote text to stderr and nothing to stdout under `--json`. It now emits a `success:false` JSON body first, matching the sibling invalid-transition and readiness paths, then returns `ErrAlreadyPrinted`.

Added behavioral tests — `TestRunPromote_ResolveFailureJSONExitsNonZeroWithBody` and `TestPromoteCore_InvalidDungeonJSONExitsNonZeroWithBody` — asserting non-zero exit (`ErrAlreadyPrinted`) plus a parseable `success:false` JSON body for both paths, closing the test-gap you flagged.

On the release-note point: `show`/`promote --json` on error now exit 1 (previously 0) while still printing the JSON body on stdout. festival-app should treat exit 1 + stdout JSON as a parseable error, not a hard crash.

obey-agent (commented):

The current head `1afe4f6` fixes the prior promote JSON contract blockers. `runPromote` now emits a JSON failure body and returns `ErrAlreadyPrinted` on selector/resolve failure under `--json`, and invalid `--dungeon` under `--json` now also emits parseable JSON before returning the sentinel for non-zero process exit.

Key findings:
- No blocking findings.
- The dungeon resolution changes remain bounded and safer than the initial version: deep recursion is limited to dungeon statuses, working statuses preserve the prior shallow date-bucket behavior, and exact matches win before loose substring matches.
- The new tests assert the two previously-blocking promote JSON paths return the non-zero sentinel with parseable `success:false` bodies.

Verification:
- Reviewed changed files in a clean detached worktree at `1afe4f61383e5dd453c4b55e71ac4296967c4360`.
- `go test ./internal/commands/promote ./internal/commands/show` passed.

Staff standard: yes. The show/promote JSON contract issue from the last review is closed.

PR #230, opened 2026-07-02
fix(create): append tasks at end by default; report renumbering; tolerate missing pre-move paths in promote auto-commit

lancekrogers:
PR-1 of 3 from the fest planning-loop UX audit (obey-campaign design workitem `workflow/design/fest-planning-loop-ux-fixes`, WI-7b386b; findings recorded live in the CA0002 festival's `fest feedback` ledger while planning a real festival through the `fest next` loop).

## F1: `fest create task` reversed dictation order (P0)

`--after` defaulted to 0, inserting every new task at position 01 and silently renumbering existing tasks down. Creating tasks one-per-invocation in intended order produced exactly reversed sequences, and the success echo printed the pre-renumber `01_<name>` for every creation, masking it (17 consecutive creations across 6 sequences all echoed `01_` during the audit).

- Default is now `-1`: append at end, auto-detected from the highest existing task number via the same pattern `fest create sequence` already uses (`create_sequence.go` had this exact design; task creation was inconsistent with its sibling).
- `--after 0` still inserts at the beginning; `--after N` unchanged.
- The echo now reports renumbering (`Renumbered N existing task(s)` with old -> new names), and the JSON contract's already-existing `renumbered` field is populated instead of always `[]`.
- `Renumberer` gains a `Changes()` accessor so callers can report the renames they trigger.

Compat: behavior change called out deliberately; prepend-by-default produced reversed plans and was undocumented, and batch `--name` invocations (the documented happy path) are unaffected in their result ordering. Explicit insertion is fully preserved.

## F4: `fest promote` auto-commit failed on never-committed festivals (P0)

`AutoCommitStatusChange` staged all changed paths in one `git add`; a pathspec matching nothing (the pre-move directory of a festival that had never been committed) failed the entire staging, so promote succeeded but the lifecycle transition landed uncommitted behind a dim warning. Paths now stage independently: a no-match path is skipped, staging fails only when NO path can be staged, and that error names every failed path. Regression test drives a real promote-shaped move with a missing pre-move path and asserts the commit contents.

## F8: scoping-flag errors teach the cwd model (P2 rider)

`fest create sequence --phase X` and friends now append a hint ("sequences are created in the current phase directory; cd 003_IMPLEMENT first") instead of a bare `unknown flag` error. Static mapping on the create subcommands only; unrelated unknown flags unchanged.

## Testing

- New: append/passthrough/empty-dir resolution tests, dictation-order integration of resolve+insert, `Changes()` rename reporting, missing-pre-move-path and all-paths-fail auto-commit tests, flag-hint golden tests.
- `just test unit` (2329 passed), `just test integration` (59 passed), `just lint all` (0 issues) at head.

PR-2 (workflow guidance rendering) and PR-3 (feedback subsystem + delegated approvals) follow per the design package.

### DIFF
--- a/internal/commands/festival/create_task.go
+++ b/internal/commands/festival/create_task.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/Obedience-Corp/fest/internal/commands/shared"
@@ -136,15 +137,16 @@ Run 'fest validate tasks' to verify task files exist in implementation sequences
 
 // resolveTaskInsertAfter maps the --after default (-1) to append-at-end by
 // detecting the highest existing task number in the sequence directory,
-// matching create sequence's behavior.
-func resolveTaskInsertAfter(ctx context.Context, sequenceDir string, after int) int {
+// matching create sequence's behavior. A parse failure is an error, not a
+// silent prepend into a possibly non-empty sequence.
+func resolveTaskInsertAfter(ctx context.Context, sequenceDir string, after int) (int, error) {
 	if after != -1 {
-		return after
+		return after, nil
 	}
 	parser := festival.NewParser()
 	tasks, err := parser.ParseTasks(ctx, sequenceDir)
-	if err != nil || len(tasks) == 0 {
-		return 0
+	if err != nil {
+		return 0, errors.Wrap(err, "detecting last task number").WithField("dir", sequenceDir)
 	}
 	maxNum := 0
 	for _, t := range tasks {
@@ -152,7 +154,34 @@ func resolveTaskInsertAfter(ctx context.Context, sequenceDir string, after int) 
 			maxNum = t.Number
 		}
 	}
-	return maxNum
+	return maxNum, nil
+}
+
+// netRenames collapses per-step rename chains from batch inserts (a->b in
+// one step, b->c in a later step) into net a->c entries, sorted by the
+// original name for stable, readable output.
+func netRenames(steps [][2]string) []string {
+	finalOf := map[string]string{}
+	originOf := map[string]string{}
+	var order []string
+	for _, s := range steps {
+		oldName, newName := s[0], s[1]
+		if origin, ok := originOf[oldName]; ok {
+			delete(originOf, oldName)
+			originOf[newName] = origin
+			finalOf[origin] = newName
+			continue
+		}
+		originOf[newName] = oldName
+		finalOf[oldName] = newName
+		order = append(order, oldName)
+	}
+	sort.Strings(order)
+	out := make([]string, 0, len(order))
+	for _, o := range order {
+		out = append(out, fmt.Sprintf("%s -> %s", o, finalOf[o]))
+	}
+	return out
 }
 
 // RunCreateTask executes the create task command logic.
@@ -206,12 +235,16 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 	mgr := tpl.NewManager()
 	loader := tpl.NewLoader()
 
-	opts.After = resolveTaskInsertAfter(ctx, absPath, opts.After)
+	after, afterErr := resolveTaskInsertAfter(ctx, absPath, opts.After)
+	if afterErr != nil {
+		return emitCreateTaskError(opts, afterErr)
+	}
+	opts.After = after
 
 	// Track all created tasks for output
 	var createdTasks []map[string]any
 	var createdPaths []string
-	var renumbered []string
+	var renameSteps [][2]string
 	var totalMarkersFilled, totalMarkersCount int
 	currentAfter := opts.After
 
@@ -229,8 +262,8 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 		}
 		for _, ch := range ren.Changes() {
 			if ch.Type == festival.ChangeRename {
-				renumbered = append(renumbered,
-					fmt.Sprintf("%s -> %s", filepath.Base(ch.OldPath), filepath.Base(ch.NewPath)))
+				renameSteps = append(renameSteps,
+					[2]string{filepath.Base(ch.OldPath), filepath.Base(ch.NewPath)})
 			}
 		}
 
@@ -380,6 +413,7 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 	}
 
 	// Output results
+	renumbered := netRenames(renameSteps)
 	remainingMarkers := totalMarkersCount - totalMarkersFilled
 
 	if opts.JSONOutput {
--- a/internal/commands/festival/create_task.go
+++ b/internal/commands/festival/create_task.go
@@ -76,16 +76,19 @@ TEMPLATE VARIABLES (automatically set from --name):
   {{ full_path }}            Complete path from festival root
 
 EXAMPLES:
-  # Create single task in current sequence
+  # Create single task in current sequence (appends at the end)
   fest create task --name "design endpoints" --json
 
-  # Create multiple tasks at once (sequential numbering)
+  # Create multiple tasks at once (sequential numbering, appended at end)
   fest create task --name "requirements" --name "design" --name "implement"
   # Creates: 01_requirements.md, 02_design.md, 03_implement.md
 
-  # Create tasks starting after position 2
+  # Insert tasks after position 2 (existing tasks renumber down, reported)
   fest create task --after 2 --name "new step" --name "another step"
   # Creates: 03_new_step.md, 04_another_step.md
+
+  # Insert at the beginning
+  fest create task --after 0 --name "prerequisite"
 
   # Create task in specific sequence
   fest create task --name "setup" --path ./002_IMPLEMENT/01_api --json
@@ -118,7 +121,7 @@ Run 'fest validate tasks' to verify task files exist in implementation sequences
 			return RunCreateTask(cmd.Context(), opts)
 		},
 	}
-	cmd.Flags().IntVar(&opts.After, "after", 0, "Insert after this number (0 inserts at beginning)")
+	cmd.Flags().IntVar(&opts.After, "after", -1, "Insert after this task number (-1 or omit to append at end; 0 inserts at beginning)")
 	cmd.Flags().StringSliceVar(&opts.Names, "name", nil, "Task name(s) - can be specified multiple times for batch creation")
 	cmd.Flags().StringVar(&opts.Path, "path", ".", "Path to sequence directory (directory containing numbered task files)")
 	cmd.Flags().StringVar(&opts.VarsFile, "vars-file", "", "JSON vars for rendering")
@@ -129,6 +132,27 @@ Run 'fest validate tasks' to verify task files exist in implementation sequences
 	cmd.Flags().BoolVar(&opts.JSONOutput, "json", false, "Emit JSON output")
 	cmd.Flags().BoolVar(&opts.AgentMode, "agent", false, "Strict mode: require markers, auto-validate, block on errors, JSON output")
 	return cmd
+}
+
+// resolveTaskInsertAfter maps the --after default (-1) to append-at-end by
+// detecting the highest existing task number in the sequence directory,
+// matching create sequence's behavior.
+func resolveTaskInsertAfter(ctx context.Context, sequenceDir string, after int) int {
+	if after != -1 {
+		return after
+	}
+	parser := festival.NewParser()
+	tasks, err := parser.ParseTasks(ctx, sequenceDir)
+	if err != nil || len(tasks) == 0 {
+		return 0
+	}
+	maxNum := 0
+	for _, t := range tasks {
+		if t.Number > maxNum {
+			maxNum = t.Number
+		}
+	}
+	return maxNum
 }
 
 // RunCreateTask executes the create task command logic.
@@ -182,9 +206,12 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 	mgr := tpl.NewManager()
 	loader := tpl.NewLoader()
 
+	opts.After = resolveTaskInsertAfter(ctx, absPath, opts.After)
+
 	// Track all created tasks for output
 	var createdTasks []map[string]any
 	var createdPaths []string
+	var renumbered []string
 	var totalMarkersFilled, totalMarkersCount int
 	currentAfter := opts.After
 
@@ -199,6 +226,12 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 		ren := festival.NewRenumberer(festival.RenumberOptions{AutoApprove: true, Quiet: true})
 		if err := ren.InsertTask(ctx, absPath, currentAfter, name); err != nil {
 			return emitCreateTaskError(opts, errors.Wrap(err, "inserting task").WithField("name", name))
+		}
+		for _, ch := range ren.Changes() {
+			if ch.Type == festival.ChangeRename {
+				renumbered = append(renumbered,
+					fmt.Sprintf("%s -> %s", filepath.Base(ch.OldPath), filepath.Base(ch.NewPath)))
+			}
 		}
 
 		// Compute new task id
@@ -372,7 +405,7 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 			OK:              true,
 			Action:          "create_task",
 			Created:         createdPaths,
-			Renumber:        []string{},
+			Renumber:        renumbered,
 			MarkersFilled:   totalMarkersFilled,
 			MarkersTotal:    totalMarkersCount,
 			MarkersUnfilled: remainingMarkers,
@@ -410,6 +443,12 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 		display.Success("Created %d tasks:", len(createdTasks))
 		for _, task := range createdTasks {
 			display.Info("  └── %s", task["id"])
+		}
+	}
+	if len(renumbered) > 0 {
+		display.Warning("Renumbered %d existing task(s):", len(renumbered))
+		for _, r := range renumbered {
+			display.Info("  └── %s", r)
 		}
 	}
 
--- a/internal/commands/status/handlers_festival.go
+++ b/internal/commands/status/handlers_festival.go
@@ -478,9 +478,20 @@ func AutoCommitStatusChange(ctx context.Context, festivalName, festivalID, oldSt
 		}
 	}
 
-	// Stage only the paths affected by the status change (lock-aware with retry).
-	if err := commitkit.StageFiles(ctx, ws.Root, changedPaths...); err != nil {
-		return "", errors.Wrap(err, "stage")
+	// Stage each affected path independently (lock-aware with retry): a
+	// pathspec matching nothing (e.g. a never-committed pre-move festival
+	// directory) must not sink staging for the paths that do exist.
+	var stageFailures []string
+	staged := 0
+	for _, p := range changedPaths {
+		if err := commitkit.StageFiles(ctx, ws.Root, p); err != nil {
+			stageFailures = append(stageFailures, fmt.Sprintf("%s (%v)", p, err))
+			continue
+		}
+		staged++
+	}
+	if staged == 0 && len(stageFailures) > 0 {
+		return "", errors.New("stage: no paths could be staged: " + strings.Join(stageFailures, "; "))
 	}
 
 	// Check if anything is actually staged.
--- a/internal/commands/status/handlers_festival.go
+++ b/internal/commands/status/handlers_festival.go
@@ -478,20 +478,18 @@ func AutoCommitStatusChange(ctx context.Context, festivalName, festivalID, oldSt
 		}
 	}
 
-	// Stage each affected path independently (lock-aware with retry): a
-	// pathspec matching nothing (e.g. a never-committed pre-move festival
-	// directory) must not sink staging for the paths that do exist.
-	var stageFailures []string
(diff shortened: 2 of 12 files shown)


internal/commands/status/handlers_festival.go
```diff
@@ -478,9 +478,20 @@ func AutoCommitStatusChange(ctx context.Context, festivalName, festivalID, oldSt
...
-	if err := commitkit.StageFiles(ctx, ws.Root, changedPaths...); err != nil {
-		return "", errors.Wrap(err, "stage")
+	// Stage each affected path independently (lock-aware with retry): a
+	// pathspec matching nothing (e.g. a never-committed pre-move festival
+	// directory) must not sink staging for the paths that do exist.
+	var stageFailures []string
+	staged := 0
+	for _, p := range changedPaths {
+		if err := commitkit.StageFiles(ctx, ws.Root, p); err != nil {
+			stageFailures = append(stageFailures, fmt.Sprintf("%s (%v)", p, err))
```
obey-agent: This loop treats every per-path `StageFiles` error as skippable and only fails when `staged == 0`, which masks genuine staging failures. `git.executeStage` runs `git add -- <path>`; the only benign error is "pathspec did not match any files" (the absent pre-move directory). If `newPath` (the moved festival) or another legitimately-changed path fails to stage for any other reason while the events/nav path stages, `staged >= 1`, no error is returned, and the commit proceeds without the festival content — a silent, hard-to-detect partial lifecycle commit in production. Tolerate only non-existent paths and propagate real errors, e.g.: `if _, statErr := os.Stat(p); os.IsNotExist(statErr) { continue }` before the `StageFiles` call, then `return "", errors.Wrap(err, "stage").WithField("path", p)` on a real failure. (`os` is already imported.) Add a test for "one existing path fails to stage, another succeeds -> error, not a partial commit."

internal/commands/festival/create_task.go
```diff
@@ -131,6 +134,27 @@ Run 'fest validate tasks' to verify task files exist in implementation sequences
...
+// resolveTaskInsertAfter maps the --after default (-1) to append-at-end by
+// detecting the highest existing task number in the sequence directory,
+// matching create sequence's behavior.
+func resolveTaskInsertAfter(ctx context.Context, sequenceDir string, after int) int {
+	if after != -1 {
+		return after
+	}
+	parser := festival.NewParser()
+	tasks, err := parser.ParseTasks(ctx, sequenceDir)
+	if err != nil || len(tasks) == 0 {
```
obey-agent: `if err != nil || len(tasks) == 0 { return 0 }` conflates a `ParseTasks` error with an empty directory, defaulting a parse failure to prepend (`after=0`). It's largely self-correcting because `InsertTask` re-parses and would resurface a persistent error, but a transient parse error here would silently prepend into a non-empty sequence — the exact reversed-order failure F1 fixes. Prefer returning the error (change the signature) or, at minimum, only fall back to `0` when the directory is genuinely empty, not when parsing failed.

internal/commands/festival/create_task.go
```diff
@@ -200,6 +227,12 @@ func RunCreateTask(ctx context.Context, opts *CreateTaskOptions) error {
 		if err := ren.InsertTask(ctx, absPath, currentAfter, name); err != nil {
 			return emitCreateTaskError(opts, errors.Wrap(err, "inserting task").WithField("name", name))
 		}
+		for _, ch := range ren.Changes() {
```
obey-agent, line 263: For a batch `--after N` insert into an existing sequence, each iteration resets `r.changes` and this collector appends after every `InsertTask`, so the same original task can be reported as renamed in successive iterations (intermediate names), making `renumbered` list more entries than distinct renamed files. Accurate per-step but potentially confusing in the echo and JSON `renumbered` field; consider de-duplicating to net original->final renames, or documenting that entries are per-insert steps.

obey-agent (commented):
## Staff Review — Verdict: REQUEST CHANGES

**Overview**
Three fixes from the planning-loop UX audit: (F1) `fest create task --after` default flips from `0` (prepend) to `-1` (append at end), auto-detected from the highest existing task number, plus renumber reporting via a new `Renumberer.Changes()`; (F4) `promote`/status-change auto-commit stages each path independently so a never-committed pre-move directory no longer sinks the whole commit; (F8) create subcommands attach a cwd-model hint to scoping-flag errors.

On the historical regression class (renumber leaving PHASE_GOAL `fest_id`/`order` and SEQUENCE_GOAL `fest_parent` stale): **this PR does NOT reintroduce it.** That class concerns phase/sequence GOAL files; this PR touches only task creation and doesn't alter `InsertTask`/`executeChanges` (still pure `os.Rename`). Better still, the append-at-end default produces ZERO renames for the common single-task case, strictly reducing renumbering exposure vs the old prepend default. (Latent pre-existing note: task-file frontmatter `number`/`task_id` still go stale on the `--after 0`/`--after N` renumber paths since `executeChanges` renames without rewriting frontmatter; parser keys off filename so nothing structural breaks — separate follow-up.)

**Blocking finding**
- **(major) F4 auto-commit masks genuine staging failures — `handlers_festival.go:488`.** The new loop treats *every* per-path `StageFiles` error as skippable and only fails when `staged == 0`. `commitkit.StageFiles` -> `git add -- <path>`; the ONLY benign error is "pathspec did not match any files" (the absent pre-move dir). Any *other* failure on a legitimately-changed path — most importantly `newPath`, the moved festival itself (race deletion, real git error) — is swallowed as long as one other path stages. Result: `HasStagedChanges` is true, the commit proceeds, returns a hash, no warning — and the lifecycle transition lands in a commit missing the actual festival content. Silent partial commit in a production on-disk lifecycle path. Tolerate only non-existent paths (`os.Stat` + `os.IsNotExist` -> `continue`) and propagate real errors; add a test for "one existing path fails, another succeeds -> error, not a partial commit."

**Minor (inline)**
- `create_task.go:146` — `if err != nil || len(tasks) == 0 { return 0 }` conflates a parse error with an empty directory, defaulting a parse failure to prepend (the reversed-order bug F1 fixes).
- `create_task.go:230` — batch `--after N` renumber reporting can double-count the same file across iterations.

**What's Done Well**
Correctly mirrors `create sequence`'s append semantics (`--after 0`/`N` preserved, documented); `Changes()` returns a defensive copy; the F4 regression test asserts commit *contents* (moved `FESTIVAL_GOAL.md` in `git log --name-only`) and the all-paths-fail test names every failed path; context propagated with `ctx.Err()` checks; internal `errors` package (no `fmt.Errorf`); F8 hint closure guarded against hijacking a real flag error.

**Staff Standard**
Structurally sound and a net improvement over baseline. The gating issue: F4 achieves resilience by tolerating *all* staging errors rather than only "path does not exist," trading one silent-failure mode (whole commit skipped) for another (silent partial commit). Tighten to existence-only + add the test, and this is a clean merge.

lancekrogers:
All three findings addressed in c462279.

**Major (silent partial commit):** exactly right, the loop traded whole-commit-skipped for partial-commit-proceeds. Staging now tolerates only the benign case: a path that no longer exists anywhere (`os.IsNotExist` after the failed add), which covers the never-committed pre-move directory without skipping tracked deletions (those stage successfully and never hit the error path). Any staging error on an existing path propagates immediately with the failing path named, and no commit happens. New tests: `TestAutoCommitStatusChange_RealStageFailurePropagates` (unreadable-file injection on an existing path alongside a healthy path asserts error + zero lifecycle commits; skipped under root) and `TestAutoCommitStatusChange_MissingOnlyPathsIsNoOp` (all paths absent is a clean no-op, empty hash, no error).

**Parse-error prepend:** `resolveTaskInsertAfter` now returns `(int, error)`; a `ParseTasks` failure fails the command loudly instead of silently prepending. Test drives the unreadable-dir case.

**Batch double-count:** per-step rename pairs are collapsed by `netRenames` into net original -> final entries (chain-aware, sorted by origin for stable output); both the human echo and the JSON `renumbered` field use the collapsed list. Table-driven tests include the exact two-insert batch shape from the finding.

`just test unit` (2333), `just lint all` clean at c462279.

internal/commands/status/handlers_festival.go
lancekrogers: Fixed in c462279. The staging loop now tolerates only truly-absent paths — `if _, statErr := os.Stat(p); os.IsNotExist(statErr) { continue }` — and any other `StageFiles` error is propagated (`return "", errors.Wrap(err, "stage").WithField("path", p)`) rather than swallowed. So a real failure on a legitimately-changed path, including `newPath` (the moved festival itself), aborts the commit instead of producing a silent partial commit. Added `TestAutoCommitStatusChange_RealStageFailurePropagates` (one existing path fails, another succeeds → error, no partial commit); the missing-only case is now an explicit no-op (`TestAutoCommitStatusChange_MissingOnlyPathsIsNoOp`).

internal/commands/festival/create_task.go
lancekrogers: Fixed in c462279. `detectLastTaskNumber` now returns `(int, error)` and wraps a parse failure (`return 0, errors.Wrap(err, "detecting last task number")`) instead of conflating it with an empty directory; the caller surfaces it via `emitCreateTaskError`, so a parse error no longer silently defaults to prepend (the reversed-order behavior F1 fixes). An empty directory still legitimately yields 0 (append at end).

internal/commands/festival/create_task.go
lancekrogers: Fixed in c462279. Batch `--after N` reporting no longer double-counts a file across iterations: `netRenames` collapses per-step rename chains to net original→final entries (sorted by origin), and both the echo and the JSON `renumbered` field use the collapsed list. Table-driven tests cover the two-insert batch shape.

obey-agent (commented):

The current head `c462279` fixes the previous auto-commit blocker. `AutoCommitStatusChange` now only tolerates paths that no longer exist and propagates staging errors for paths that do exist, which avoids the silent partial-commit failure mode from the earlier version.

Key findings:
- No blocking findings.
- Missing pre-move paths remain benign, but real staging failures now return an error and prevent a lifecycle commit from landing without the moved festival content.
- The create-task follow-ups are also sound: parse failures while resolving append position now fail instead of falling back to prepend, and batch renumber reporting is collapsed to net rename chains.

Verification:
- Reviewed changed files in a clean detached worktree at `c462279f687083bc8113da6a02db815e6b34c485`.
- `go test ./internal/commands/festival ./internal/commands/status ./internal/commands` passed.

Staff standard: yes. The PR is now a clean merge candidate for the planning-loop UX and status auto-commit fixes.

lancekrogers:
Applied the one non-blocking finding from the review rather than deferring it: the latent frontmatter staleness on renumber paths is fixed in 2b71bd6.

`executeChanges` now refreshes metadata after every rename, and not just for the task case the review named: renamed task files get `fest_id`/`fest_order` from their new filename, and renamed sequence/phase directories get their goal file's `fest_id`/`fest_order` refreshed plus `fest_parent` on immediate children (tasks under a sequence, sequence goals under a phase), closing the whole staleness class including the long-standing directory-rename variant.

Implementation note: the rewrite is line-surgical inside the frontmatter block (only values of keys already present are replaced) because a `frontmatter.Parse` -> `Replace` round-trip would drop unknown keys, since the schema struct has no catch-all. Files without frontmatter are untouched, modes preserved, and a refresh failure warns without aborting the rename batch (filenames stay the source of truth).

Tests: task rename refresh with custom-key and body byte-preservation, sequence rename refreshing goal + child `fest_parent` while leaving the child's own id and the goal's phase parent untouched, and the no-frontmatter no-op. `just test unit` (2336), `just test integration` (59), `just lint all` clean at 2b71bd6.

lancekrogers:
Follow-up to the frontmatter-staleness fix, correcting my own design call after discussion: 2b71bd6 worked around the frontmatter package's lossy round-trip by writing a line-surgical editor in the festival package, which was a parallel implementation of something the frontmatter package should own.

Now fixed at the root in the latest commit:

- **The actual defect:** the `Frontmatter` schema had no catch-all, so EVERY `Parse` -> mutate -> `Inject` round-trip dropped unknown user keys, and that path runs in production far beyond renumber: task status updates (`progress/manager.go`), goal status changes (`handlers_festival.go:101`), and migrations. The schema gains an inline `Extra map[string]any` (excluded from JSON output) so unknown keys survive all of them; regression test covers parse -> mutate -> inject with custom keys.
- **Ownership:** the surgical rewrite moved into the frontmatter package as exported `UpdateFields` / `UpdateFieldsInFile` (targeted single-field updates preserving formatting, complementing full Parse/Inject rewrites); renumber's `refreshRenamedMetadata` now calls it and keeps only the which-files-which-keys domain logic. No frontmatter editing lives outside the package anymore.

`just test unit` (2338), `just test integration` (59), `just lint all` clean at head.

internal/festival/execute.go
```diff
@@ -73,6 +74,10 @@ func (r *Renumberer) executeChanges() error {
 					WithField("from", change.OldPath).
 					WithField("to", change.NewPath)
 			}
+			if err := refreshRenamedMetadata(change.NewPath); err != nil {
+				fmt.Printf("Warning: renamed %s but could not refresh its frontmatter: %v\n",
```
obey-agent: **[Blocking] Unconditional stdout warning breaks the `--json` output contract.**

This `fmt.Printf` writes to stdout unconditionally, while every other message in `executeChanges` is gated behind `!r.options.Quiet` or `r.options.Verbose`. All three create commands run the renumberer with `Quiet: true` (create_task.go:199, create_sequence.go:180, create_phase.go:254), and `emitCreateTaskJSON` writes a single JSON document to `os.Stdout` (create_task.go:462-465). If `refreshRenamedMetadata` errors mid-batch (permission, concurrent deletion, disk full, the same racy filesystem class this PR chain targets), this line prints ahead of the JSON blob and corrupts any consumer doing `json.Unmarshal(stdout)`.

Fix: send the warning to `os.Stderr` (warnings belong there regardless), or gate it behind `if !r.options.Quiet`. Add a test that Quiet mode with a forced refresh failure still yields parseable stdout.

internal/festival/execute.go
```diff
@@ -160,3 +165,65 @@ func (r *Renumberer) createBackup() error {
...
+	if err != nil || !info.IsDir() {
+		return err
+	}
+
+	var goalFile, number string
+	var childPatterns []string
+	if m := sequenceParsePattern.FindStringSubmatch(base); m != nil {
+		goalFile, number = "SEQUENCE_GOAL.md", strings.TrimLeft(m[1], "0")
+		childPatterns = []string{"*.md"}
+	} else if m := phaseParsePattern.FindStringSubmatch(base); m != nil {
```
obey-agent, line 195: **[Blocking] Phase-rename branch has no test coverage, unlike its sequence sibling.**

`renumber_frontmatter_test.go` covers `TestRenumber_TaskRenameRefreshesFrontmatter` and `TestRenumber_SequenceRenameRefreshesGoalAndChildren`, but there is no phase-rename test. This branch uses different constants from the sequence branch (`phaseParsePattern`, `PHASE_GOAL.md`, child glob `*/SEQUENCE_GOAL.md`) and is reached by real entry points (`fest create phase --after`, `fest renumber phases`). The `InsertPhase`/`RenumberPhases` tests in `renumber_test.go` pass without exercising this path because their fixtures have no real `---` frontmatter, so `UpdateFieldsInFile` is a no-op. The function docstring explicitly claims phase parity, so this on-disk state-safety path is asserted-but-untested.

Fix: add a phase-rename test mirroring the sequence one, asserting PHASE_GOAL.md `fest_id`/`fest_order` refresh plus child SEQUENCE_GOAL.md `fest_parent` refresh, on a real `InsertPhase` or `RenumberPhases` call with frontmatter fixtures.

obey-agent (commented):
## Staff Review: fest #230 (re-review at head 6379519)

**Verdict: Request Changes.**

### Prior findings: resolved
The three findings from the earlier reviews are genuinely fixed and stable at the current head, reconfirmed against the code rather than trusted from the reply threads:
- Auto-commit no longer swallows real staging errors (os.Stat/IsNotExist gate; handlers_festival.go untouched since the fix).
- `resolveTaskInsertAfter` surfaces `ParseTasks` failures instead of silently prepending.
- Batch `--after N` no longer double-counts renames (net origin-to-final collapsing).

### New unreviewed scope
The head advanced twice since the last "merge-ready" review. Commit 2b71bd64 added frontmatter-refresh-on-rename, and 6379519 followed up by moving the field writer into the `frontmatter` package (exported `UpdateFields`/`UpdateFieldsInFile`) plus an inline `Extra` map so unknown keys survive round-trips. The unknown-key fix is good work that addresses a real defect affecting status updates and migrations, not just renumber. Two issues in this new scope block merge:

1. **Output-contract violation (execute.go:78).** An unconditional `fmt.Printf` warning on refresh failure writes to stdout inside a `Quiet`-mode renumber that the three `--json`-emitting create commands rely on. On a refresh error it corrupts the JSON document. Send it to stderr or gate on `!Quiet`. Inline comment posted.

2. **Untested phase-rename branch (execute.go:195).** The phase orchestration (PHASE_GOAL.md refresh plus `*/SEQUENCE_GOAL.md` child fest_parent refresh) has no test, while the sibling sequence branch is well covered and the docstring claims parity. The existing phase renumber tests do not assert frontmatter because their fixtures lack it. Add a phase-rename test mirroring the sequence one. Inline comment posted.

### What's done well
- fest_id/fest_order/fest_parent semantics verified against the real schema rather than assumed.
- Line-surgical rewrite deliberately preserves unknown keys and formatting, and that is tested (`custom_key: keep-me` survives).
- Parent-reference scoping is correct per level (a phase rename does not touch grandchild task fest_parent).
- Refresh failures warn and never abort the load-bearing rename, a sound and explicit tradeoff (aside from the stdout target in item 1).
- Build, vet, gofmt, and the full festival and commands test run are clean at head.

### Staff standard
Not yet. The core design is sound and the prior findings are genuinely resolved, but I would not ship this exact head as my own work until the stdout leak is closed and the phase-rename branch gets the same test treatment the sequence branch already has. Both are small, contained follow-ups.

obey-agent (changes requested):
## Staff Re-Review - Verdict: REQUEST CHANGES

The current head `6379519` still needs the two fixes from the current-head review.

Blocking findings:
- `internal/festival/execute.go` still writes the frontmatter-refresh warning with `fmt.Printf` during a `Quiet` renumber path. The create commands that depend on quiet renumbering also emit JSON; on a refresh failure this warning goes to stdout and corrupts the JSON stream. Send it to stderr or suppress it when `Quiet` is set.
- The phase-rename branch in `refreshRenamedMetadata` is still untested. Sequence renames have coverage for goal and child `fest_parent` refresh, but the analogous phase path (`PHASE_GOAL.md` plus child `*/SEQUENCE_GOAL.md`) does not. Add a phase rename regression test so the documented parity is locked in.

Verification performed:
- `git diff --check ff900049191cf1c51eabaed264321353ca7a83dd...HEAD`
- Reviewed `internal/festival/execute.go` and `internal/festival/renumber_frontmatter_test.go`
- `go test ./internal/festival ./internal/frontmatter ./internal/commands/festival ./internal/commands/status ./internal/commands`

The underlying frontmatter writer extraction and unknown-key preservation are good; these are contained follow-ups before merge.

internal/festival/execute.go
lancekrogers: Fixed in c9e5d1e. The frontmatter-refresh failure warning now goes to `os.Stderr` (`fmt.Fprintf(os.Stderr, ...)`) instead of stdout, so the quiet renumber path used by the JSON-emitting create commands no longer risks corrupting the stdout JSON stream. Routed to stderr rather than suppressed under `Quiet` so a genuine "renamed but frontmatter is now stale" data-integrity warning still surfaces instead of vanishing in JSON mode.

internal/festival/execute.go
lancekrogers: Fixed in c9e5d1e. Added `TestRenumber_PhaseRenameRefreshesGoalAndChildren`: it builds `001_ALPHA/PHASE_GOAL.md` with a child `01_seq/SEQUENCE_GOAL.md`, runs `InsertPhase(..., 0, ...)` to shift the phase to `002_ALPHA`, and asserts the `PHASE_GOAL.md` `fest_id`/`fest_order` refresh (and preserved `fest_parent`) plus the child `*/SEQUENCE_GOAL.md` `fest_parent` refresh to the new phase id — while the child's own `fest_id`/`fest_order` stay untouched. Mirrors the sequence-rename coverage so the documented phase parity is locked in.

obey-agent (approved):
## Staff Re-Review - Verdict: APPROVE

No blocking findings on the updated head `c9e5d1e`.

The two prior request-changes items are resolved:
- The frontmatter-refresh warning now writes to stderr with `fmt.Fprintf(os.Stderr, ...)`, so quiet renumbering can still surface the data-integrity warning without corrupting JSON stdout from the create commands.
- `TestRenumber_PhaseRenameRefreshesGoalAndChildren` now covers the phase branch: `PHASE_GOAL.md` gets refreshed to the new phase id/order, child `*/SEQUENCE_GOAL.md` gets its `fest_parent` updated, and the child sequence's own id/order stay untouched.

I also rechecked the broader changed surface: append-by-default task creation, net rename reporting, scoped flag hints, status auto-commit staging behavior, frontmatter unknown-key preservation, and the shared `frontmatter.UpdateFields` implementation. I did not find a remaining correctness or contract blocker.

Verification performed:
- `git diff --check ff900049191cf1c51eabaed264321353ca7a83dd...HEAD`
- `go test ./internal/festival ./internal/frontmatter ./internal/commands/festival ./internal/commands/status ./internal/commands`
- `just lint all`
- `just test unit` (90 packages, 2339/2339 tests passed)

Staff standard: yes. I would be comfortable submitting this as my own work now; the previous blockers are closed and the new frontmatter behavior has appropriate regression coverage.

PR #231, opened 2026-07-02
fix(security): verified install.sh, default-off marketplace extensions (SEC-01/02)

lancekrogers:
## Summary

Findings SEC-01 (HIGH) and SEC-02 (MEDIUM) from `workflow/code_reviews/fable-2026-07-01-code-review/fest/findings.md`.

### SEC-01 — install.sh: personal-repo, unverified, undocumented curl-pipe installer

**Correction from an earlier version of this PR:** the first fix repointed `install.sh`'s downloader at `Obedience-Corp/fest`, but that's the wrong distribution model. fest is not released as a standalone per-repo binary at all — `Obedience-Corp/fest`'s GitHub releases carry notes only, no binary assets (`gh release view v0.4.6 --repo Obedience-Corp/fest --json assets` is empty; `.justfiles/release.just`'s `dev`/`rc`/`stable`/`promote` recipes only call `gh release create --notes-file`).

camp and fest actually ship bundled through the **Festival packaging repo** (`Obedience-Corp/festival`), which is the real canonical distributor: it publishes real checksummed release assets (`checksums.txt` plus macOS/linux/deb/rpm/apk bundles), has its own `curl -fsSL` install.sh, and the primary install path is `brew install --cask Obedience-Corp/tap/festival` (verified against the tap's `Casks/festival.rb`, generated by GoReleaser).

So the actual fix:
- `install.sh` no longer downloads or installs anything itself. It's a thin, safe redirect: it prints the supported install paths and exits 0.
  - Prebuilt binaries: `brew install --cask Obedience-Corp/tap/festival`, or the Festival installer directly (`curl -fsSL https://raw.githubusercontent.com/Obedience-Corp/festival/main/install.sh | bash`).
  - From source (fest only): `go install github.com/Obedience-Corp/fest/cmd/fest@latest`.
- No curl-download of a fest-only binary that doesn't exist, no checksum-URL invention, no `sudo mv`, no personal-repo pin.
- README's Installation section now leads with the festival packaging repo as the canonical distributor, documents fest-only source installs as the secondary path, and notes that `install.sh` itself only prints guidance.

### SEC-02 — marketplace extension source defaults ON, load errors swallowed

- `internal/features/features.go`: `MarketplaceExtensionSource` now defaults **off** (`envFlag(..., false)`); opt in with `FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=1`. Extensions are agent-guiding methodology content (a prompt-injection surface for agent-driven work), so loading `~/.obey/fest/marketplace/extensions` should be opt-in, not opt-out.
- `internal/extensions/loader.go`: `loadFromDirectory` no longer silently swallows every failure. It now logs to stderr when a directory exists but can't be read (permission errors, not just "doesn't exist"), when an individual extension fails to load, and when one source's extension shadows an already-loaded extension of the same name from a different source.

### SEC-03 — noted, not fixed (out of scope)

`github/downloader.go` pulls by mutable git ref with no integrity pinning for `fest sync`. A full fix is a larger design change (pin to resolved commit SHAs, verify signed tags/content); left as a follow-up rather than over-scoping this PR.

## Test plan

- [x] `just lint all` — 0 issues (exit 0)
- [x] `just lint vet` — clean (exit 0)
- [x] `just test unit` — 91 packages, 2326/2326 tests passed (exit 0)
- [x] Added `internal/features/features_test.go`: default-off when unset, explicit enable (`1/true/on/yes`), explicit disable (`0/false/off/no`), unknown capability.
- [x] Added `TestLoadAll_MarketplaceDefaultOff` in `internal/extensions/loader_test.go` alongside the existing explicit on/off tests.
- [x] Ran the corrected `install.sh` locally — prints the install paths, exits 0, downloads/installs nothing.

### DIFF
--- a/internal/features/features.go
+++ b/internal/features/features.go
@@ -37,9 +37,11 @@ func envFlag(key string, def bool) bool {
 		return def
 	}
 	switch strings.ToLower(v) {
+	case "1", "true", "on", "yes":
+		return true
 	case "0", "false", "off", "no":
 		return false
 	default:
-		return true
+		return def
 	}
 }
--- a/internal/features/features.go
+++ b/internal/features/features.go
@@ -19,13 +19,13 @@ func Supported() []string {
 }
 
 // Enabled reports whether a capability is active. The marketplace extension
-// source defaults on (a missing marketplace dir is a harmless no-op for users
-// who never run the installer) and can be turned off with
-// FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=0.
+// source defaults off (extensions are agent-guiding methodology content, so
+// loading them is opt-in) and can be turned on with
+// FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=1.
 func Enabled(name string) bool {
 	switch name {
 	case MarketplaceExtensionSource:
-		return envFlag("FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE", true)
+		return envFlag("FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE", false)
 	default:
 		return false
 	}
--- a/README.md
+++ b/README.md
@@ -172,8 +172,21 @@ These festivals span Go, Rust, Python, and web projects - from simple CLI fixes 
 
 ## Installation
 
+fest ships bundled with camp as part of the [Festival packaging repo](https://github.com/Obedience-Corp/festival), which is the canonical, checksum-verified distributor for prebuilt binaries:
+
 ```bash
-# Install from source
+# macOS/Linux, via Homebrew
+brew install --cask Obedience-Corp/tap/festival
+
+# or via the Festival installer directly
+curl -fsSL https://raw.githubusercontent.com/Obedience-Corp/festival/main/install.sh | bash
+```
+
+See the [festival repo](https://github.com/Obedience-Corp/festival#install) for the full package list (npm, deb, rpm, apk, Arch) and published `checksums.txt`.
+
+To install fest on its own from source:
+
+```bash
 git clone https://github.com/Obedience-Corp/fest
 cd fest
 just install
@@ -185,15 +198,7 @@ Or with Go:
 go install github.com/Obedience-Corp/fest/cmd/fest@latest
 ```
 
-### install.sh
-
-`install.sh` downloads a prebuilt binary from the [Obedience-Corp/fest releases](https://github.com/Obedience-Corp/fest/releases), verifies its sha256 checksum against the release's published `checksums.txt`, and installs it to `/usr/local/bin/fest` (using `sudo` if that directory isn't writable):
-
-```bash
-curl -fsSL https://raw.githubusercontent.com/Obedience-Corp/fest/main/install.sh | bash
-```
-
-It refuses to install anything it cannot verify: if a release has no matching binary archive or no published `checksums.txt`, it fails with an error instead of installing an unverified binary. Prebuilt binaries are not yet published on every release; if the script fails for that reason, use `go install` or build from source above.
+This repo's `install.sh` does not download or install a binary itself; running it prints the install paths above.
 
 ## Shell Integration (Recommended)
 
--- a/README.md
+++ b/README.md
@@ -185,6 +185,16 @@ Or with Go:
 go install github.com/Obedience-Corp/fest/cmd/fest@latest
 ```
 
+### install.sh
+
+`install.sh` downloads a prebuilt binary from the [Obedience-Corp/fest releases](https://github.com/Obedience-Corp/fest/releases), verifies its sha256 checksum against the release's published `checksums.txt`, and installs it to `/usr/local/bin/fest` (using `sudo` if that directory isn't writable):
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/Obedience-Corp/fest/main/install.sh | bash
+```
+
+It refuses to install anything it cannot verify: if a release has no matching binary archive or no published `checksums.txt`, it fails with an error instead of installing an unverified binary. Prebuilt binaries are not yet published on every release; if the script fails for that reason, use `go install` or build from source above.
+
 ## Shell Integration (Recommended)
 
 Add to your shell config for quick navigation commands:
--- a/install.sh
+++ b/install.sh
@@ -1,116 +1,26 @@
 #!/bin/bash
-# Installation script for fest CLI
+# fest is not distributed as a standalone per-repo release binary; it ships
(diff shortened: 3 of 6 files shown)


obey-agent (commented):

**Overview**
This PR resolves the two targeted security issues cleanly. The installer no longer pretends to install an unverifiable fest-only binary, the README points users to the actual Festival distribution surface, and marketplace extension loading is now explicit opt-in.

**Key Findings**
- No blocking findings. I did not find a correctness, security, or maintainability issue that warrants an inline comment.

**What's Done Well**
- The `install.sh` change favors a boring, accurate redirect over inventing a standalone release path that the repo does not actually publish.
- The feature-flag tests cover default-off, explicit enable, explicit disable, and unknown capability behavior.
- Extension loader diagnostics improve operator visibility without changing the existing load-and-continue contract.

**Verification**
- `go test ./internal/features ./internal/extensions`
- `./install.sh` exits 0 and prints install guidance without downloading or installing anything.

**Staff Standard**
Yes. I would be comfortable submitting this as my own work for the scoped SEC-01/SEC-02 fixes. SEC-03 remains correctly called out as a separate integrity-pinning design problem rather than being half-fixed here.

internal/features/features.go
```diff
@@ -19,13 +19,13 @@ func Supported() []string {
...
-// who never run the installer) and can be turned off with
-// FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=0.
+// source defaults off (extensions are agent-guiding methodology content, so
+// loading them is opt-in) and can be turned on with
+// FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=1.
 func Enabled(name string) bool {
 	switch name {
 	case MarketplaceExtensionSource:
-		return envFlag("FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE", true)
+		return envFlag("FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE", false)
```
obey-agent, line 28: For this security-sensitive opt-in, `envFlag` treats *any* non-empty value not in `{0,false,off,no}` as enabled (the fall-through `default: return true`). So a user trying to disable with a non-canonical word — e.g. `FEST_FEATURE_MARKETPLACE_EXTENSION_SOURCE=disabled`, `=off!`, or a typo like `=of` — would silently *enable* the marketplace source. The unset default is correctly off, so this isn't a blocker, but for a fail-safe security control consider an affirmative allowlist (only `1/true/on/yes` enable; everything else stays off), and assert a garbage value stays off in `features_test.go`.

obey-agent (commented):

**Overview**
Two-part security PR. SEC-01 rewrites `install.sh` from a `curl -L | sudo mv` bootstrap pointed at a personal repo into a thin print-and-exit script that defers to the checksum-verified Festival packaging repo. SEC-02 flips marketplace extensions from default-on to default-off and surfaces previously-swallowed load errors. I traced both paths adversarially; both fixes are correct and fail-safe.

**Key Findings**
- SEC-01 is the strongest possible fix: rather than bolting verification onto a `curl|bash|sudo mv` path, it *removes the download/execute path entirely*. The new script is `set -euo pipefail`, a single-quoted heredoc (no variable expansion, no injection surface), no `curl`/`sudo`/`eval`/temp files, exits 0. Confirmed no build tooling depends on it (`just install` builds from source); the prior script was already broken (personal repo, no valid assets), so the no-op isn't a regression.
- SEC-02 is genuinely off at the only entry point: `MarketplaceExtensionsDir()` is loaded solely inside the `if features.Enabled(...)` block in `LoadAll`; grep confirms no ungated caller. Extension loading is pure manifest/data parsing (no code execution on load), so default-off neutralizes the injected-content surface. Tests cover default-off in both `features_test.go` and `loader_test.go`.
- (minor, inline) `envFlag` fall-through defaults unrecognized values to *enabled*, so a non-canonical disable value would silently turn a security-sensitive flag on.

**What's Done Well**
Self-corrected the first attempt to the actually-correct distributor; README leads with the checksum-verified cask path; the new cross-source shadow warning is a useful security diagnostic; the split is right — fail-open for availability (keep loading other extensions), fail-closed for the security default (source off). `fmt.Fprintf(os.Stderr, ...)` is logging, not error construction, so no error-package violation.

**Staff Standard**
Correct threat model, fail-safe default, no bypass found, behavioral tests for both security paths, no compat regression. Merge-ready; the one item below is minor hardening. Leaving the Approve for Lance.

obey-agent (commented):

The current head `70aaf5f` is still merge-ready. The installer is a safe print-and-exit redirect to the actual Festival distribution surface, and marketplace extensions remain default-off with explicit opt-in.

Key findings:
- No blocking findings.
- `install.sh` does not download, execute, move, or sudo anything; it points users to the checksum-verified Festival package path or source install.
- Marketplace extensions are gated at the only `LoadAll` entry point, and loader diagnostics preserve the load-and-continue behavior while surfacing unreadable/invalid/shadowed extension cases.

Verification:
- Reviewed changed files in a clean detached worktree at `70aaf5ff8252bb08536b054dc17943c0b67b74fc`.
- `go test ./internal/features ./internal/extensions` passed.
- `./install.sh` exits 0 and prints guidance only.

Staff standard: yes. This is the right scoped fix for SEC-01/SEC-02, with SEC-03 correctly left as a separate integrity-pinning design problem.

internal/features/features.go
lancekrogers: Fixed in de93804. `envFlag` now switches on explicit truthy tokens (`1`/`true`/`on`/`yes`) → true and explicit falsy tokens (`0`/`false`/`off`/`no`) → false, and any unrecognized value falls back to the caller's `def` instead of enabling. For this security-sensitive default-off flag a non-canonical value (typo, `disabled`, etc.) now stays off. Added `TestEnabled_MarketplaceExtensionSourceUnrecognizedValueStaysOff` asserting unrecognized values don't flip it on.

obey-agent (approved):
## Staff Re-Review - Verdict: APPROVE

No blocking findings on the updated head `de93804`.

The new commit resolves the remaining feature-flag hardening note: `envFlag` now only enables on recognized truthy tokens (`1`, `true`, `on`, `yes`), disables on recognized falsy tokens, and falls back to the caller's default for unrecognized values. For the marketplace extension source that means typos like `disabled`, `maybe`, or `2` stay off instead of enabling a security-sensitive default-off content source.

Verification performed:
- `git diff --check ff900049191cf1c51eabaed264321353ca7a83dd...HEAD`
- `go test ./internal/features ./internal/extensions`
- `./install.sh` exits 0 and prints guidance only

PR #232, opened 2026-07-02
feat(commands): add fest plugins command to list discovered plugins

lancekrogers:
## Problem

fest has full git-style plugin dispatch — any `fest-<name>` binary on PATH runs as `fest <name>`, plus a discovery layer covering the active config repo manifest (`plugins/manifest.yml`) and PATH — but none of it is surfaced. There is no way for users or agents to find out what plugins are installed; `camp` has had `camp plugins` for this, fest had nothing.

## Change

- New global-scope `fest plugins` command (system group), built directly on the existing `PluginDiscovery` — no new discovery logic.
- Human output: sorted command names with executable paths; manifest entries whose executable is missing show an `(exec not found: fest-<name>)` marker; manifest summaries render as a dim second line (the synthetic `Plugin: <exec>` default summary is suppressed).
- `--json` for agents: `{"plugins": [...], "count": N}` with snake_case keys, matching the `fest extension list` shape. `DiscoveredPlugin.Source/Path` gain json tags (previously untagged; the type was never JSON-marshaled before, so no consumer sees a casing change). Empty result emits `"plugins": []`, not `null`.
- Command name is `plugins` to match camp exactly (confirmed with Lance).

## Verification

- `just test unit`: 92 packages, 2370/2370 pass. `just lint all`: 0 issues.
- 5 new tests in `internal/commands/plugins_list_test.go`: empty state, non-executable skipped, human output, JSON shape/sort/casing, manifest summary + missing-exec marker (isolated via `FEST_CONFIG_DIR` + `PATH`).
- Live check with a locally built binary and a fake `fest-demo` on PATH:

```
$ fest plugins
Installed Plugins:
  demo                 .../fakebin/fest-demo
$ fest demo
demo        # dispatch unaffected
```

`fest --help` now lists the command under System Commands.

## Follow-up (out of scope)

Inlining an "Installed plugins" section into `fest --help` / `camp --help` output itself (so the plugin names appear without running the listing command) — planned as a small paired change across both CLIs.

### DIFF
--- a/internal/commands/plugin.go
+++ b/internal/commands/plugin.go
@@ -1,13 +1,19 @@
 package commands
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"slices"
+	"sort"
 	"strings"
 
 	festerrors "github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/plugins"
+	"github.com/Obedience-Corp/fest/internal/scope"
+	"github.com/Obedience-Corp/fest/internal/ui"
+	"github.com/spf13/cobra"
 )
 
 var errPluginHandled = errors.New("plugin handled")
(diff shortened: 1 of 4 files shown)


obey-agent (approved):
## Staff Review - Verdict: APPROVE

No blocking findings on head `2fec2bc`.

This is a focused addition of `fest plugins` on top of the existing plugin discovery layer. The command stays global-scope, avoids duplicating discovery logic, sorts output for stable human and JSON consumers, and emits a clean agent-facing shape with `plugins: []` for the empty case. Manifest-only plugins with missing executables are visible with a clear marker, while PATH-only executable plugins still show the resolved path.

What I checked:
- command registration under System Commands and global scope
- human output, manifest summary suppression/default handling, missing-exec marker
- JSON shape and field casing through the embedded `Plugin` tags plus new `source`/`path` tags
- dispatch parity: the listing command uses `PluginDiscovery`, the same source used by dispatch

Verification performed:
- `git diff --check 9f93494cf37123ea9dcc8c937842d7b2bc98ec1b...HEAD`
- `go test ./internal/commands ./internal/plugins`
- `go test ./internal/commands -run 'TestPluginsList' -v`
- `go run ./cmd/fest plugins --json`
- `just lint all`
- `just test unit` (92 packages, 2370/2370 tests passed)

Staff standard: yes. I would be comfortable submitting this as my own work; the implementation is scoped, test-backed, and integrated with the existing plugin model.

PR #234, opened 2026-07-06
feat(promote): colorized, status-ordered tab completion matching fest go

lancekrogers:
## What

Bring `fest promote` tab completion to parity with `fest go` (fgo): **status ordering** (active → ready → planning) and **theme-aware colorization**. Along the way, two structural cleanups requested in review:

1. Completion colors now come from the canonical `ui` theme palette, not a hardcoded ANSI table.
2. The shell-init scripts are extracted from Go string literals into embedded, testable shell files (porting camp's `internal/shell` pattern).
3. The interactive festival picker now colorizes its `[status]` labels from the same palette (affects `fest promote`, `fgo`, and `fest watch`).

## Behavior

- `fest promote <TAB>` orders festivals active → ready → planning (then recency), colorized by status — matching `fgo <TAB>`. Previously it was alphabetical with no color.
- Ordering holds for both empty and typed prefixes, across zsh (colorized widget) and bash/fish/cobra-native (`__complete`).
- `fest promote -<TAB>` still offers flags (`--dungeon/--force/--json/--no-commit`).

## How

**Ordering** — `promotePickerOptions()` uses `OrderByStatusThenRecency`; `OrderedSelectorNames` preserves that order even when fuzzy-filtering a typed prefix.

**Color (parity fix)** — new `ui.StatusColorSequence(status)` derives the raw ANSI escape from the active theme palette (`GetStatusColor`) and is emitted unconditionally so it survives non-TTY completion capture. Both promote and `fest go` completion source from it, so they can't drift. This also fixes a latent bug: the old completion table flattened dungeon/someday/archived/ritual to grey/dim, while the palette colors them distinctly.

**Shell extraction** — `internal/shell/` now holds the bash/zsh and fish scripts as `scripts/*.sh`, embedded via `//go:embed` and served by `shell.Generate`. The zsh promote widget lives in the script file, not a Go heredoc. New tests assert content and run real `bash -n` / `zsh -n` / `fish -n` syntax checks. Generated output is byte-identical to the previous heredocs.

**Picker colorization** — `picker.Item` gains an optional non-matched `Prefix` (with `PrefixColor`) rendered as a colored column via the picker's stderr renderer, so the festival status label is colorized without disturbing fuzzy matching (the festival name stays the matched `Name`). The shared festival picker sets the prefix from `ui.GetStatusColor`, giving `fest promote`/`fgo`/`fest watch` colored `[status]` labels.

## Testing

`go build`, `go vet`, `gofmt`, full `go test ./...`, `golangci-lint run ./...` (0 issues) all pass. Plus: a functional zsh test proving both widget branches, `bash -n`/`zsh -n` syntax tests on the embedded script, a palette-mapping test locking the fixed dungeon/someday/ritual colors, and a byte-diff confirming shell-init output is unchanged.

Deferring final approval/merge to a human reviewer.


internal/commands/shared/completion_color.go
```diff
@@ -0,0 +1,84 @@
...
+func OrderedSelectorNames(candidates []FestivalPickCandidate, toComplete string) []string {
+	selectors := selectorCandidatesFromPickCandidates(candidates, false)
+	if strings.TrimSpace(toComplete) == "" {
+		names := make([]string, 0, len(selectors))
+		for _, c := range selectors {
+			names = append(names, c.Name)
+		}
+		return names
+	}
+	matches := fuzzyMatchSelectorCandidates(toComplete, selectors)
```
obey-agent: This still loses the status ordering when Cobra passes a non-empty completion prefix. `OrderedSelectorNames` preserves the picker order only for an empty `toComplete`; once the user has typed a prefix, `fuzzyMatchSelectorCandidates` sorts by score and then name, so equal matches fall back to alphabetical order instead of active -> ready -> planning.

I reproduced it with active `zzz-feature-ZF0003`, ready `mmm-feature-MF0002`, and planning `aaa-feature-AF0001`: `fest __complete promote ''` returns active/ready/planning, but `fest __complete promote feature` returns planning/ready/active. Bash/fish and Cobra-native completion users hit this path after typing text. Please keep the candidate order after filtering, or add an ordered fuzzy-filter helper, and cover the non-empty `toComplete` case in tests.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This PR is well-scoped and moves the promote completion implementation toward parity with `fgo`: the shared color source is a good extraction, the zsh display path follows the existing `fgo` pattern, and the empty-prefix ordering behavior works. The remaining issue is that the Cobra completion path still falls back to alphabetical ordering once the user has typed a non-empty prefix, which undercuts the headline behavior for bash/fish and native completion users.

**Key Findings:**
- `internal/commands/shared/completion_color.go`: `OrderedSelectorNames` preserves candidate order only when `toComplete` is empty. With a typed prefix, equal fuzzy matches are sorted by name, so `fest __complete promote feature` can return planning before ready before active. I left an inline comment with a concrete reproduction.

**What's Done Well:** The color constants are consolidated cleanly, the navigation completion path delegates to the shared palette, and the new tests cover the core empty-prefix status ordering and color-line format. Full `go test ./...` passes in a detached PR worktree.

**Staff Standard:** I would be comfortable with the direction, but I would not merge this as-is because the non-empty completion-prefix path still violates the PR's stated status-ordering guarantee.

obey-agent (changes requested):
Requesting changes for the non-empty completion-prefix ordering regression called out in the inline comment and review summary. Empty-prefix promote completion is fixed, but typed-prefix completion can still fall back to alphabetical ordering instead of active -> ready -> planning.

obey-agent:
**Review follow-up — addressed in b71c65d**

Both the requested change and an additional zsh regression found in a second review pass are fixed:

1. **Typed-prefix ordering** (the Request Changes finding): `OrderedSelectorNames` now preserves active → ready → planning order when a completion prefix is typed, instead of falling back to fuzzy-score/alphabetical. The inline reproduction (active `zzz-feature-ZF0003`, ready `mmm-feature-MF0002`, planning `aaa-feature-AF0001`) now returns status order for `fest __complete promote feature`; covered by a new test (`TestPromoteCompletionKeepsStatusOrderWithTypedPrefix`).

2. **zsh flag completion** (found separately): the promote widget branch fired on flag-prefixed words, swallowing `fest promote -<TAB>` flag completion. Added a `"${words[CURRENT]}" != -*` guard so flags fall through to the generic `__complete` path. Verified functionally with a stubbed `compadd`: `fest promote <TAB>` → colorized promote branch, `fest promote -<TAB>` → `--dungeon/--force/--json/--no-commit` flags.

Gate at this head: `go build ./...`, `go vet ./...`, `gofmt`, full `go test ./...`, `golangci-lint run ./...` (0 issues), and `zsh -n` all pass. `fest go` completion output is byte-identical across all statuses (no fgo regression).

Deferring the final approval/merge to a human reviewer.

obey-agent (commented):
**Verdict: Approve**

**Overview:** The follow-up commit resolves the prior completion-order regression. `OrderedSelectorNames` now filters typed-prefix fuzzy matches while preserving the existing active -> ready -> planning candidate order, and the zsh branch correctly lets flag-prefixed completion fall through to Cobra's generic flag completion path.

**Key Findings:** None blocking. I did not leave new inline comments because the previously identified issue is fixed and I did not find a replacement concern.

**What's Done Well:** The shared palette extraction removes duplication with `fgo`, the promote picker/completion path now uses one ordering contract, and the regression test covers the exact typed-prefix failure mode from the earlier review.

**Staff Standard:** Yes. I would be comfortable merging this as my own work after the current green local checks.

obey-agent (approved):
Approved on re-review. The typed-prefix status-order regression is fixed, flag-prefix completion still reaches the generic Cobra path, and local test/build/vet/shell syntax checks pass.

obey-agent:
**Review follow-up — two structural fixes (commits 064c07f, 62a99cf)**

1. **Color parity** (colors are stored in a constant): completion colorizing no longer uses a hardcoded ANSI table. It now derives from the canonical theme palette via `ui.StatusColorSequence` (backed by `GetStatusColor`), so promote and `fest go` completions match how fest colors statuses everywhere, and follow the user's dark/light/high-contrast theme. This also fixes a latent divergence where the old table flattened dungeon/someday/archived/ritual to grey/dim.

2. **Shell code in Go strings**: the shell-init scripts are extracted from Go heredocs into embedded, testable shell files (`internal/shell/scripts/*.sh`, `//go:embed`, served by `shell.Generate`), porting camp's `internal/shell` pattern. New tests run real `bash -n` / `zsh -n` / `fish -n` syntax checks. Generated `shell-init` output is byte-identical to before.

Gate at this head: `go build`, `go vet`, `gofmt`, full `go test ./...`, `golangci-lint` (0 issues), plus the functional zsh widget test. Deferring approval/merge to a human.

obey-agent (commented):
**Verdict: Approve**

**Overview:** The latest revision keeps the earlier promote completion fixes and cleanly extends the implementation by sourcing status completion colors from the `ui` palette and moving shell integration out of Go heredocs into embedded shell files. I reviewed the new shell package, embedded bash/zsh and fish scripts, palette bridge, `fest go` completion color path, and promote completion path.

**Key Findings:** None blocking. I did not leave new inline comments because I did not find a correctness, compatibility, or maintainability issue worth calling out on the current diff.

**What's Done Well:** The shell extraction makes the scripts directly testable, the status color source now matches the existing theme palette instead of a parallel hardcoded table, and the prior typed-prefix ordering and flag-completion edge cases remain covered. The new shell syntax tests are the right shape for this change.

**Staff Standard:** Yes. I would be comfortable merging this as my own work after the current green checks.

**Verification:** Focused package tests, full `go test ./...`, `go vet ./...`, `go build ./...`, generated zsh/bash shell syntax checks, promote completion reproduction, color completion output check, and `golangci-lint run ./...` all pass. Fish is not installed in this environment, so the fish syntax command was skipped locally; the unit test has the same skip behavior.

obey-agent (approved):
Approved on latest head 62a99cf. I reviewed the shell extraction and palette-backed completion colors; ordering and flag completion still reproduce correctly, and local test/build/vet/lint/shell checks pass.

PR #246, opened 2026-07-09
fix(link): never silently evict another festival's project link

lancekrogers:
## The bug (root cause of "promote dropped my link")

Campaign-history autopsy of the CF0001 incident shows `fest promote` was innocent: its `UpdateNavigationAfterMove` correctly rewrites `festival_path` on every non-terminal move (verified against commit 1157b76e4 in obey-campaign, and end-to-end in a scratch campaign).

The links actually vanished when a **second festival linked the same project**. `SetLinkWithPath` enforces one-festival-per-project by evicting the previous holder, and the old guard (#201) only blocked when the holder was in `active/`. Planning/ready festivals were evicted silently. In the recorded incident, creating/linking MM0001 against `projects/camp` deleted CF0001's link hours before CF0001's promote; the missing link was then noticed after the promote and misattributed to it.

## The fix

- `fest link` / `fest go link`: relinking a project held by ANY other festival now errors, naming the holder, unless `--force`. Forced takeovers print `Replaced link: project was linked to <festival>` and emit `replaced_festival` in `--json`.
- `fest create festival --project`: never evicts; warns, skips the auto-link, and the summary states the real reason.
- `navigation.SetLinkWithPath` returns the evicted festival name so every caller must handle the takeover explicitly. Same-festival relinks (including promote's path refresh) report no eviction.
- Removed the now-obsolete `isActiveFestivalPath` special case.

## Verification

- `just lint all` 0 issues; `just test unit` 2446/2446; `just docs` regenerated.
- Scratch-campaign end-to-end with the built binary: steal without `--force` blocked with holder named + hint; `--force` succeeds and reports the eviction; `create festival --project` on a held project warns and leaves the link intact; promote planning->ready preserves the link with updated `festival_path` (regression).

### DIFF
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -1903,7 +1903,7 @@ fest go link [path] [flags]
 ### Options
 
 ```
-      --force   overwrite an existing active festival link
+      --force   relink a project already linked to another festival
   -h, --help    help for link
 ```
 
@@ -2691,7 +2691,7 @@ fest link [path] [flags]
 ### Options
 
 ```
-      --force   overwrite an existing active festival link
+      --force   relink a project already linked to another festival
   -h, --help    help for link
       --json    output in JSON format
       --show    show current link
--- a/docs/cli-reference/fest_go_link.md
+++ b/docs/cli-reference/fest_go_link.md
@@ -35,7 +35,7 @@ fest go link [path] [flags]
 ### Options
 
 ```
-      --force   overwrite an existing active festival link
+      --force   relink a project already linked to another festival
   -h, --help    help for link
 ```
 
--- a/docs/cli-reference/fest_link.md
+++ b/docs/cli-reference/fest_link.md
@@ -37,7 +37,7 @@ fest link [path] [flags]
 ### Options
 
 ```
-      --force   overwrite an existing active festival link
+      --force   relink a project already linked to another festival
   -h, --help    help for link
       --json    output in JSON format
       --show    show current link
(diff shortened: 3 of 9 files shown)


obey-agent (commented):
## Staff Review - Merge Ready

**Verdict:** Approve

**Overview:** Correct root-cause fix. The campaign-history autopsy is convincing: promote's link update is sound, and the data loss came from silent one-festival-per-project eviction with an active-only guard. Extending the guard to all holder statuses, making forced takeovers loud (text + `replaced_festival` JSON), and refusing to evict from the create-festival path closes every silent-eviction entry point (`fest link`, `fest go link` both directions, `fest create festival --project`).

**Code quality:** `SetLinkWithPath` returning the evicted name pushes the reporting obligation to every caller via the type system; the same-festival guard (`oldFestival != festivalName`) keeps promote's path refresh eviction-free; `guardProjectConflict` centralizes the policy so the three command sites cannot drift; obsolete `isActiveFestivalPath` removed with its tests replaced by table-driven guard coverage.

**Verification (local, at head 756a8ad):** `just lint all` 0 issues; `just test unit` 2446/2446; `just docs` zero drift. End-to-end in a scratch campaign: block-without-force names the holder with an actionable hint; forced takeover reports the eviction; create-festival skip preserves the holder's link and states the reason in the summary; promote planning->ready keeps the link with `festival_path` updated.

**Follow-up worth tracking (not blocking):** navigation.yaml is whole-file read-modify-write with no cross-process locking; concurrent long-running fest sessions can still last-writer-wins each other. Worth an intent for flock-or-merge hardening.

**Staff standard:** Yes - evidence-first diagnosis, minimal surface, every entry point covered with tests.

PR #247, opened 2026-07-09
fix(navigation): serialize navigation.yaml mutations with flock + atomic writes

lancekrogers:
Follow-up to #246. That PR fixed silent link eviction; this one closes the residual loss mechanism: navigation.yaml was whole-file read-modify-write with no cross-process locking, so a concurrent fest process saving a stale snapshot silently dropped links added in between (last writer wins).

## Changes
- `navigation.Update(ctx, fn)`: takes an exclusive flock on `navigation.yaml.lock`, loads the freshest state inside the lock, applies `fn`, saves. `fn` returning an error aborts without saving.
- Every mutating call site migrated (`fest link`, `fest go link` both directions, `fest unlink`, `fest go map`/`unmap`, `fest create festival` auto-link + rollback, `UpdateNavigationAfterMove` for promote/status moves). No raw Load+Save mutators remain in commands.
- `Save` now writes temp + rename, so readers and crashed writers never observe a partial file, and no temp files are left behind.
- `!unix` builds get a no-op lock (pre-existing last-writer-wins behavior); windows cross-build verified.

## Verification
- New tests: 12 concurrent `Update` writers all preserved (fails as a lost-update on the old code), fn-error aborts save, cancelled context rejected, no temp-file leftovers. Passed with `-race`.
- `just lint all` 0 issues; `just test unit` 2450/2450.
- CLI-level: `fest link` / `fest link --show` work through the lock in a scratch campaign; `.campaign/fest/navigation.yaml.lock` created alongside.

### DIFF
--- a/internal/commands/festival/create_festival.go
+++ b/internal/commands/festival/create_festival.go
@@ -555,7 +555,7 @@ func writeFestYaml(ctx context.Context, cfg *createConfig) (*createResult, error
 	res := &createResult{festConfig: festConfig}
 
 	if cfg.opts.Project != "" {
-		resolveProjectLink(cfg, festConfig, res)
+		resolveProjectLink(ctx, cfg, festConfig, res)
 	}
 
 	if cfg.festivalType != nil {
@@ -578,7 +578,7 @@ func writeFestYaml(ctx context.Context, cfg *createConfig) (*createResult, error
 }
 
 // resolveProjectLink resolves and optionally links a project path for the festival.
-func resolveProjectLink(cfg *createConfig, festConfig *config.FestivalConfig, res *createResult) {
+func resolveProjectLink(ctx context.Context, cfg *createConfig, festConfig *config.FestivalConfig, res *createResult) {
 	workspaceRoot := filepath.Dir(cfg.festivalsRoot)
 	resolved, err := ResolveProjectPath(cfg.opts.Project, workspaceRoot)
 	if err != nil {
@@ -599,8 +599,7 @@ func resolveProjectLink(cfg *createConfig, festConfig *config.FestivalConfig, re
 		return
 	}
 
-	nav, navErr := navigation.LoadNavigation()
-	if navErr == nil {
+	updateErr := navigation.Update(ctx, func(nav *navigation.Navigation) error {
 		// A project holds one festival link at a time. Creating a festival must
 		// never silently evict another festival's link; skip and tell the user.
 		if existing, _, hasConflict := nav.ProjectConflict(cfg.dirName, resolved); hasConflict {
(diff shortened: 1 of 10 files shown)


obey-agent (commented):
## Verdict: **Approve**

## Overview
Follow-up to #246 that closes the residual navigation.yaml lost-update path: whole-file read-modify-write without cross-process locking. `navigation.Update(ctx, fn)` takes an exclusive flock, loads fresh state under the lock, applies `fn`, and saves; `Save` is temp+rename. All listed mutators migrate to `Update`.

## Review notes
- **Correct shape:** load-inside-lock is the right fix for last-writer-wins; concurrent 12-writer test is the proof that matters (and would fail without the lock).
- **Atomic Save:** temp in same dir + rename + cleanup on error; `UpdatedAt` still stamped in `Save` so shortcut path doesn't regress timestamps.
- **Call-site migration:** link / go link / unlink / map-unmap / create festival auto-link+rollback / promote / status move all go through `Update`. Public `Save` remains for the Update path only in practice.
- **Error paths:** fn error aborts without save (tested); cancelled context rejected before lock work.
- **Platform:** `!unix` no-op lock is an honest fallback; windows cross-build called out.

## Non-blocking
- No `fsync` before rename (common for campaign config; accept for this class of file).
- `runUnlink` still does a pre-lock `LoadNavigation` for project→festival discovery; mutation is reloaded under lock. Tiny TOCTOU on discovery only.
- Creating `.campaign/fest/navigation.yaml.lock` beside the yaml is expected; worth a one-line note in operator docs if lock files surprise users.

## Staff Standard
Yes — ship it. This is the correct systems fix for silent link loss under concurrency.

Reviewed against: https://github.com/Obedience-Corp/fest/pull/247 @ `fc957dc`

obey-agent (commented):
## Staff Review - Merge Ready

**Verdict:** Approve

**Overview:** Closes the residual link-loss mechanism after #246: navigation.yaml mutations now go through `navigation.Update(ctx, fn)` - exclusive flock, load-inside-lock, apply, save - so a concurrent fest process can no longer clobber links with a stale snapshot. Save is atomic (temp + rename). Every mutating call site is migrated; grep confirms no raw Load+Save mutators remain in commands.

**Design notes verified in code:** fn-error aborts without saving; interactive pickers run BEFORE the lock is taken (no lock held across prompts); `UpdateNavigationAfterMove` preserves its silent-skip-outside-campaign behavior while now warning if a real save fails; `!unix` no-op lock keeps the windows cross-build intact (verified).

**Verification (local, at head of nav-write-locking):**
- New tests: 12 concurrent writers all preserved (this is a lost-update failure on the old code), fn-error abort, cancelled-context rejection, no temp leftovers - all passing under `-race`
- `just lint all`: 0 issues; `just test unit`: 2450/2450
- CLI: `fest link`/`--show` work through the lock in a scratch campaign; lock file created alongside navigation.yaml

**Staff standard:** Yes - the Update closure makes the load-inside-lock discipline structural rather than conventional.

PR #245, opened 2026-07-09
feat: add approval auto mode guardrails

lancekrogers:
Replaces #238 (closed: opened by obey-agent against campaign identity rules; agent account is review/comment-only). Same branch, now with main merged in and review fixes applied.

## Summary
- Adds explicit `fest workflow approve --auto` with fail-closed judge execution: the judge command is resolved from `--judge-command` or the `hooks.approval_judge.command` hook in `.festival/config.yaml`; if neither is set, `--auto` fails closed. There is **no default judge command**.
- The judge receives JSON on stdin (schema `fest.approval.judge/v1`) and must return an approve/reject decision with a non-empty reason. Missing commands, timeouts, non-zero exits, malformed JSON, unknown decisions, and empty reasons all fail closed.
- `fest init` scaffolds a commented `hooks:` example in `.festival/config.yaml` for discoverability.
- Documents the protocol in `fest understand workflow` and updates agent workflow templates.

## Integration with #241 (decision metadata)
- Approve path unified on `ApproveWithAudit(feedback, decision)`; auto mode records the judge verdict as decision metadata (`decision_actor=agent`, `decision_summary=<judge reason>`) alongside the durable audit text, for both approve and reject outcomes.
- `--auto` is mutually exclusive with `--as`/`--summary` (the judge determines the actor and summary).
- Fixed a stale doc claim that `ob judge` is the default judge command (there is no default; fail-closed).

## Verification
- `just lint all`: 0 issues; `just test unit`: 2428/2428; `just docs` idempotent; `git diff --check origin/main...HEAD` clean.
- End-to-end against a scratch workspace with the dev binary: unconfigured `--auto` fails closed (exit 1, actionable hint); a real judge subprocess reject blocks the step with full audit + decision metadata in `progress_events.jsonl`; judge approve advances to the next step; judge receives the v1 schema request on stdin; `--auto --as agent` rejected by flag exclusivity; `fest init` writes the commented hooks placeholder and it does not parse into active hooks.

PR #241, opened 2026-07-08
fix: record workflow decision metadata

lancekrogers:
## Summary
- Add `--as user|agent` and `--summary` to workflow approval/rejection commands.
- Require explicit summaries for agent approvals and use rejection reasons as the agent rejection summary fallback.
- Persist decision actor, summary, and timestamp through workflow state and progress events.
- Clear stale decision metadata when reset or failed-remediation recheck reopens a step.

## Audit caveats checked
- Addresses the still-live approval metadata caveat from `workflow/design/fest-planning-loop-ux-fixes`: checkpoint decisions now record who made them and the approval/rejection rationale.
- Kept this separate from the feedback criteria branch so review can focus on workflow state/event persistence.

## Tests
- `go test ./internal/commands/workflow ./internal/guidance/workflow ./internal/progress`
- `just test all`

PR #248, opened 2026-07-09
fix(workflow): reserve the agent decision actor for the --auto judge path

lancekrogers:
## Problem

Agents have started clearing blocking checkpoints with `fest workflow approve --as agent`, bypassing the human-AI checkpoint model. This regressed on 2026-07-09:

- #241 (decision metadata) added `--as agent` to manual `approve`/`reject` with no guard beyond `--summary`. The help text, the validation hint (`use --as user or --as agent`), and a worked example in reject's help (`fest workflow reject --as agent --reason ...`) all read as a sanctioned agent lane, so agents that hit a gate and check `--help` now find documented instructions for clearing it themselves.
- #245 (approval auto mode guardrails) locked down `--auto` (fail-closed judge, explicit config) but left the `--as agent` manual path unguarded. Internally, the agent actor was meant to mark judge-mediated decisions (`runApproveAuto` stamps it after a real verdict); the flag let any headless agent claim that actor with a self-written summary.

Contrast: `fest workflow skip` requires an interactive TTY and is annotated operator-only. `approve --as agent` had neither.

## Fix

The agent actor is reserved for judge-mediated decisions recorded by `fest workflow approve --auto`, so an agent-actor entry in the decision audit trail always corresponds to an explicit operator delegation. Manual `approve`/`reject` accept only the user actor.

The `--as` flag itself is now **hidden from help, completion, and the generated CLI reference** — it survives only as an erroring tombstone so agents that learned `--as agent` during the window it was advertised get a stop-and-hand-off hint rather than an unknown-flag error they would retry without the flag (which would succeed and be recorded as `user`):

```
Error: --as agent is not allowed for manual approval
Hint: blocking checkpoints are operator decisions: stop, present the work, and ask the operator to run 'fest workflow approve'; agent-actor decisions enter the audit trail only through 'fest workflow approve --auto' with a configured approval judge (hooks.approval_judge.command)
```

Rejection gets the same treatment; the `reject --as agent` example is removed from the help and docs. `--as user` behavior is unchanged for anything that adopted it.

## Deliberately out of scope

A skip-style TTY gate on plain `fest workflow approve` would close the remaining unlabeled headless path, but festival-app's approve button shells out to exactly `fest workflow approve` without a TTY on behalf of a human click (`src-tauri/src/cli/commands.rs`). Hardening that path needs an app-aware design (mediated-operator marker or PTY) and is captured as a campaign intent rather than bundled here.

## Testing

- Unit: agent actor rejected for approval and rejection with the redirect hint; invalid actors still rejected; user default and summary trimming covered; hint always names the `--auto` delegation path.
- Full `go test ./...` green; `just lint all` 0 issues; `go vet` clean.
- Smoke-tested the built binary inside a real festival: `approve --as agent` and `reject --as agent` fail closed with the hint before touching workflow state; `--as` no longer appears in `--help`.

Docs regen commits are separate and scoped to the affected pages.

### DIFF
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -6972,6 +6972,10 @@ Auto approval:
   Manual approval is the default. Use --auto only when an operator has explicitly
   delegated this checkpoint decision to an external judge command.
 
+  Agents must not clear checkpoints themselves: --as agent is rejected. An
+  agent-actor decision is recorded only when the operator delegates via --auto
+  and the judge returns a verdict.
+
   The judge command receives JSON on stdin using schema fest.approval.judge/v1
   and must return JSON on stdout with decision "approve" or "reject" and a
   reason. Missing commands, timeouts, non-zero exits, malformed JSON, unknown
@@ -6992,7 +6996,7 @@ fest workflow approve [flags]
 ### Options
 
 ```
-      --as string                decision actor: user or agent (default "user")
+      --as string                decision actor (user; agent decisions require --auto with a configured judge) (default "user")
       --auto                     delegate this checkpoint decision to the configured approval judge command
   -h, --help                     help for approve
       --judge-command string     approval judge command for --auto (overrides the .festival/config.yaml hooks.approval_judge.command hook)
@@ -7128,7 +7132,7 @@ Examples:
 ```bash
   fest workflow reject --reason "needs revision"
   fest workflow reject --reason "PR not ready" --remediation-phase 005_FIX_PR_302
-  fest workflow reject --as agent --reason "missing acceptance proof"
+  fest workflow reject --reason "missing acceptance proof" --summary "reviewed the diff against the task spec"
 ```
 
 ```
@@ -7138,7 +7142,7 @@ fest workflow reject [flags]
 ### Options
 
 ```
-      --as string                  decision actor: user or agent (default "user")
+      --as string                  decision actor (user; agent decisions require 'fest workflow approve --auto' with a configured judge) (default "user")
   -h, --help                       help for reject
   -r, --reason string              reason for rejection (required)
       --remediation-phase string   link a remediation phase for a failed gate (e.g. 005_FIX_PR_302)
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -6996,7 +6996,6 @@ fest workflow approve [flags]
 ### Options
 
 ```
-      --as string                decision actor (user; agent decisions require --auto with a configured judge) (default "user")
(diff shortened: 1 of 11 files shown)


obey-agent (commented):
## Verdict: **Approve**

## Overview
Closes a real gate-bypass regression: manual `fest workflow approve|reject --as agent` looked like a sanctioned agent lane (help, hints, reject example) and let agents clear blocking checkpoints without operator or judge mediation. Manual path now accepts only `user`; `agent` is reserved for `approve --auto` after a real judge verdict.

## Review notes
- **Correct product model:** matches `workflow skip` (operator/TTY) spirit — agents stop and hand off rather than self-clear gates.
- **Auto path intact:** `runApproveAuto` still stamps `DecisionMetadata{Actor: decisionActorAgent, Summary: decision.Reason}` — audit trail for delegated decisions preserved.
- **UX:** rejection message + hints redirect to operator `approve`/`reject` or configured judge; docs regenerated (including removing the `reject --as agent` example).
- **API cleanup:** drops agent summary fallback from `normalizeDecision` (no longer needed); tests cover agent rejection, invalid actor, user default/trim, and hint content.

## Non-blocking
- `--as user` remains redundant but harmless (default).
- Reject still doesn't require `--summary` for users (reason is required); fine.

## Staff Standard
Yes — small, correct, fail-closed. Merge promptly so agents stop learning the old help text.

Reviewed against: https://github.com/Obedience-Corp/fest/pull/248 @ `1845542`

obey-agent (commented):
## Verdict: **Approve** (delta re-review)

## Overview
Original gate-actor fix still stands. Three follow-up commits (`9102f00`…`1d75ed7`) tighten the same fail-closed product model: hide the deprecated `--as` flag, and make checkpoint/step guidance **judge-aware** so agents are only told to run `approve --auto` when `hooks.approval_judge.command` is configured.

## Delta assessment
- **Templates:** no-judge path tells the agent to stop and hand off (does not advertise bare `--auto` as step 1); judge path gives a clear auto sequence with fail-closed language.
- **`approvalJudgeConfigured`:** nil ctx / missing workspace / load error ⇒ false (fail closed). Good.
- **Tests:** `navigator_format_judge_test.go` covers both branches and negative config cases.
- **Product risk:** low — reduces agent self-approval pressure without changing the enforcement path already locked in earlier commits.

## Non-blocking
- Config is re-read on each format call; fine for CLI volume.
- No security concern: guidance-only change; enforcement remains on the approve command path.

## Staff Standard
Yes — this is the right UX follow-through on the gate-actor fix. Still mergeable.

Reviewed against: https://github.com/Obedience-Corp/fest/pull/248 @ `1d75ed7`

PR #243, opened 2026-07-09
feat(show,list): move status listings to fest list; cyclable fest show TUI

lancekrogers:
## Summary

Reassigns each command to the surface it is best at (design package: `obey-campaign/workflow/design/fest-show-list-ux-redesign/`, workitem WI-db00a4):

- **`fest list` owns status listings.** `fest list <tab>` now completes the full status vocabulary with descriptions (`active`, `ready`, `planning`, `ritual`, `completed`, `dungeon`, `dungeon/*`, `all`), sourced from `id.StatusDirectories` so it cannot drift. `fest list all` is a positional alias for `--all`.
- **`fest show` owns single-festival display.** The five status subcommands are hidden + deprecated (they keep working; the notice goes to stderr so `--json` stdout stays clean). `fest show <tab>` completes festival selectors.
- **Bare `fest show`** outside a festival in an interactive terminal now cycles browseable festivals (active/ready/planning/ritual, status-then-recency) with the same arrow-key UX as `fest watch`; `q`/Ctrl+C exit. In-festival, non-TTY, and `--json` behavior unchanged.
- **zsh: `fest show <tab>`** on an empty prefix opens the festival picker TUI and inserts the pick; with a prefix it gets colorized status-ordered menu completion (same pattern as `fest promote <tab>`). New hidden helpers: `fest show pick`, `fest show completions --color`. bash gets plain name completion (no safe TUI-on-tab there).

## Commit structure

1. `feat(show,list)` — the feature: code, tests, shell integration, feature-scoped doc pages.
2. `docs` — `just docs` regeneration at the feature HEAD: combined `fest-reference.md`, previously-unregenerated `fest understand` pages, EOF normalization on workflow/walk/watch pages. Split out so the feature commit reviews clean.

## Design decisions / tradeoffs

- Statuses were NOT converted to cobra subcommands on `list` — that would shadow the positional path and break `fest list active --sort progress`. Completion vocabulary gives the same tab UX with zero regression.
- The cycle engine moved from `watch` into `show` (`show/cycle.go`) because `promote` imports `show` — an `ExtraKeys` hook lets `watch` keep its `p` promote without `show` importing `promote`. `q`/`Q` quit is registered by show through the same hook (not the base classifier) so `fest watch` keypress behavior is untouched.
- A `RenderFallback` preserves watch's exact pre-change non-TTY output (no CRLF translation or cycle-key hints in piped output).
- Deleting the deprecated `fest show <status>` aliases outright is deferred one release (production users may script them).

[...]

PR #249, opened 2026-07-09
feat(show,watch): scrollable per-festival viewport in the cycle TUI

lancekrogers:
Follow-up to #243. Long festivals overflow the cleared frame in the `fest show` / `fest watch` cycle TUI with no way to reach the top of the content.

## Design
- `FrameState` (new) is shared between the cycle key loop and the watch renderer: the key loop moves the viewport offset (clamped to content extents recorded on each draw) and signals a redraw channel; the renderer repaints the cached frame through the viewport. Scrolling never tears down the file-watch session, so it stays responsive and the watcher stays attached.
- Keys: up/down arrows scroll by line, `space`/`b` page down/up (pager convention). Registered extra keys (watch's `p` promote, show's `q` quit) win over the built-ins. Offset resets when cycling to another festival; survives live refreshes.
- Footer gains a range indicator only when content overflows: `↑↓/spc/b scroll (6-15/128)`.
- `fest watch` gets the same scrolling for free via the shared engine.

## Explicitly unchanged
Non-TTY output, `--json`, plain `fest show --watch`/`fest watch` outside cycle mode (no raw mode, so no key loop), standalone workflow watch, and short-content frames. The viewport engages only when a frame is attached AND the terminal size is known AND content overflows.

## Verification
- New tests: viewport slicing/clamping at both edges, page scrolling, redraw signaling (change-only), reset-on-cycle, session draw with injected heights, nil-frame and unknown-height fallbacks; key classification for the new keys incl. extra-key precedence. All with `-race`.
- `just lint all` 0 issues; `just test unit` 2454/2454; `just test integration` 59/59 (Docker, incl. watch PTY tests); `just docs` zero drift; windows cross-build of show/watch.
- Real binary: non-TTY `fest show` output byte-identical behavior (fails fast / renders once, no scroll artifacts).
- Needs interactive verification (cannot be simulated headlessly): arrow scrolling + footer indicator in a real terminal over a long festival.

### DIFF
--- a/internal/commands/show/cycle.go
+++ b/internal/commands/show/cycle.go
@@ -23,6 +23,14 @@ const (
 	CycleQuit
 	// CycleExtra dispatches a caller-registered ExtraKeys handler.
 	CycleExtra
+	// CycleScrollUp moves the content viewport up one line.
+	CycleScrollUp
+	// CycleScrollDown moves the content viewport down one line.
+	CycleScrollDown
+	// CyclePageUp moves the content viewport up one page.
+	CyclePageUp
+	// CyclePageDown moves the content viewport down one page.
+	CyclePageDown
 )
 
 // ExtraKeyHandler runs a custom action for a caller-registered key (e.g. watch's
@@ -41,7 +49,7 @@ type ExtraKeyHandler func(ctx context.Context, festival *FestivalInfo, rawState 
 // non-interactive consumers.
 type CycleOptions struct {
 	Detect         func(ctx context.Context, path string) (*FestivalInfo, error)
-	Render         func(ctx context.Context, festival *FestivalInfo, cycling bool) error
+	Render         func(ctx context.Context, festival *FestivalInfo, cycling bool, frame *FrameState) error
 	RenderFallback func(ctx context.Context, festival *FestivalInfo) error
 	ExtraKeys      map[byte]ExtraKeyHandler
 }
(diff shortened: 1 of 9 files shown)


obey-agent (commented):
## Staff Review - Merge Ready

**Verdict:** Approve

**Overview:** Clean viewport design for the cycle TUI. `FrameState` cleanly separates concerns: the key loop owns the offset, the renderer owns extents and drawing, and the redraw channel keeps scrolling inside the live render session so the file watcher never detaches. `fest watch` inherits scrolling through the shared engine.

**Code notes verified:** scroll keys are handled before the render-cancel path in `RunCycle` (no watcher churn while scrolling); offset resets on festival switch and persists across live refreshes with clamping; registered extra keys (watch's `p`, show's `q`) take precedence over the new space/b built-ins; the viewport engages only when a frame is attached AND terminal height is known AND content overflows - so non-TTY, `--json`, plain watch, and standalone workflow watch are untouched (footer test coverage updated accordingly).

**Verification (local, at head of show-scroll):**
- New tests: slicing/clamping at both edges, page scroll, change-only redraw signaling, reset-on-cycle, injected-height session draws, nil-frame + unknown-height fallbacks, key classification incl. extra-key precedence - all under `-race`
- `just lint all`: 0 issues; `just test unit`: 2454/2454; `just test integration`: 59/59 (Docker, incl. watch PTY tests); `just docs`: zero drift; windows cross-build of show/watch
- Real binary: non-TTY `fest show` behavior unchanged

**Remaining manual check (interactive-only):** arrow scrolling + footer range indicator in a real terminal over a long festival.

**Staff standard:** Yes - the cached-content redraw path and change-only signaling show real care for terminal responsiveness.

PR #251, opened 2026-07-10
refactor: remove dead fest commits command

lancekrogers:
## Summary

Removes the dead `fest commits` query command and the fossil task-ref highlighter that shared its retired format.

## Why remove instead of fix

`fest commits` has silently returned `No commits found` for every query since the commit reference format migrated:

- Its discovery grep matched the retired `[FEST-xxxxxx]` task-ref format (`commits.go:27`), while `fest commit` writes `[{campaign}:{cid}-FE-{fid}]` tags (e.g. `[obey-campaign:8deed8b4-FE-CF0001]`). Verified live from inside a completed festival with dozens of tagged commits.
- Its "not inside a festival" error advertised `--festival <path>`, but the command never defined that flag (`Error: unknown flag: --festival`).
- It predated the `shared.PickFestival` no-context picker that newer commands use.

The 2026-07-09 campaign CLI usage audit measured zero usage across human and agent history. Since the command could not return results, nobody can have been depending on it; the breakage window doubled as a no-user proof. A real fix would be a redesign, not a pattern swap: the new tag carries only the festival id, so `--task`/`--sequence`/`--phase` filtering would need message-body matching designed from scratch, and campaign-wide history queries belong with campaign-side tooling if ever wanted. Owner decision: remove.

## Changes

- Delete `internal/commands/commit/commits.go` and its root registration (query group).
- Remove `highlightTaskRefs` from `fest commit` output: it colorized the same retired `[FEST-xxxxxx]` format and has been a no-op on every message the current format produces.
- Regenerate CLI reference docs (separate commit, pure deletions).

## Testing

- `just lint all`: 0 issues. `just test unit-raw`: no failures.
- Real binary: `fest commits` now reports `unknown command "commits"`; gone from `fest --help`.

Tracking: campaign intent `fest-commits-discovery-is-outdated-20260709-230616` (also notes the outdated pitch-deck demo line for the next deck refresh).


internal/commands/commit/commit.go
```diff
@@ -286,7 +286,7 @@ func outputResult(result *CommitResult) error {
 		if result.Success {
 			fmt.Println(ui.H1("Commit"))
 			fmt.Printf("%s %s\n", ui.Label("Hash"), ui.Value(result.Hash))
-			fmt.Printf("%s %s\n", ui.Label("Message"), highlightTaskRefs(result.Message))
+			fmt.Printf("%s %s\n", ui.Label("Message"), result.Message)
```
obey-agent, line 289: Removing highlightTaskRefs is correct for the retired [FEST-xxxxxx] pattern, but commit messages still carry tags like [campaign:cid-FE-fid]. Consider a follow-up (not this PR) that highlights the *current* tag grammar so commit success output stays scannable. Not a merge blocker.

internal/commands/root.go
```diff
@@ -362,10 +362,6 @@ func init() {
 	commitCmd.GroupID = "workflow"
```
obey-agent, line 362: Clean unregistration. Flag for merge: this branch and #252 both delete chunks of docs/cli-reference/fest.md and fest-reference.md — expect a trivial docs conflict if they land close together; rebase the second merge on main after the first.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Correct removal of a command that has been functionally dead since the commit-tag migration. Evidence in the PR body (wrong grep format, missing `--festival` flag, zero usage) justifies delete-over-fix. Docs and root registration are cleaned up consistently.

## Key Findings
- **Current-tag highlighting** (inline on `commit.go`): optional follow-up to color the live `[campaign:…-FE-…]` format; not a blocker.
- **Merge with #252** (inline on `root.go`): both PRs touch the same CLI reference docs — expect a small conflict if merged independently.

## What's Done Well
- Honest product call: redesigning discovery is a new feature, not a one-line fix.
- Also removes the fossil `highlightTaskRefs` no-op tied to the old format.
- Docs regeneration kept pure deletions.

## Staff Standard
Yes. This is the right kind of CLI surface cleanup after an audit: remove the broken affordance rather than leave a lying command in stable help.

PR #252, opened 2026-07-10
refactor: move fest tui to the dev release channel

lancekrogers:
## Summary

Moves `fest tui` from the stable command surface to the dev release channel, alongside `fest explore`. Stable builds no longer expose it; dev builds keep it, marked dev-only.

## Why dev-gate instead of update or remove

The dedicated fest-tui audit (`workflow/audits/fest-tui-2026-07-09/` in the campaign) found:

- It is a **create-only huh menu presenting itself as festival management** ("Interactive festival management" in help suggestions, "creation and editing" in its Short). No edit, browse, status, or workflow actions exist in the menu.
- **Near-zero real use**: 2 human launches ever (vs 34 `fest status`, 13 `fest create`); zero agent launches (correctly `agent_allowed: false`).
- Bare `fest create` already provides the same interactive forms on the stable surface (`StartCreateTUI`), and that path is hot. It is untouched by this PR.

Updating it into real management would duplicate festival-app and `fest watch`/`fest explore`. Removing it deletes working, type-current create forms plus wizard/goal menu entries. Dev-gating gets the public-surface win (one less misleading top-level command for already-confused users) while keeping the code as a seed for a future terminal management UI, using the release-channel pattern that already exists for `explore`.

## Changes

- `root.go` no longer registers the command; `release/register_dev.go` registers it via the existing `shared.NewTUICommand` hook with `MarkDevOnly`, nil-guarded for `no_charm`-style builds without the hook.
- Charm command Short corrected to what it does: creating festivals, phases, sequences, and tasks (no editing exists).
- Generic help suggestions drop the `fest tui` row, which would otherwise recommend a command stable builds do not have.
- Tests: dev-profile registration + release-channel annotation, stable-profile absence, dev command-surface inclusion, updated contextual-help expectation.
- Docs regenerated in a separate commit (pure deletions of the stable reference page).

## Testing

- `just lint all`: 0 issues. `just test unit-raw`: no failures. Both build profiles compile.
- Real binaries: stable → `Error: unknown command "tui" for "fest"`; dev → listed with the corrected Short.

Context: 2026-07-09 CLI usage audit follow-through (campaign intent `decommission-dead-cli-surface-per-20260709-234257`, Step 1). Related: #251 (fest commits removal).


internal/commands/release/register_dev.go
```diff
@@ -12,4 +13,14 @@ func registerDev(root *cobra.Command) {
 	exploreCmd.GroupID = "navigation"
 	MarkDevOnly(exploreCmd)
 	root.AddCommand(exploreCmd)
+
+	// fest tui is a create-only menu with near-zero human uptake; the public
+	// create surface is bare `fest create` (same forms). Kept in the dev
+	// channel as a seed for a future terminal management UI.
+	if shared.NewTUICommand != nil {
```
obey-agent, line 20: Good use of the existing dev-registration hook with a nil-guard for no_charm builds. Matches explore. One product note: if any external docs/scripts still invoke fest tui on stable, they will start failing closed — worth a one-line release note ("stable: fest tui moved to dev build") when this ships.

internal/commands/tui/charm.go
```diff
@@ -28,7 +28,7 @@ func init() {
 func NewTUICommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "tui",
-		Short: "Interactive UI (Charm) for festival creation and editing",
+		Short: "Interactive UI (Charm) for creating festivals, phases, sequences, and tasks",
```
obey-agent, line 31: Correcting Short to create-only is accurate and removes the "management" overclaim. Consider also checking Long help text (if any) for residual "editing" language so the command surface is consistent.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Dev-gating `fest tui` is the right intermediate between “misleading stable command” and “delete working create forms.” Reuses the existing release-channel pattern (`explore`), corrects the Short, drops stable help suggestions, and locks behavior with stable/dev registration tests.

## Key Findings
- **Release note** (inline on `register_dev.go`): stable users who still type `fest tui` will get unknown-command — call it out in the release notes.
- **Help copy consistency** (inline on `charm.go`): Short is fixed; skim Long for residual “editing” claims.
- Coordinate merge order with #251 (shared docs deletions).

## What's Done Well
- Nil-guard on `shared.NewTUICommand` for no_charm builds.
- Bare `fest create` left as the stable interactive path — no user-value regression for the hot path.
- Tests for both profiles and command-surface inclusion.

## Staff Standard
Yes. Clean channel split with evidence-backed rationale and no over-deletion of seed code.

PR #254, opened 2026-07-10
fix: eliminate 5s tty-query stall at startup; q quits watch; watch survives write storms

lancekrogers:
## Problem

Recording a fest demo froze the recorder and eventually took the whole machine down (WindowServer watchdog kill at 01:29 after logd's firehose queues hung under memory pressure). Diagnosis with a controlled pty harness traced it to three fest-side defects:

1. **Every fest command stalls up to 5s on ptys that don't answer terminal queries.** bubbletea v1 calls `lipgloss.HasDarkBackground()` from a package `init()` (`tea_init.go:21`), which makes termenv write OSC 11 + DSR queries to the tty and block in `waitForData(5s)`. Interactive terminals answer instantly so humans never see it; recorder/CI/agent ptys that never answer freeze every command — even `fest version --no-color` — for the full timeout before the first byte of output. Verified by SIGQUIT goroutine dump of a stalled process.
2. **`fest watch` could only be exited with Ctrl+C** — its cycle registered only `p`/`P`, unlike show's cycle which quits on `q`. A recorder driving the view with `q` waits forever on a session that never ends.
3. **Sustained writes starve the watch refresh.** The trailing debounce timer resets on every fs event, so events arriving faster than 100ms apart defer the callback until the storm ends — the live view goes stale exactly when things are busiest.

## Fix

- New `internal/bginit` package seeds `lipgloss.SetHasDarkBackground(...)` from `COLORFGBG` (dark fallback, mirroring termenv's own non-query fallback) and is imported first in `cmd/fest` so it initializes before the bubbletea subtree. lipgloss then skips the terminal query everywhere (`explicitBackgroundColor`). An explicit `light`/`dark`/`high-contrast` theme refines the value in `ui.InitPalette`; `adaptive` keeps the env seed rather than reintroducing the query.
- `fest watch` cycle registers `q`/`Q` to quit; the cycle footer now advertises `q/Ctrl+C exit` in both show and watch.
- `watch.Config.MaxWait` caps debounce coalescing (500ms for the display refreshers: show watch, standalone watch, progress watch).

## Tradeoff

Adaptive background detection no longer queries the terminal, so light terminals without `COLORFGBG` get dark-variant adaptive colors in huh forms unless the user sets `fest config theme set light`. The CLI's main palette already treated adaptive as dark, so the visible delta is small — and it buys deterministic, instant startup on every pty.

## Verification (controlled pty harness, before → after)

| Scenario | Before | After |
| --- | --- | --- |
| `fest version`, mute pty | 5.03s, OSC11+2×DSR emitted | 0.35s, zero queries |
| `fest show <name>`, mute pty | 5.36s stall then output | 0.04s, colored, exits |
| `fest watch` + `q` | ignored (Ctrl+C only) | exits rc=0 |
| `fest show` cycle + `q` | exits | exits (no regression) |
| `fest watch` under 50 writes/s for 6s | 3 redraws (stale view) | 14 redraws (~2/s), output bounded |

Gates: `just build dev` ✓, `just test unit` 2473/2473 ✓, `just lint all` 0 issues ✓

### DIFF
--- a/internal/watch/watcher_test.go
+++ b/internal/watch/watcher_test.go
@@ -193,6 +193,49 @@ func TestWatcher_Debouncing(t *testing.T) {
 	cancel()
 }
 
+func TestWatcher_MaxWaitFiresDuringSustainedChanges(t *testing.T) {
+	tmpDir := t.TempDir()
+	tmpFile := filepath.Join(tmpDir, "test.txt")
+	if err := os.WriteFile(tmpFile, []byte("initial"), 0644); err != nil {
+		t.Fatalf("failed to create temp file: %v", err)
+	}
+
+	var callCount atomic.Int32
+
+	cfg := Config{
+		Paths:    []string{tmpFile},
+		Debounce: 100 * time.Millisecond,
+		MaxWait:  250 * time.Millisecond,
+	}
+
+	w, err := New(cfg, func() {
+		callCount.Add(1)
+	})
+	if err != nil {
+		t.Fatalf("New() error = %v", err)
+	}
+	defer func() { _ = w.Close() }()
+
+	// Sustained events every 20ms reset the trailing debounce timer forever;
+	// only the MaxWait cap lets the callback fire while the storm is running.
+	storm := 700 * time.Millisecond
+	deadline := time.Now().Add(storm)
+	for time.Now().Before(deadline) {
+		w.scheduleCallback()
+		time.Sleep(20 * time.Millisecond)
+	}
+	duringStorm := callCount.Load()
+
+	if duringStorm < 1 {
+		t.Errorf("expected at least 1 callback during a %v storm with MaxWait=%v, got %d", storm, cfg.MaxWait, duringStorm)
+	}
+
+	time.Sleep(300 * time.Millisecond)
+	if total := callCount.Load(); total <= duringStorm {
+		t.Errorf("expected a trailing callback after the storm, got %d during and %d total", duringStorm, total)
+	}
+}
+
 func TestWatcher_WatchDirectory(t *testing.T) {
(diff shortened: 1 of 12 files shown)


internal/watch/watcher_test.go
```diff
@@ -193,6 +193,49 @@ func TestWatcher_Debouncing(t *testing.T) {
 	cancel()
 }
 
+func TestWatcher_MaxWaitFiresDuringSustainedChanges(t *testing.T) {
+	tmpDir := t.TempDir()
```
obey-agent: **[blocking]** This newly added test creates and writes a host `t.TempDir`, which violates the campaign filesystem-safety contract for tests that exercise real filesystem mutation. The test already calls `scheduleCallback` directly, so it does not need `fsnotify` or a real path: instantiate a package-local `Watcher` with `Config{Debounce, MaxWait}` and the callback, then `defer w.cancelPendingTimer()`. That keeps this a deterministic scheduler unit test and avoids adding another host-mutation case.

obey-agent (approved):
Verdict: Approve

The startup fix is technically sound and verified against the actual failure mode: bginit seeds lipgloss before the bubbletea subtree, explicit themes refine the cached value without terminal queries, watch cycles now honor q/Q, and MaxWait bounds refresh starvation during sustained event storms.

One blocking review finding was fixed at 060fa3b: the newly added scheduler test no longer creates or writes a host TempDir and now exercises scheduleCallback as a pure unit test.

Verification rerun:
- focused bginit/watch/show/progress packages passed
- full just test unit gate passed across 96 packages
- just lint all: 0 issues
- controlled mute-PTY fest version: 0.42s, no five-second query stall

I would be comfortable submitting and operating this change.

PR #256, opened 2026-07-12
feat(workflow): add --json to fest workflow status

lancekrogers:
## Summary

Adds a structured `--json` output to `fest workflow status` so consumers can read the current workflow position without parsing the human-readable text. This is the implementation half of RI0002 sequence 02 (design workitem WI-1e4ac2); the festival-app consumer will use it to populate `FestContext.workflow` and `FestContext.workflow_step`.

Default text output is unchanged.

## Contract (schema `fest.workflow.status/v1`)

Top-level fields: `schema_version`, `festival_id`, `festival_name`, `festival_path`, `phase`, `phase_path`, `mode` (`workflow` or `gate`), `workflow`, `workflow_step`, `current_step`, `total_steps`, `complete`, and `steps[]`.

Each `steps[]` entry: `number`, `name`, `status`, `is_current`, `has_checkpoint`, `goal`, and `feedback` / `remediation_phase` (omitted when empty).

Null and empty behavior:

- When the workflow is complete, `current_step` and `workflow_step` are `null` and `complete` is `true`.
- A phase with no navigable workflow returns a valid object with `complete: true`, `current_step: null`, and `steps: []` (an empty array, never `null`).
- Missing festival or phase context returns the same non-zero command error as text mode.

## Implementation

- `status.go`: adds the `--json` flag on `newStatusCmd` and threads `jsonOutput` through `runStatus`, branching to structured output before the text path. Text rendering is untouched.
- `status_json.go`: private DTOs plus `collectWorkflowStatus`, built entirely from structured navigator state (`nav.GetSteps()`, `nav.GetWorkflowState()`, `nav.Ctx`) and never from the rendered string. Festival id and name come from `config.LoadFestivalConfig` (best effort; empty if unavailable). Gate mode is derived from `nav.IsGate()`.
- Encodes with `encoding/json` (`MarshalIndent`); JSON errors use the repo `errors.Parse` convention, matching the existing `--json` commands.

## Tests

`status_json_test.go` covers the normal workflow snapshot, the completed state (null current step), gate mode, empty-steps array serialization, and a real command-path test asserting both `--json` emits parseable JSON and text mode still renders the human header without leaking JSON.

## Validation

- `just lint all`: 0 issues
- `just test unit-raw`: 0 failures
- Real end-to-end run of `fest workflow status --json` against a scaffolded festival produced the full contract, and text mode plus `--help` (now documenting `--json`) were confirmed unchanged.

The CLI reference under `docs/cli-reference/` was regenerated in a separate commit; only the workflow-status entry changed (the new flag and schema note), with no unrelated command drift.

### DIFF
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -7331,6 +7331,9 @@ Shows:
   - Remaining steps
   - Checkpoint status if applicable
 
+Use --json for a stable machine-readable snapshot (schema fest.workflow.status/v1)
+that consumers can read without parsing the human-readable output.
+
 ```
 fest workflow status [flags]
 ```
@@ -7339,6 +7342,7 @@ fest workflow status [flags]
 
 ```
   -h, --help   help for status
+      --json   output as JSON
 ```
 
 ### Options inherited from parent commands
--- a/docs/cli-reference/fest_workflow_status.md
+++ b/docs/cli-reference/fest_workflow_status.md
@@ -12,6 +12,9 @@ Shows:
   - Remaining steps
   - Checkpoint status if applicable
 
+Use --json for a stable machine-readable snapshot (schema fest.workflow.status/v1)
+that consumers can read without parsing the human-readable output.
+
 ```
 fest workflow status [flags]
 ```
@@ -20,6 +23,7 @@ fest workflow status [flags]
 
 ```
   -h, --help   help for status
+      --json   output as JSON
 ```
 
 ### Options inherited from parent commands
--- a/internal/commands/workflow/status.go
+++ b/internal/commands/workflow/status.go
@@ -20,7 +20,8 @@ import (
 )
 
 func newStatusCmd() *cobra.Command {
-	return &cobra.Command{
+	var jsonOutput bool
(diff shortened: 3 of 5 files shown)


obey-agent (approved):
## Verdict: Approve

## Overview
First review of head `89935f5`. Adds `--json` to `fest workflow status` with a versioned schema (`workflow-status/v1` via `schema_version`), structured step list, gate vs workflow mode, and null current-step semantics when complete or empty. Docs regenerated for the CLI reference.

## Key Findings
- `collectWorkflowStatus` / `renderWorkflowStatusJSON` separation is clean; empty workflows serialize `steps: []` (not null) and `complete: true` — good agent contract.
- Error path uses fest `errors.Parse` for marshal failures.
- Tests cover normal, complete, gate mode, empty steps array, and the real `runStatus` entrypoint for both JSON and text (including “no JSON leak” in text mode).
- Festival config load for id/name is best-effort (`err == nil`); acceptable for status enrichment.

## What's Done Well
Stable schema_version field; pointer fields for current step/name so complete state is unambiguous JSON.

## Staff Standard
Yes — ship it.

PR #257, opened 2026-07-12
feat(list): fest list --watch live multi-festival status board

lancekrogers:
## Summary

- Add `fest list --watch` / `-w` to continuously refresh the multi-festival status board in place
- Reuses existing list filters (`status`, `--progress`, `--sort`, `--filter-project`, date filters)
- Explicitly **does not** cycle between festivals (that remains `fest watch`)
- Rejects `--watch --json` and non-TTY usage; fsnotify on status dirs + 2s poll for progress updates

Design workitem: `workflow/design/fest-list-watch` (`WI-2a7950`)

## Test plan

- [x] `go test ./internal/commands/list/`
- [x] Smoke: `fest list --watch --json` errors
- [x] Smoke: piped `fest list --watch` requires interactive terminal
- [x] Smoke: one-shot `fest list` still renders the board
- [ ] Manual: `fest list --watch` in a real campaign TTY; confirm progress/status moves refresh without cycling UI

### DIFF
--- a/internal/commands/list/list.go
+++ b/internal/commands/list/list.go
@@ -395,269 +395,53 @@ func applyFilters(festivals []*show.FestivalInfo, opts *listOptions) ([]*show.Fe
 var dungeonSubstatuses = []string{"dungeon/completed", "dungeon/archived", "dungeon/someday"}
 
 func listDungeon(ctx context.Context, festivalsDir string, opts *listOptions, campaignRoot string) error {
-	result := make(map[string]interface{})
-	var totalCount int
-	allFestivals := make(map[string][]*show.FestivalInfo)
-	var allFestivalsList []*show.FestivalInfo
-
-	order := make([]string, 0, len(dungeonSubstatuses))
-	for _, status := range dungeonSubstatuses {
-		festivals, err := show.ListFestivalsByStatus(ctx, festivalsDir, status, campaignRoot)
-		if err != nil {
-			continue
-		}
-		festivals, err = applyFilters(festivals, opts)
-		if err != nil {
-			return err
-		}
-		if len(festivals) > 0 {
-			// Default dungeon listings order by bucket date (newest first);
-			// explicit --sort or --alpha still wins.
-			if opts.sortBy == "" && !opts.alpha {
-				sortByStatusDate(festivals)
-			} else {
-				applySorting(festivals, opts.sortBy, opts.alpha)
-			}
-			allFestivals[status] = festivals
-			order = append(order, status)
-			totalCount += len(festivals)
-			allFestivalsList = append(allFestivalsList, festivals...)
-		}
+	board, err := collectDungeonBoard(ctx, festivalsDir, opts, campaignRoot)
+	if err != nil {
+		return err
 	}
-
-	var progressMap map[string]*progress.FestivalProgress
-	if opts.progress {
-		progressMap = fetchProgressForFestivals(ctx, allFestivalsList)
-	}
-
 	if opts.json {
-		for status, festivals := range allFestivals {
-			result[status] = festivalsToMapWithProgress(festivals, progressMap)
+		result := make(map[string]interface{}, len(board.Festivals)+1)
+		for status, festivals := range board.Festivals {
+			result[status] = festivalsToMapWithProgress(festivals, board.Progress)
 		}
-		result["total"] = totalCount
+		result["total"] = board.Total
 		return outputJSON(result)
 	}
-
-	fmt.Print(formatDungeonHuman(allFestivals, order, progressMap, totalCount, opts.progress))
+	fmt.Print(formatDungeonHuman(board.Festivals, board.Order, board.Progress, board.Total, opts.progress))
 	return nil
 }
 
 func listByStatus(ctx context.Context, festivalsDir, status string, opts *listOptions, campaignRoot string) error {
-	festivals, err := show.ListFestivalsByStatus(ctx, festivalsDir, status, campaignRoot)
+	board, err := collectStatusBoard(ctx, festivalsDir, status, opts, campaignRoot)
 	if err != nil {
 		return err
 	}
-
-	festivals, err = applyFilters(festivals, opts)
-	if err != nil {
-		return err
-	}
-
-	// Dungeon listings default to newest bucket date first; explicit sort
-	// flags still take precedence.
-	if opts.sortBy == "" && !opts.alpha && strings.HasPrefix(status, "dungeon/") {
-		sortByStatusDate(festivals)
-	} else {
-		applySorting(festivals, opts.sortBy, opts.alpha)
-	}
-
-	// Fetch detailed progress if requested
-	var progressMap map[string]*progress.FestivalProgress
-	if opts.progress {
-		progressMap = fetchProgressForFestivals(ctx, festivals)
-	}
-
 	if opts.json {
 		return outputJSON(map[string]interface{}{
 			"status":    status,
-			"count":     len(festivals),
-			"festivals": festivalsToMapWithProgress(festivals, progressMap),
+			"count":     len(board.Festivals),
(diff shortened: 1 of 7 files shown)


internal/commands/list/list_watch.go
```diff
@@ -0,0 +1,145 @@
...
+			fmt.Println()
+			return nil
+		case err := <-errCh:
+			if err != nil && ctx.Err() == nil {
+				return err
+			}
+			fmt.Println()
+			return nil
+		case <-ticker.C:
+			if err := render(false); err != nil {
```
obey-agent: **Race: concurrent renderers without synchronization.**

`watch.New` fires `onChange` from an `AfterFunc` goroutine, while this select also redraws on `ticker.C` on the main goroutine. Both call `clearListScreen` + `fmt.Print` with no mutex.

Under write storms (or simply every 2s poll overlapping a debounced fs event), stdout frames can interleave and garble the board.

**Fix:** serialize all paints through one path — e.g. a single `renderMu`, or only poll *or* only fsnotify (prefer: event-driven callback + optional poll, but one mutex around the full clear+write frame). Mirror whatever `show`/`progress` watch does if they already solved this.

internal/commands/list/list_watch.go
```diff
@@ -0,0 +1,145 @@
...
+	fmt.Println()
+	label := filepath.Base(festivalsDir)
+	if label == "" || label == "." {
+		label = "festivals"
+	}
+	if polling {
+		fmt.Println(ui.Dim(fmt.Sprintf("watching %s — Ctrl+C to quit • Polling every %s", label, listPollingInterval)))
+		return
+	}
+	fmt.Println(ui.Dim(fmt.Sprintf("watching %s — Ctrl+C to quit", label)))
```
obey-agent, line 145: **Footer is misleading on the hybrid path.**

This hybrid always runs a 2s ticker for progress updates (good), but `render(false)` always prints the pure event-driven footer ("Ctrl+C to quit" with no polling hint). Operators cannot tell the board is hard-refreshing every 2s regardless of filesystem events.

Either:
1. Pass `polling=true` (or a hybrid label) when the ticker is active, or
2. Drop the always-on ticker and document that progress-only changes lag until a status-dir event (worse UX).

Prefer (1): footer should reflect reality.

internal/commands/list/list.go
```diff
@@ -515,18 +525,139 @@ func listAll(ctx context.Context, festivalsDir string, opts *listOptions, campai
...
 
-	if totalCount == 0 {
-		fmt.Println(ui.Warning("No festivals found."))
-		fmt.Println(ui.Info("Create a festival with: fest create festival"))
-		return nil
+	fmt.Print(formatAllHuman(allFestivals, statusOrder, progressMap, totalCount, opts.progress))
+	return nil
+}
+
+// formatListBoard builds the human-readable list board for the current filters.
```
obey-agent: **Duplicated collection logic will drift.**

`listDungeon` / `listByStatus` / `listAll` still gather festivals for one-shot (JSON + human), while `formatDungeonBoard` / `formatStatusBoard` / `formatAllBoard` re-implement the same gather/filter/sort/progress pipeline for watch.

The comment above claims formatListBoard is "Used by one-shot list (via list*)" but one-shot human output only shares `format*Human` string helpers — not the collection path. Sort defaults, filter edge cases, and progress behavior can diverge silently.

**Staff standard:** one collection function returning a board model, then JSON encode vs human format vs watch refresh all consume that model. Collapsing this now is cheaper than the second bug report when `--sort progress` behaves differently under `--watch`.

internal/commands/list/list_watch_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/Obedience-Corp/fest/internal/id"
+)
+
+func TestListWatchPathsIncludeStatusBuckets(t *testing.T) {
```
obey-agent, line 11: **Missing integration coverage for the live loop.**

Unit tests cover flag registration, `--json` rejection, non-TTY rejection, and path membership — good defensive surface.

They do **not** cover:
- first frame content matches one-shot `fest list` for the same filter
- cancel/SIGINT exits cleanly without hanging
- a status-dir create/move appears on a subsequent frame

The design workitem (`04-tests-and-acceptance.md`) called for a bounded container harness similar to `tests/integration/fest_watch_test.go`. Without that, this is shippable as a CLI convenience but not proven under the same isolation rules as other fest watch paths.

Add at least one integration test that starts `fest list --watch` under timeout, asserts festival names in the first frame, and expects timeout/signal exit codes like the watch harness.

internal/commands/list/list.go
```diff
@@ -515,18 +525,139 @@ func listAll(ctx context.Context, festivalsDir string, opts *listOptions, campai
...
 	}
-	return nil
+	var progressMap map[string]*progress.FestivalProgress
+	if opts.progress {
+		progressMap = fetchProgressForFestivals(ctx, festivals)
+	}
+	return formatStatusHuman(status, festivals, progressMap, opts.progress), nil
+}
+
+func formatAllBoard(ctx context.Context, festivalsDir string, opts *listOptions, campaignRoot string) (string, error) {
```
obey-agent: **Maintainability: list.go is past the project size comfort zone.**

This PR grows `list.go` by ~150 lines of near-duplicate board builders. Pre-existing file was already large; the new board path should live beside `list_watch.go` (or a `list_board.go`) so the one-shot command file stays scannable.

Not a merge blocker alone, but combined with the duplication finding it is a strong signal to extract before the next list feature.

obey-agent (commented):
## Staff Review — REQUEST CHANGES

### Overview
Solid product fit: `fest list --watch` is the right surface for a multi-festival board, validation (`--json` / non-TTY) is crisp, and reusing `internal/watch` matches existing progress/show patterns. The PR is not merge-ready yet because the live loop has a real concurrency hazard, the collection path is duplicated in a way that will drift, and the design’s integration bar was not met.

### Key Findings
1. **Concurrent render race** (`list_watch.go`) — fsnotify `onChange` and the 2s ticker both paint without a mutex; frames can interleave on stdout. Serialize all clears+writes.
2. **Misleading hybrid footer** — ticker always runs, but footer claims pure event-driven mode. Reflect polling/hybrid honestly.
3. **Duplicated board collection** (`list.go`) — `list*` and `format*Board` re-implement gather/filter/sort/progress. One model, many renderers.
4. **No integration test for the live loop** — flag unit tests are good; still need bounded first-frame + cancel coverage in the container harness (as the design doc specified).
5. **File growth** — board builders should not keep bloating `list.go`; extract with the dedupe.

### What’s Done Well
- Correct CLI choice (`--watch` on list, not a new top-level command; no festival cycling).
- Clear validation errors with agent-oriented hints.
- Reuse of fest’s debounced `watch` package and 2s polling cadence consistency with show/progress.
- Docs regenerated (`fest_list.md` + combined reference).
- Unit tests for the failure modes agents will hit first.

### Staff Standard
Would I be proud to submit this as my own work **as-is**? **No** — the race and dual collection paths are the kind of issues that surface as flaky terminal UX and silent behavior drift. After mutex serialization (or single refresh loop), one shared board collector, and at least one integration smoke for first frame/cancel, this becomes a clean Approve for a focused CLI feature.

**Verdict: Request Changes** — address findings 1–3 before merge; 4 strongly recommended in the same PR; 5 can ride along with the dedupe.

obey-agent:
## Addressed review feedback

- **Render race:** hybrid fsnotify + ticker paints through one mutex (full clear+write frame).
- **Footer:** hybrid path always shows polling interval (matches the always-on 2s ticker).
- **Collection dedupe:** shared `collect*Board` model for one-shot, JSON, and `--watch`; board builders extracted to `list_board.go`.

Unit tests for list package green.

PR #264, opened 2026-07-13
feat(workflow): async approval judge (fire-and-forget)

lancekrogers:
## Summary

Fire-and-forget auto-judge so `fest next` / `fest workflow approve --auto` does not block the TTY while a long-running judge evaluates a checkpoint.

- Default `--auto` launches a detached `workflow judge-exec` runner, records `BeginJudge` with PID, and returns immediately
- Checkpoint stays blocked; `fest show` / watchers can render **waiting-on-judge**
- New `--wait` flag restores the previous blocking path
- Duplicate launch while the recorded judge PID is still alive is rejected; stale PIDs (crash/reboot) can relaunch
- **Safety:** refuse to re-exec `*.test` / go test package binaries as the judge runner (this path previously fork-bombed machines under `go test`)
- Unit tests mock `launchJudgeProcess` so package tests never spawn real detached judges

## Worktree

`fest@fest-async-judge` → `projects/worktrees/fest/fest-async-judge`

## Test plan

- [ ] `go test` only with mocked launch paths (do **not** run unmocked auto-judge tests on older WIP without the go-test binary guard)
- [ ] Manual: `fest workflow approve --auto` on a gate → returns immediately, progress shows judge running
- [ ] Manual: `fest show` / waiting-on-judge UI while runner alive
- [ ] Manual: judge completes → `fest next` continues or stays blocked on reject
- [ ] Manual: `--wait` still blocks until verdict
- [ ] Confirm no orphan `workflow.test` / `judge-exec` processes after tests

## Notes / risk

This is a **draft** backup of in-progress work recovered after a machine freeze caused by process storms from detached `workflow.test` re-exec under `go test`. Review carefully before merge.

### DIFF
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
@@ -25,6 +26,7 @@ type approvalJudgeOptions struct {
 	Auto         bool
 	JudgeCommand string
 	Timeout      time.Duration
+	Wait         bool
 }
 
 type approvalJudgeRequest struct {
@@ -89,7 +91,12 @@ Auto approval:
 
       hooks:
         approval_judge:
-          command: ob judge`,
+          command: ob judge
+
+  By default --auto launches the judge in the background and returns
+  immediately; the checkpoint stays blocked until the verdict lands, and
+  'fest show' renders the waiting-on-judge state while it runs. Use --wait
+  to block until the judge returns instead.`,
 		Annotations: map[string]string{
 			"scope": string(scope.Festival),
 		},
@@ -111,6 +118,7 @@ Auto approval:
 	cmd.Flags().BoolVar(&opts.Auto, "auto", false, "delegate this checkpoint decision to the configured approval judge command")
 	cmd.Flags().StringVar(&opts.JudgeCommand, "judge-command", opts.JudgeCommand, "approval judge command for --auto (overrides the .festival/config.yaml hooks.approval_judge.command hook)")
 	cmd.Flags().DurationVar(&opts.Timeout, "judge-timeout", opts.Timeout, "maximum time to wait for the approval judge (0 waits until it returns)")
+	cmd.Flags().BoolVar(&opts.Wait, "wait", false, "block until the judge returns instead of launching it in the background")
 	cmd.MarkFlagsMutuallyExclusive("auto", "as")
 	cmd.MarkFlagsMutuallyExclusive("auto", "summary")
 
@@ -315,25 +323,45 @@ func AutoDelegateBlockingCheckpoints(ctx context.Context, nav *wf.Navigator) err
 }
 
 func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, step wf.WorkflowStep, opts approvalJudgeOptions) error {
+	// A judge may already be evaluating this checkpoint from a prior --auto.
+	// Refuse a duplicate launch while it is alive; a running record whose
+	// process died (crash, reboot) is stale and may be relaunched.
+	if ss := nav.GetWorkflowState().GetStepState(currentStepNum); ss != nil && ss.Judge != nil &&
+		ss.Judge.Status == wf.JudgeRunning && judgeProcessAlive(ss.Judge.Pid) {
+		return festerrors.Validation("an approval judge is already evaluating this checkpoint").
+			WithField("judge_command", ss.Judge.Command).
+			WithField("pid", fmt.Sprintf("%d", ss.Judge.Pid)).
+			WithHint("the checkpoint unblocks when the judge returns; watch it with 'fest show' and re-run 'fest next' afterwards")
+	}
+
+	if !opts.Wait {
+		return launchApproveAuto(ctx, nav, currentStepNum, step, opts)
+	}
+
+	// --wait: run the judge in-process and block until the verdict lands.
 	// Record the judge invocation before running it so watchers can render
 	// the waiting-on-judge state and a hung or crashed judge leaves a trace.
-	if err := nav.BeginJudge(ctx, opts.JudgeCommand); err != nil {
+	if err := nav.BeginJudge(ctx, currentStepNum, opts.JudgeCommand, os.Getpid()); err != nil {
 		return festerrors.Wrap(err, "recording judge start")
 	}
 
 	decision, audit, err := judgeApproval(ctx, nav, step, opts)
 	if err != nil {
-		if recErr := nav.RecordJudgeOutcome(ctx, wf.JudgeFailed, err.Error()); recErr != nil {
+		if recErr := nav.RecordJudgeOutcome(ctx, currentStepNum, wf.JudgeFailed, err.Error()); recErr != nil {
 			fmt.Printf("%s failed to record judge outcome: %v\n", ui.Warning("⚠"), recErr)
 		}
 		return err
 	}
 
+	return applyApproveAutoVerdict(ctx, nav, currentStepNum, step, decision, audit)
+}
+
+func applyApproveAutoVerdict(ctx context.Context, nav *wf.Navigator, currentStepNum int, step wf.WorkflowStep, decision *approvalJudgeResponse, audit string) error {
 	judgeDecision := wf.DecisionMetadata{Actor: decisionActorAgent, Summary: decision.Reason}
 
 	switch decision.Decision {
 	case "approve":
-		if err := nav.RecordJudgeOutcome(ctx, wf.JudgeApproved, decision.Reason); err != nil {
+		if err := nav.RecordJudgeOutcome(ctx, currentStepNum, wf.JudgeApproved, decision.Reason); err != nil {
 			return festerrors.Wrap(err, "recording judge outcome")
 		}
 		if err := nav.ApproveWithAudit(ctx, audit, judgeDecision); err != nil {
@@ -343,7 +371,7 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		fmt.Printf("  %s: %s\n", ui.Label("Reason"), decision.Reason)
 		return showNextStep(ctx, nav, nav.GetSteps())
 	case "reject":
-		if err := nav.RecordJudgeOutcome(ctx, wf.JudgeRejected, decision.Reason); err != nil {
+		if err := nav.RecordJudgeOutcome(ctx, currentStepNum, wf.JudgeRejected, decision.Reason); err != nil {
 			return festerrors.Wrap(err, "recording judge outcome")
 		}
 		if err := nav.RejectWithDecision(ctx, audit, judgeDecision); err != nil {
@@ -363,7 +391,7 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		unsupported := festerrors.Validation("approval judge returned unsupported decision").
 			WithField("decision", decision.Decision).
 			WithHint("allowed decisions are approve and reject")
-		if recErr := nav.RecordJudgeOutcome(ctx, wf.JudgeFailed, unsupported.Error()); recErr != nil {
+		if recErr := nav.RecordJudgeOutcome(ctx, currentStepNum, wf.JudgeFailed, unsupported.Error()); recErr != nil {
 			fmt.Printf("%s failed to record judge outcome: %v\n", ui.Warning("⚠"), recErr)
 		}
 		return unsupported
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -299,6 +299,22 @@ func AutoDelegateBlockingCheckpoints(ctx context.Context, nav *wf.Navigator) err
 			return nil
 		}
 
+		// fest next must not error while a prior fire-and-forget judge is still
+		// alive — report waiting-on-judge and leave the checkpoint blocked.
+		if stepState.Judge != nil && stepState.Judge.Status == wf.JudgeRunning &&
+			judgeProcessAlive(stepState.Judge.Pid) {
+			cmd := stepState.Judge.Command
+			if cmd == "" {
+				cmd = judgeCommand
+			}
+			fmt.Printf("%s Approval judge still running: %s (pid %d)\n",
+				ui.Value("⚖", ui.JudgeColor), cmd, stepState.Judge.Pid)
+			fmt.Println("  Checkpoint stays blocked until the verdict lands.")
+			fmt.Printf("  Watch progress: %s\n", ui.Accent("fest workflow status"))
+			fmt.Printf("  After it returns, run %s again.\n", ui.Accent("fest next"))
+			return nil
+		}
+
 		fmt.Printf("%s Checkpoint delegated to approval judge (%s)\n", ui.Info("→"), judgeCommand)
 		fmt.Printf("  Step %d: %s\n", current, step.Name)
 
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -182,7 +182,7 @@ func runApproveWithOptions(ctx context.Context, decision wf.DecisionMetadata, op
 	}
 
 	return withJudgeStepLock(ctx, nav.Ctx.PhasePath, currentStepNum, func() error {
-		fresh, err := getWorkflowNavigator(ctx)
+		fresh, err := reloadWorkflowNavigator(ctx, nav)
 		if err != nil {
 			return err
 		}
@@ -359,7 +359,7 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		return err
 	}
 	err = withJudgeStepLock(ctx, nav.Ctx.PhasePath, currentStepNum, func() error {
-		fresh, err := getWorkflowNavigator(ctx)
+		fresh, err := reloadWorkflowNavigator(ctx, nav)
 		if err != nil {
 			return err
 		}
@@ -382,7 +382,7 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 
 	decision, audit, err := judgeApproval(ctx, nav, step, opts)
 	if err != nil {
-		if recErr := recordJudgeFailureIfOwned(ctx, nav.Ctx.PhasePath, judgeExecPayload{
+		if recErr := recordJudgeFailureIfOwned(ctx, nav, judgeExecPayload{
 			StepNumber: currentStepNum, RunID: runID,
 		}, err.Error()); recErr != nil {
 			fmt.Printf("%s failed to record judge outcome: %v\n", ui.Warning("⚠"), recErr)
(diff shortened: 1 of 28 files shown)


internal/commands/workflow/judge_exec.go
```diff
@@ -0,0 +1,227 @@
...
+	payloadPath := filepath.Join(festDir, fmt.Sprintf("judge-step-%d.json", currentStepNum))
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return festerrors.Wrap(err, "encoding judge-exec payload")
+	}
+	if err := os.WriteFile(payloadPath, data, 0o600); err != nil {
+		return festerrors.IO("writing judge-exec payload", err).WithField("path", payloadPath)
+	}
+
+	pid, err := launchJudgeProcess(payloadPath, nav.Ctx.PhasePath, filepath.Join(festDir, "judge.log"))
```
obey-agent: Blocking race: the detached child is started before `BeginJudge` is durably recorded. An instant judge can append its outcome/advance events before the parent appends `wf_judge_started`; replay then leaves the completed checkpoint marked `running` with a dead PID. Worse, if `BeginJudge` fails, this function returns an error while the already-detached child can still approve the checkpoint. Introduce a durable per-run lease/ID before the child is allowed to evaluate (for example a starting record plus child handshake), update it with the PID, and terminate/refuse the child if registration fails. Add a deterministic fast-child interleaving test.

internal/commands/workflow/judge_exec.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// The judge may have run for a long time; re-read durable state and only
+	// apply the verdict if the delegated checkpoint is still the current one
+	// (an operator may have approved or reset it manually in the meantime).
+	fresh, err := getWorkflowNavigator(ctx)
+	if err != nil {
+		return err
+	}
+	if fresh.GetWorkflowState().CurrentStep != payload.StepNumber {
```
obey-agent: Blocking ownership issue: reloading and checking only `CurrentStep` does not prove this detached runner still owns the decision. A manual reject keeps the same step number, and resetting a step-1 workflow also keeps step 1; in both cases the old child can subsequently overwrite the operator decision. A stale runner can likewise race a relaunched runner. Persist a unique judge run ID in the payload/events/state and compare-and-set on `status=running` plus the same run ID before recording or applying any verdict; manual approve/reject/reset must invalidate that lease.

internal/commands/workflow/approve.go
```diff
@@ -315,25 +339,45 @@ func AutoDelegateBlockingCheckpoints(ctx context.Context, nav *wf.Navigator) err
 }
 
 func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, step wf.WorkflowStep, opts approvalJudgeOptions) error {
+	// A judge may already be evaluating this checkpoint from a prior --auto.
+	// Refuse a duplicate launch while it is alive; a running record whose
+	// process died (crash, reboot) is stale and may be relaunched.
+	if ss := nav.GetWorkflowState().GetStepState(currentStepNum); ss != nil && ss.Judge != nil &&
```
obey-agent: Blocking concurrency issue: this duplicate check is a non-atomic read followed later by payload write/process launch/state append. Two simultaneous `fest workflow approve --auto` invocations can both observe no live judge and launch competing children (also writing the same `judge-step-N.json`). Serialize launch per checkpoint with a filesystem lock or durable compare-and-set lease; the ownership token should also make only one verdict eligible to land. Add a concurrent-launch integration test.

internal/commands/workflow/judge_exec_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if err == nil {
+		t.Fatal("expected refusal when executable is a go test binary")
+	}
+	if !strings.Contains(err.Error(), "go test binary") {
+		t.Fatalf("error = %v, want go test binary refusal", err)
+	}
+}
+
+func TestRunApproveAuto_AsyncFireAndForget(t *testing.T) {
+	dir := setupWorkflowFestival(t)
```
obey-agent: These new host unit tests create workflow directories and write payload/state files through `setupWorkflowFestival` and `launchApproveAuto`. The repository-wide filesystem-safety contract requires real filesystem mutation coverage to run in the container harness. Please keep pure payload/guard logic as host unit tests, but move the async launch/state-write cases into `tests/integration/async_judge_test.go` (which already provides the correct isolation boundary).

internal/commands/workflow/judge_exec.go
obey-agent: Blocking: the detached child is started before `BeginJudge` is durably saved. An instant child can apply an approved/rejected outcome and then have that terminal judge state overwritten back to `running` by line 74. If `BeginJudge` fails, the function returns an error while an untracked child is still free to mutate the checkpoint. Please establish durable run ownership before the child can evaluate (for example, a persisted run ID/state plus a child handshake), and make launch failure resolve that same run. Add a deterministic test for a child completing before the parent finishes its launch bookkeeping.

internal/commands/workflow/judge_exec.go
obey-agent: Blocking: `CurrentStep` is not enough to prove this runner still owns the decision. A manual rejection leaves `CurrentStep` unchanged, so a later async approval reaches line 226 and overwrites the operator decision. A reset/relaunch that returns to the same step, or concurrent launches that race before `BeginJudge`, has the same stale-verdict problem. Persist a unique judge run ID and compare-and-set the verdict only while that exact run is still `running`; otherwise discard it without changing the newer/manual judge state. Please cover manual reject and superseded-run cases.

obey-agent (commented):
Verdict: Request Changes

Overview: This is a useful direction: fire-and-forget judging, durable lifecycle visibility, a blocking compatibility path, and container-isolated process coverage are all appropriate for the workflow surface. The current ownership protocol is not safe enough to let a detached process mutate checkpoint state, however; it can lose or overwrite decisions under realistic process and operator races.

Key Findings:
- Blocking: the child starts before the parent persists `BeginJudge`. A fast child can finish first and have its terminal state overwritten to `running`; a persistence failure leaves an untracked child running. See the inline comment on `launchJudgeProcess`.
- Blocking: the child validates only `CurrentStep`, not ownership of the active judge run. A manual rejection, reset back to the same step, relaunch, or concurrent launch can be overwritten by a stale verdict. See the inline comment on the final state check.
- Merge gate: `internal/guidance/workflow/navigator.go` is not gofmt-clean.
- The PR currently has no reported GitHub checks.
- Verification: the targeted container suite `TestAsyncJudge_*` passed (4 tests). Those tests cover happy-path detach/wait/visibility, but not the two conflicting-transition cases above.

What's Done Well:
- The Go test binary refusal directly addresses the previously observed process-storm failure.
- Real detach/re-exec behavior is kept in the container harness, with host unit tests using a launch seam.
- Explicit step attribution for judge events is an improvement over implicitly recording against whatever step is current.

Staff Standard: I would be proud of the direction and the safety-conscious test isolation, but I would not submit or merge this version until judge ownership is durable and verdict application is compare-and-set against the exact active run.

obey-agent (commented):
Verdict: Request Changes

Overview: The async judge direction is valuable and the draft contains thoughtful safety recovery after the earlier process storm: test-binary refusal, containerized real-process coverage, explicit step targeting, and waiting-state visibility. The current launch/apply protocol is not concurrency-safe yet, so it can leave false running state or allow an obsolete judge to override an operator/newer run.

Key Findings:
- Blocking: the child starts before the durable judge-start record. A fast child can finish first, and a start-record failure does not stop the detached child from applying a verdict.
- Blocking: verdict ownership is checked only by step number. Manual reject/reset on the same step or a newer relaunch can be overwritten by the old runner.
- Blocking: duplicate prevention is not atomic; concurrent invocations can both launch and race on the same payload/state.
- New filesystem-mutating async unit tests should move behind the existing container integration boundary.

What’s Done Well:
- The explicit `.test`/`.test.exe` guard directly addresses the known recursive re-exec failure.
- PID, judge lifecycle events, JSON status, show output, and checkpoint guidance are wired through one existing state/event model.
- All four containerized `TestAsyncJudge_*` tests pass; targeted compile/vet and diff checks are clean.

Staff Standard: Not yet. I would be comfortable submitting this after introducing a durable per-run lease/ID, atomic launch ownership, and compare-and-set verdict application with adversarial concurrency/manual-override tests.

internal/guidance/workflow/navigator.go
```diff
@@ -268,17 +268,17 @@ func (n *Navigator) GetContextFiles(ctx context.Context) ([]string, error) {
 // the current step, before the judge command executes, so concurrent watchers
 // (fest show --watch, fest progress) can render the waiting-on-judge state
 // and a crashed or timed-out judge still leaves a trace.
-func (n *Navigator) BeginJudge(ctx context.Context, command string) error {
-	return n.recordJudge(ctx, EmitJudgeStartedEvents(n.stateKey(), n.workflowState.CurrentStep, command), func(at time.Time) {
-		n.workflowState.BeginJudge(command, at)
+func (n *Navigator) BeginJudge(ctx context.Context, step int, command string, pid int) error {
```
obey-agent: Merge gate for this file: `gofmt -d internal/guidance/workflow/navigator.go` reports a non-canonical standard-library import order (`time` needs to sort after `path/filepath`). Please run the repository formatting gate and commit the normalized file.

internal/commands/workflow/judge_exec.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+	lockPath := filepath.Join(lockDir, fmt.Sprintf("judge-step-%d.lock", step))
+	lockFile, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0o600)
+	if err != nil {
+		return festerrors.IO("opening judge step lock", err).WithField("path", lockPath)
+	}
+	defer func() { _ = lockFile.Close() }()
+
+	for {
+		err = syscall.Flock(int(lockFile.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
```
obey-agent: Blocking portability regression: `just build windows` fails on this head because `syscall.Flock`/`LOCK_*` are unavailable on Windows, and the same build also rejects `SysProcAttr{Setsid: true}` below. `fest` has an explicit Windows build target, so this PR cannot meet the repository build contract as written. Please move lock/process-detach primitives behind platform-specific files (with an appropriate Windows implementation and tests), or make an explicit product decision to drop/disable Windows support before merge. The current gate fails with six compiler errors.

obey-agent (commented):
Verdict: Request Changes

Overview: Commit `94c649c` does address every finding from the prior review; the old anchors are outdated and those seven threads are now resolved. The concurrency design is materially stronger, but a fresh repository-level gate found a new portability blocker, so this is not yet merge-ready or staff-quality across fest's supported build surface.

Previously Requested Changes — Addressed:
- Launch ordering: the parent persists a run lease before process start, claims the PID durably, and the child waits for that claim before evaluation.
- Stale verdict ownership: payload, events, and state carry a unique run ID; verdict application is compare-and-set against the active run, and manual decisions cancel it.
- Duplicate launch: a per-checkpoint cross-process lock serializes launch/transition state; the concurrent integration test proves only one runner starts.
- Filesystem safety: host tests retain only pure guard checks; real payload/state/process coverage is in the container harness.
- Formatting: the changed Go files are gofmt-clean.

New Blocking Finding:
- `just build windows` fails with six compiler errors because `syscall.Flock`/lock constants and `SysProcAttr.Setsid` are Unix-only. See the new inline comment. Fest exposes an explicit Windows build target, so platform-specific process/lock implementations or an explicit support decision are required before merge.

Verification:
- Six uncached containerized `TestAsyncJudge_*` tests pass, including fast-child ordering, concurrent launch, and manual override.
- golangci-lint, go vet, full package compilation, local build, formatting, and diff checks pass on the current host.
- Windows build fails.

What's Done Well:
- The lease/claim/return protocol is explicit and event replay rejects stale run IDs.
- Operator decisions remain authoritative.
- Adversarial behaviors are tested at the correct container boundary.

Staff Standard: Not yet. I would be proud of the concurrency correction, but I would not submit or merge a fest change that breaks an advertised platform build.

internal/commands/workflow/approve.go
```diff
@@ -315,45 +350,84 @@ func AutoDelegateBlockingCheckpoints(ctx context.Context, nav *wf.Navigator) err
...
+	if !opts.Wait {
+		return launchApproveAuto(ctx, nav, currentStepNum, step, opts)
+	}
+
+	runID, err := newJudgeRunID()
+	if err != nil {
+		return err
+	}
+	err = withJudgeStepLock(ctx, nav.Ctx.PhasePath, currentStepNum, func() error {
+		fresh, err := getWorkflowNavigator(ctx)
```
obey-agent: Blocking regression: four package unit tests fail on this head because the wait path re-resolves state via `getWorkflowNavigator(ctx)`, which depends on CWD / scope festival injection rather than the `nav` already in hand:

- `TestRunApproveAuto_ApproveAdvancesAndRecordsAudit`
- `TestRunApproveAuto_RejectBlocksStepWithAudit`
- `TestRunApproveAuto_RecordsJudgeLifecycle`
- `TestRunApproveAuto_JudgeFailureLeavesDurableTrace`

They call `runApproveAuto` with a temp-dir navigator and no chdir, so BeginJudge / apply never land (`not in a festival: festival root ... not found`). Fail-closed error recording (`recordJudgeFailureIfOwned`) has the same dependency, so a failed judge can leave **no durable trace** when resolution fails.

Fresh reload under the lock is the right concurrency idea, but reload from the **already-resolved phase path** (or inject festival/phase into ctx from `nav.Ctx`), not from ambient CWD detection. Package tests should pass without requiring callers to chdir into the fixture.

internal/commands/workflow/judge_exec.go
obey-agent: Blocking portability: `syscall.Flock` / `LOCK_EX` / `LOCK_NB` / `LOCK_UN` are Unix-only. `GOOS=windows go build ./internal/commands/workflow` fails here.

This repo already has the correct pattern in `internal/navigation/lock_unix.go` + `lock_other.go` (`//go:build unix` / `!unix`). Split the judge lock the same way — either real Windows file locking, or an explicit non-Unix fallback with documented race semantics — so `just build windows` keeps working.

internal/commands/workflow/judge_exec.go
```diff
@@ -0,0 +1,414 @@
...
+		return 0, festerrors.IO("opening judge log", err).WithField("path", logPath)
+	}
+	defer func() { _ = logFile.Close() }()
+
+	cmd := exec.Command(exe, "workflow", "judge-exec", "--payload", payloadPath)
+	cmd.Dir = phaseDir
+	cmd.Stdout = logFile
+	cmd.Stderr = logFile
+	// New session so the runner outlives the parent fest process (fire-and-forget).
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
```
obey-agent: Blocking portability: `SysProcAttr{Setsid: true}` is not available on Windows (same `GOOS=windows` build failure). Detach/session creation must live behind a unix build tag (or use Windows job/creation-flags equivalent). Without this, fire-and-forget cannot compile on an advertised fest platform.

internal/commands/workflow/judge_exec.go
```diff
@@ -0,0 +1,414 @@
...
+// relaunching the judge forever.
+func judgeProcessAlive(pid int) bool {
+	if pid <= 0 {
+		return false
+	}
+	proc, err := os.FindProcess(pid)
+	if err != nil {
+		return false
+	}
+	return proc.Signal(syscall.Signal(0)) == nil
```
obey-agent: Portability follow-on once the Windows build compiles: `proc.Signal(syscall.Signal(0))` is not a reliable liveness probe on Windows (`FindProcess` always succeeds; `Signal` is unsupported). That would make every recorded PID look dead (`judgeLeaseActive` → false), so duplicate-launch protection collapses and relaunches can storm.

Provide a platform-specific `judgeProcessAlive` (unix: signal 0; windows: OpenProcess / equivalent), co-located with the lock/detach split.

obey-agent (commented):
## Verdict: Request Changes

### Overview
Fire-and-forget approval judging is the right product move, and commit `94c649c` made the ownership protocol real: durable run leases, PID claim-before-evaluate, CAS verdict application, cross-process step locks, and operator cancel that supersedes stale runners. That concurrency work addresses the earlier review. This head is still not merge-ready: it breaks the advertised Windows build, and four workflow package unit tests fail on the wait path’s ambient navigator re-resolve.

### Key Findings
- **Blocking — Windows build**: `GOOS=windows go build ./internal/commands/workflow` fails on `syscall.Flock`/`LOCK_*` and `SysProcAttr.Setsid`. Fest ships `just build windows` / release windows artifacts; mirror the existing `internal/navigation/lock_{unix,other}.go` pattern (and a detach helper) before merge. See inline on `judge_exec.go`.
- **Blocking — process liveness portability**: even after compile fixes, `Signal(0)` is not a Windows-safe alive check; lease protection would misclassify every PID as dead. Platform-specific `judgeProcessAlive` is required with the same split.
- **Blocking — unit tests / durable failure path**: `go test ./internal/commands/workflow` fails four `TestRunApproveAuto_*` cases. The wait path and failure recording re-call `getWorkflowNavigator(ctx)` (CWD/scope) instead of reloading from the navigator/phase already in hand, so BeginJudge/outcome never persist outside a live CLI cwd. See inline on `approve.go:362`.
- **Prior concurrency findings**: launch-before-lease, step-only ownership, non-atomic duplicate launch, and host filesystem process tests are **addressed** on this head. Adversarial container cases (`TestAsyncJudge_ConcurrentLaunch*`, `ManualRejectSupersedes*`, event ordering for fast children) are the right boundary.
- **No GitHub checks** reported on the branch yet.

### What's Done Well
- Go-test binary refusal and container-only real re-exec coverage directly prevent the process-storm that froze the machine.
- Run ID + claim + CAS + manual cancel is the correct ownership model; event replay ignores stale `wf_judge_returned` for foreign run IDs.
- Waiting-on-judge visibility is wired through status/show/JSON and checkpoint guidance so agents know not to spin-relaunch.
- Default fire-and-forget with explicit `--wait` preserves the old blocking contract.

### Staff Standard
Not yet. I would be proud of the async design and the concurrency correction, but I would not submit or merge a fest change that fails package unit tests and breaks an advertised platform build. Fix Windows process/lock abstractions, make wait-path reload phase-scoped (not CWD-scoped), and green `go test ./internal/commands/workflow` + `just build windows` before re-review.

obey-agent (approved):
Approved — this is now staff-quality and merge-ready.

The blocking findings are addressed in the implementation, not just discussed:
- Durable run IDs, per-step cross-process locking, lease/PID claim handshake, ownership-checked verdict application, and safe manual supersession close the concurrency/process races.
- Locked reloads derive from the already-resolved navigator context, preserve the injected persistence backend and navigator identity, and no longer depend on ambient CWD/phase discovery.
- Locking, detachment, and PID liveness are implemented and tested per platform; Linux, macOS, and Windows builds/vet/compilation pass.
- Filesystem/process behavior is exercised in containers, including concurrent launch, manual supersession, wait behavior, fest-next delegation, and real fire-and-forget verdict application.

Verified on head 8bc069f:
- just build all
- just lint all (0 issues)
- go vet for current, windows/amd64, darwin/arm64, and linux/amd64
- cross-platform workflow package test compilation
- full containerized internal/commands/workflow suite
- all six uncached TestAsyncJudge_* container integration tests

All review threads are resolved. GitHub reports the PR mergeable/CLEAN; this repository currently exposes no required remote status checks on the PR.

PR #266, opened 2026-07-13
fix(workflow): harden auto-judge authority and readiness

lancekrogers:
## Summary

Hardens the festival auto-judge product path after CC0001-style failures where:

1. Agents ran `approve --auto` before presentation artifacts existed
2. Agents cleared judge rejects with manual `approve --summary "…"` recorded as `decision_actor=user`
3. Operator-validation checkpoints were treated like artifact reviews

### Changes (PR1 of auto-judge hardening design)

- **Checkpoint classes** — parse `**Checkpoint class:**` and classify via heuristics (`artifact_review` vs `operator_attestation`)
- **`--auto` refuses `operator_attestation`** — no model call for “did the user validate?”
- **Deterministic readiness** — PRESENT-like steps require non-empty `output_specs/PRESENTATION.md` (or listed **Evidence:**) before the judge runs
- **Authority** — when `hooks.approval_judge` is configured, manual approve requires an interactive TTY + typed `APPROVE`; non-interactive agents fail closed. After a reject, actor is `user_override`. Scripted operator path: `--override-judge --summary "…"`
- **Resubmit UX** — reject/advance/approve messages tell agents to re-run `approve --auto`, not `advance`
- **Judge cwd** — judge process runs with `Dir = festival path`

Design: `workflow/design/fest-auto-judge-hardening/README.md` (campaign)

## Test plan

- [x] `go test ./internal/guidance/workflow/ ./internal/commands/workflow/`
- [x] Unit coverage for class, readiness, TTY/override, WorkDir
- [ ] Manual: PRESENT without PRESENTATION.md → readiness block, no judge
- [ ] Manual: after judge reject, non-TTY `fest workflow approve --summary "…"` fails when auto-judge configured
- [ ] Manual: `operator_attestation` gate → `--auto` refused

## Follow-ups

- PR2: multi-file evidence pack (`fest.approval.judge/v2`)
- PR3: `ob judge` multi-evidence render
- PR4: template annotations for ingest/plan gates

### DIFF
--- a/internal/commands/workflow/approval_hardening_test.go
+++ b/internal/commands/workflow/approval_hardening_test.go
@@ -170,22 +170,36 @@ func TestResolveManualApprovalDecision_JudgeConfiguredRequiresTTY(t *testing.T) 
 	}
 }
 
-func TestResolveManualApprovalDecision_PriorJudgeRejectRequiresTTYWithoutHook(t *testing.T) {
+func TestResolveManualApprovalDecision_DurableJudgeRejectRequiresTTYWithoutHook(t *testing.T) {
 	orig := stdinIsInteractiveFn
 	stdinIsInteractiveFn = func() bool { return false }
 	t.Cleanup(func() { stdinIsInteractiveFn = orig })
 
-	blocked := &wf.StepState{
-		Status:        wf.StepStatusBlocked,
-		DecisionActor: decisionActorAgent,
-		Judge:         &wf.JudgeState{Status: wf.JudgeRejected},
+	tests := []struct {
+		name    string
+		command string
+	}{
+		{name: "one-off judge command", command: "custom-judge --strict"},
+		{name: "workspace hook removed after rejection", command: "ob judge"},
 	}
-	_, err := resolveManualApprovalDecision(
-		wf.DecisionMetadata{Actor: decisionActorUser},
-		false, false, false, blocked, 4, "PRESENT",
-	)
-	if err == nil || !strings.Contains(err.Error(), "interactive operator TTY") {
-		t.Fatalf("prior reject without current hook must remain protected: %v", err)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			blocked := &wf.StepState{
+				Status:        wf.StepStatusBlocked,
+				DecisionActor: decisionActorAgent,
+				Judge: &wf.JudgeState{
+					Status:  wf.JudgeRejected,
+					Command: tt.command,
+				},
+			}
+			_, err := resolveManualApprovalDecision(
+				wf.DecisionMetadata{Actor: decisionActorUser},
+				false, false, false, blocked, 4, "PRESENT",
+			)
+			if err == nil || !strings.Contains(err.Error(), "interactive operator TTY") {
+				t.Fatalf("durable reject without current hook must remain protected: %v", err)
+			}
+		})
 	}
 }
 
--- a/internal/commands/workflow/approval_hardening_test.go
+++ b/internal/commands/workflow/approval_hardening_test.go
@@ -2,12 +2,13 @@ package workflow
 
 import (
 	"context"
-	"os"
-	"path/filepath"
+	"errors"
 	"strings"
 	"testing"
 
+	"github.com/Obedience-Corp/fest/internal/config"
 	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
+	"github.com/Obedience-Corp/fest/internal/scope"
 )
 
 func TestCheckAutoJudgeAllowed_OperatorAttestation(t *testing.T) {
@@ -28,13 +29,14 @@ func TestCheckAutoJudgeAllowed_OperatorAttestation(t *testing.T) {
 }
 
 func TestCheckApprovalReadiness_PresentMissingPresentation(t *testing.T) {
-	phase := t.TempDir()
 	step := wf.WorkflowStep{
 		Number:     4,
 		Name:       "PRESENT",
 		Checkpoint: wf.CheckpointUserApproval,
 	}
-	err := checkApprovalReadiness(phase, step)
+	err := checkApprovalReadinessWithInspector("/phase", step, func(_, _ string) (bool, error) {
+		return false, nil
+	})
 	if err == nil {
 		t.Fatal("expected readiness failure")
 	}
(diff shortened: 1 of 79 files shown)


internal/commands/workflow/operator_approve.go
```diff
@@ -0,0 +1,144 @@
...
+	stepNum int,
+	stepName string,
+) (wf.DecisionMetadata, error) {
+	if decision.Actor == "" {
+		decision.Actor = decisionActorUser
+	}
+
+	priorReject := stepHasOpenJudgeReject(stepState)
+
+	if !judgeConfigured {
```
obey-agent: [P1] Enforce the override boundary from durable step state, not only from the current hook configuration. A reject produced by `--auto --judge-command ...` has `priorReject=true` but no workspace hook; removing a hook after a reject has the same shape. This early return then lets a non-TTY agent run plain `approve`, clears the rejection, and merely labels its forged decision `user_override`. When `priorReject` is true, require interactive confirmation or `--override-judge` with rationale regardless of whether a hook is currently configured, and add both one-off-command and removed-hook regression tests.

internal/commands/workflow/approve.go
```diff
@@ -187,6 +213,13 @@ func runApproveWithOptions(ctx context.Context, decision wf.DecisionMetadata, op
 	return showNextStep(ctx, nav, steps)
 }
 
+// approvalJudgeConfigured reports whether the workspace has a judge hook
+// (flag-less resolution). Used to decide whether manual approve needs a TTY.
+func approvalJudgeConfigured(ctx context.Context) bool {
+	cmd, err := resolveApprovalJudgeCommand(ctx, "")
```
obey-agent: [P1] Do not collapse configuration load failures into “judge not configured.” `resolveApprovalJudgeCommand` can fail because `.festival/config.yaml` is malformed or unreadable; this boolean turns that into the legacy open-manual path, allowing non-interactive approval precisely when the authority configuration cannot be trusted. Return `(bool, error)` or inspect the config directly so absence is distinguishable from load/parse failure, propagate failures, and cover malformed config.

internal/guidance/workflow/parser.go
```diff
@@ -147,35 +158,160 @@ func parseActionItems(actionsBlock string) []string {
...
+		"confirm the user reviewed and approved",
+		"explicitly approve the plan",
+		"did the user explicitly approve",
+	}
+	for _, h := range attestationHints {
+		if strings.Contains(blob, h) {
+			return CheckpointClassOperatorAttestation
+		}
+	}
+	return CheckpointClassArtifactReview
```
obey-agent: [P1] This classification fails open: every unrecognized blocking checkpoint becomes `artifact_review` and is eligible for `--auto`. A human-only checkpoint such as “Obtain customer sign-off” misses this finite phrase list and gets delegated to the model. Authority checks need an allowlist: recognize positive artifact-review forms, but default ambiguous unannotated checkpoints to `operator_attestation` until templates carry explicit classes. Add an unknown human-attestation case to prove the default is safe.

internal/guidance/workflow/parser.go
```diff
@@ -147,35 +158,160 @@ func parseActionItems(actionsBlock string) []string {
...
+	if len(m) < 2 {
+		return CheckpointClassUnspecified
+	}
+	switch strings.ToLower(strings.TrimSpace(m[1])) {
+	case string(CheckpointClassArtifactReview), "artifact", "review":
+		return CheckpointClassArtifactReview
+	case string(CheckpointClassOperatorAttestation), "operator", "attestation", "user":
+		return CheckpointClassOperatorAttestation
+	default:
+		return CheckpointClassUnspecified
```
obey-agent: [P1] An explicit but invalid `Checkpoint class` annotation is silently converted to “unspecified,” after which the heuristic can auto-judge it. That defeats the explicit authority format and hides typos such as `operator_atestation`. Preserve whether the annotation was present and return a parse/validation error for unknown values; only genuinely absent annotations should enter legacy classification. Please add malformed-annotation coverage.

internal/commands/workflow/approval_readiness.go
```diff
@@ -0,0 +1,95 @@
...
+			continue
+		}
+		found++
+		base := filepath.Base(rel)
+		if strings.EqualFold(base, "PRESENTATION.md") {
+			presentationFound = true
+		}
+	}
+
+	if presentationRequired && !presentationFound {
```
obey-agent: [P1] The implementation does not honor the documented “PRESENTATION.md or listed Evidence” contract. For a PRESENT step with an explicit non-empty evidence list such as `output_specs/review.md`, `found` is positive but `presentationFound` remains false, so readiness still rejects it. Treat explicit `EvidencePaths` as the declared presentation pack (or require PRESENTATION.md everywhere and update the contract); add both alternate-pack and missing-pack tests.

internal/commands/workflow/approval_readiness.go
```diff
@@ -0,0 +1,95 @@
...
+	presentationRequired := wf.IsPresentationStep(step)
+	presentationFound := false
+
+	for _, rel := range paths {
+		rel = filepath.Clean(rel)
+		if rel == "." || strings.HasPrefix(rel, "..") {
+			continue
+		}
+		full := filepath.Join(phasePath, rel)
+		info, err := os.Stat(full)
```
obey-agent: [P1] This lexical containment check plus `os.Stat` follows symlinks, so `output_specs/PRESENTATION.md` can point outside the phase and satisfy readiness with an unrelated non-empty file. Because this check gates delegated approval, resolve each evidence path and prove it remains under the cleaned phase root, or reject symlinks with `Lstat`. Include an escape-symlink negative test in the isolated integration harness.

internal/commands/workflow/approval_hardening_test.go
```diff
@@ -0,0 +1,183 @@
...
+	if err == nil {
+		t.Fatal("expected readiness failure")
+	}
+	if !strings.Contains(err.Error(), "PRESENTATION.md") && !strings.Contains(err.Error(), "presentation") {
+		t.Fatalf("error = %v, want presentation mention", err)
+	}
+}
+
+func TestCheckApprovalReadiness_PresentWithFiles(t *testing.T) {
+	phase := t.TempDir()
```
obey-agent: [P1] These new readiness tests create directories and write evidence files through host `t.TempDir`. Campaign policy requires filesystem-mutating workflows to run in container isolation. Move the filesystem cases to the repository containerized integration harness, keeping only pure injected-filesystem cases in unit tests.

internal/commands/workflow/reject.go
```diff
@@ -143,7 +143,8 @@ func runRejectWithRemediationDecision(ctx context.Context, reason, remediationPh
 		fmt.Printf("  %s: %s\n", ui.Label("Summary"), decision.Summary)
 	}
 	fmt.Println("The step is now blocked. Address the feedback and revise the work.")
-	fmt.Println("When ready, run " + ui.Accent("fest workflow advance") + " to resubmit.")
+	fmt.Println("When ready, re-submit with: " + ui.Accent("fest workflow approve --auto"))
```
obey-agent: [P2] This unconditional recovery hint is invalid for two supported cases: with no configured judge, `approve --auto` fails immediately; for `operator_attestation`, it is deliberately refused. The blocked-step message in `advance` has the same issue. Render guidance from the checkpoint class and judge availability so users are only told to run commands that can succeed.

obey-agent (commented):
Verdict: REQUEST CHANGES

Overview: The durable judge lifecycle, working-directory fix, explicit operator confirmation, and readiness-before-model sequencing are strong directions. However, this authority hardening still fails open in several reachable paths, so it cannot safely ship as the boundary that prevents agents from clearing operator gates.

Key findings:
- A durable judge/readiness reject can be cleared non-interactively when the judge came from a one-off command or the workspace hook is later absent.
- Malformed or unreadable judge configuration is treated as no judge and opens the legacy manual-approval path.
- Ambiguous and malformed checkpoint classes default to auto-judgeable artifact review.
- Explicit alternate evidence packs are rejected despite the documented contract, while symlinked evidence can escape the phase boundary.
- New filesystem-mutating readiness tests run on the host instead of the required container harness.
- Recovery hints recommend `approve --auto` even where no judge exists or the checkpoint class forbids it.

What is done well: Judge execution now runs from the festival root, readiness is checked before model spend, reject/resubmit state is durable, and scripted overrides require a meaningful rationale.

Validation: I reviewed every changed file and the referenced runtime paths. I did not run the branch test suite from the main checkout; GitHub reports no checks for this head, and the changed filesystem tests must not be run directly on the host under campaign policy.

Staff standard: I would not submit this as production-ready until the authority decisions fail closed from durable state and configuration errors, and the evidence boundary is made explicit and contained.

obey-agent (commented):
Verdict: APPROVE

Overview: The authority and readiness hardening is now production-ready on the latest head. All eight original findings are addressed, and the branch has been reconciled with the asynchronous approval-judge architecture from main without weakening durable ownership, fail-closed behavior, or operator-only attestations.

Key findings:
- Durable judge/readiness rejects remain protected even after a one-off command or hook removal; malformed authority configuration fails closed.
- Checkpoint classification rejects invalid explicit values and defaults ambiguous human gates to operator attestation.
- Evidence packs honor explicit paths, require all listed files, contain symlinks within the phase, and are checked before both foreground and detached judge execution.
- Async launch and blocking wait paths use the step lock correctly, refuse duplicate live judges before readiness mutation, and recheck preflight in the detached child before model invocation.
- Recovery guidance now advertises only valid routes, and operator-attestation rendering never delegates merely because a workspace hook exists.
- All prior inline threads are resolved; I found no new actionable issue on head 490f4a9.

What is done well: The implementation extends the existing durable judge lifecycle rather than creating a parallel path, preserves run ownership when verdicts race with operator actions, keeps filesystem mutation coverage in the container harness, and documents the resulting CLI behavior clearly.

Validation: focused authority/classification tests, template rendering tests, all-package compile, just build quick, just lint all (0 issues), containerized readiness containment, and the complete async-judge integration suite all pass.

Staff standard: Yes. I would be proud to submit this as my own production change.

PR #271, opened 2026-07-14
feat(workflow): clarify judge state and retry commands

lancekrogers:
## Summary

- add `fest workflow judge` for explicit approval-judge runs and safe rejudging after judge-owned rejection
- show judge status and concise feedback consistently in `fest workflow status`, `fest show`, and `fest watch`
- remove judge command/PID/run/audit metadata from routine status payloads and normalize legacy feedback
- end blocked `fest next` instructions with the rejudge command when auto-judge recovery is available

## Behavior

A judge rejection remains blocked until the agent addresses the feedback. `fest workflow judge` reopens only judge-owned/readiness blocks, records a `wf_judge_recheck` event, and then runs the configured judge. Ordinary operator rejections and operator-attestation checkpoints remain human-owned. `fest workflow approve --auto` remains a backwards-compatible alias.

The machine-readable workflow status contract keeps its top-level `schema_version`; only noisy per-step judge execution metadata was removed.

## Validation

- `go test ./...`
- `go vet ./...`
- `just docs`
- `just build quick`
- visually checked the real blocked tcount gate with `workflow status`, `workflow show`, `show --inprogress`, and `workflow status --json`

Ready for review.

### DIFF
--- a/internal/commands/workflow/approval_guidance.go
+++ b/internal/commands/workflow/approval_guidance.go
@@ -32,6 +32,15 @@ func approvalRecoveryLinesFor(ctx context.Context, nav *wf.Navigator, step wf.Wo
 			"To enable auto-judge, configure hooks.approval_judge.command first.",
 		}
 	}
+	if nav != nil {
+		state := nav.GetWorkflowState()
+		stepState := state.GetStepState(step.Number)
+		if stepState != nil && stepState.Status == wf.StepStatusBlocked && !wf.IsJudgeRejection(stepState) {
+			return []string{
+				"Operator override: " + ui.Accent("fest workflow approve") + " (interactive)",
+			}
+		}
+	}
 	return []string{
 		"Operator override:       " + ui.Accent("fest workflow approve") + " (interactive)",
 		"Re-run approval judge:   " + ui.Accent("fest workflow judge"),
--- a/internal/commands/workflow/approval_guidance.go
+++ b/internal/commands/workflow/approval_guidance.go
@@ -33,8 +33,8 @@ func approvalRecoveryLinesFor(ctx context.Context, nav *wf.Navigator, step wf.Wo
 		}
 	}
 	return []string{
-		"Re-submit to the judge: " + ui.Accent("fest workflow approve --auto"),
 		"Operator override:       " + ui.Accent("fest workflow approve") + " (interactive)",
+		"Re-run approval judge:   " + ui.Accent("fest workflow judge"),
 	}
 }
 
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -25,6 +25,7 @@ const approvalJudgeSchemaVersion = "fest.approval.judge/v1"
 
 type approvalJudgeOptions struct {
 	Auto          bool
+	Rejudge       bool
 	JudgeCommand  string
 	Timeout       time.Duration
 	WorkDir       string // festival path; set as judge process cwd
@@ -79,8 +80,8 @@ Auto approval:
   delegates blocking checkpoints away from human review. With that hook set,
   fest next auto-invokes the judge on blocking WORKFLOW.md / GATES.md steps.
 
-  Use --auto to re-run the judge explicitly (for example after a reject or a
-  failed judge invocation). Agents must not clear checkpoints with --as agent;
+  Use 'fest workflow judge' to re-run the judge explicitly after a rejection;
+  '--auto' remains a backwards-compatible alias. Agents must not clear checkpoints with --as agent;
   agent-actor decisions are recorded only via the judge path.
 
   Checkpoint classes:
@@ -91,7 +92,7 @@ Auto approval:
   Presentation-like steps require non-empty evidence (e.g. output_specs/PRESENTATION.md)
   before the judge is invoked. Missing evidence blocks deterministically without a model call.
 
-  After a judge reject, re-submit with: fest workflow approve --auto
+  After a judge reject, re-submit with: fest workflow judge
   Operator override (interactive TTY, or --override-judge --summary "..."):
   records decision_actor=user_override.
 
@@ -120,6 +121,7 @@ Auto approval:
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if opts.Auto {
+				opts.Rejudge = true
 				return runApproveWithOptions(cmd.Context(), wf.DecisionMetadata{}, opts)
 			}
 			decision, err := normalizeDecision("approval", actor, opts.Summary)
@@ -502,6 +504,9 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		if fresh.GetWorkflowState().CurrentStep != currentStepNum {
 			return festerrors.Validation("checkpoint changed before approval judge started")
 		}
+		if err := reopenJudgeRejectionIfRequested(ctx, fresh, currentStepNum, opts); err != nil {
+			return err
+		}
 		freshSteps := fresh.GetSteps()
 		if currentStepNum < 1 || currentStepNum > len(freshSteps) {
 			return festerrors.Validation("checkpoint changed before approval judge started")
@@ -542,6 +547,20 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		_, err = applyApproveAutoVerdict(ctx, fresh, currentStepNum, step, runID, decision, audit)
 		return err
 	})
+}
+
+// reopenJudgeRejectionIfRequested clears only a judge-owned blocked state.
+// This keeps re-judging safe: an agent cannot use the judge command to bypass
+// an ordinary operator rejection or an operator-attestation checkpoint.
+func reopenJudgeRejectionIfRequested(ctx context.Context, nav *wf.Navigator, step int, opts approvalJudgeOptions) error {
+	if !opts.Rejudge {
+		return nil
+	}
+	state := nav.GetWorkflowState().GetStepState(step)
+	if state == nil || state.Status != wf.StepStatusBlocked {
+		return nil
+	}
+	return nav.ReopenJudgeRejection(ctx, step)
 }
 
 func applyApproveAutoVerdict(ctx context.Context, nav *wf.Navigator, currentStepNum int, step wf.WorkflowStep, runID string, decision *approvalJudgeResponse, audit string) (bool, error) {
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -504,20 +504,23 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int, 
 		if fresh.GetWorkflowState().CurrentStep != currentStepNum {
 			return festerrors.Validation("checkpoint changed before approval judge started")
 		}
-		if err := reopenJudgeRejectionIfRequested(ctx, fresh, currentStepNum, opts); err != nil {
-			return err
-		}
 		freshSteps := fresh.GetSteps()
 		if currentStepNum < 1 || currentStepNum > len(freshSteps) {
 			return festerrors.Validation("checkpoint changed before approval judge started")
 		}
 		step = freshSteps[currentStepNum-1]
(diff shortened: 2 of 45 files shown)


internal/guidance/workflow/state.go
```diff
@@ -99,6 +101,54 @@ type JudgeState struct {
...
+
+// IsJudgeRejection reports whether a blocked step was produced by an approval
+// judge or by the deterministic readiness checks that precede one. Ordinary
+// operator rejections intentionally return false and cannot be reopened by
+// the agent-facing judge command.
+func IsJudgeRejection(stepState *StepState) bool {
+	if stepState == nil || stepState.Status != StepStatusBlocked {
+		return false
+	}
+	if stepState.Judge != nil && stepState.Judge.Status == JudgeRejected {
```
obey-agent, line 172: ⚠️ [P1] Preserve operator rejection ownership when the Judge record is already rejected. RejectWithDecision does not clear a terminal JudgeRejected record, so after a judge rejection an operator can run fest workflow reject and change DecisionActor to user while Judge.Status remains rejected; this predicate still returns true and fest workflow judge can reopen that human-owned rejection. That bypasses the authority boundary stated in the command docs. Gate the JudgeRejected fallback on the current decision actor (or clear/replace the prior judge outcome on manual rejection), and add a regression test for judge rejection followed by operator rejection.

internal/commands/workflow/approve.go
```diff
@@ -544,6 +549,20 @@ func runApproveAuto(ctx context.Context, nav *wf.Navigator, currentStepNum int,
...
+// an ordinary operator rejection or an operator-attestation checkpoint.
+func reopenJudgeRejectionIfRequested(ctx context.Context, nav *wf.Navigator, step int, opts approvalJudgeOptions) error {
+	if !opts.Rejudge {
+		return nil
+	}
+	state := nav.GetWorkflowState().GetStepState(step)
+	if state == nil || state.Status != wf.StepStatusBlocked {
+		return nil
+	}
+	return nav.ReopenJudgeRejection(ctx, step)
```
obey-agent: ⚠️ [P1] Do not persist the re-open before auto-judge preflight succeeds. This helper clears the blocked state and saves wf_judge_recheck, but both the synchronous and detached paths call prepareAutoJudgeReadiness only afterward. If the checkpoint is changed to operator_attestation (or otherwise becomes non-auto-judgeable) after a prior judge rejection, fest workflow judge returns an error while leaving the step in_progress and erasing the previous feedback; fest next then no longer presents the blocked recovery state. Validate classification/readiness before mutating the rejection, or restore the blocked state on every preflight failure, with a regression test.

internal/commands/workflow/status_json.go
```diff
@@ -98,14 +94,10 @@ func collectWorkflowStatus(nav *wf.Navigator) workflowStatusJSON {
 		}
 		if stepState := state.GetStepState(step.Number); stepState != nil {
 			status = stepState.Status
-			feedback = stepState.Feedback
+			feedback = wf.DisplayFeedback(stepState.Feedback)
```
obey-agent, line 103: ⚠️ [P1] Version the status payload when removing judge fields. workflowStatusSchema remains fest.workflow.status/v1 and its comment promises a stable contract, but this PR removes judge_command, judge_pid, judge_run_id, and judge_detail from that same version. Existing JSON consumers will silently receive a different shape; moreover, failed-run detail is now not exposed anywhere in --json because it lives only in Judge.Detail. Retain deprecated fields, add a concise failure field, or publish this as a v2 schema before removing the old contract.

internal/guidance/workflow/state.go
```diff
@@ -99,6 +101,54 @@ type JudgeState struct {
...
+// complete judge audit string; keep accepting those records while presenting
+// only the actionable reason to users and agents.
+func DisplayFeedback(feedback string) string {
+	feedback = strings.TrimSpace(feedback)
+	if feedback == "" {
+		return ""
+	}
+
+	if strings.HasPrefix(feedback, "approval auto mode:") {
+		if idx := strings.LastIndex(feedback, " reason="); idx >= 0 {
```
obey-agent: ⚠️ [P2] Parse the legacy audit without using LastIndex on free-form reason text. A legitimate judge reason can itself contain the delimiter (for example, "missing proof because reason=was unclear"); LastIndex then treats the inner text as the field boundary, and strconv.Unquote fails or returns a truncated/mangled reason. Parse the structured audit fields or locate the final quoted reason value with an escape-aware parser, and add a test covering reason= inside the reason.

obey-agent (commented):
Verdict: Request Changes

Overview:
This is a thoughtful UX and lifecycle improvement: the explicit judge command, concise human output, durable event replay, and detached-run integration are coherent. The unit, vet, and async container suites pass, but the current authority and recovery behavior can reopen a human rejection or lose a blocked checkpoint, and the status JSON changes its v1 shape without versioning.

Key findings:
- P1: IsJudgeRejection still treats any JudgeRejected record as reopenable after an operator subsequently rejects the same checkpoint, allowing fest workflow judge to bypass the newer human-owned decision.
- P1: Rejudge clears/persists the blocked state before auto-judge preflight; a class/configuration failure can leave the step in_progress with feedback erased.
- P1: workflow status remains fest.workflow.status/v1 while removing judge fields, and failed judge detail disappears from JSON; retain compatibility or introduce a versioned payload with concise failure detail.
- P2: DisplayFeedback uses LastIndex on a delimiter that may appear inside a legacy reason, producing corrupted feedback for valid reasons containing reason=.

What is done well:
- Strong fail-closed detached judge ownership and lock/lease handling.
- Good separation between operator attestation and configured judge execution in the normal path.
- Concise feedback is consistently propagated through text renderers and event materialization.
- Tests cover the existing asynchronous process lifecycle and replay behavior well; please extend them for the authority and preflight cases above.

Staff standard:
I would be proud of the direction and most of the implementation, but I would not submit this revision as production-ready until the authority boundary, failed-rejudge recovery, and status contract are corrected.

internal/guidance/workflow/state.go
```diff
@@ -468,6 +556,11 @@ func (s *WorkflowState) RejectWithDecision(feedback string, decision DecisionMet
 	state.Status = StepStatusBlocked
 	state.Feedback = feedback
 	state.RemediationPhase = ""
+	if decision.Actor != "agent" && state.Judge != nil && state.Judge.Status != JudgeCanceled {
+		// Replace any prior terminal judge outcome so a later judge command
+		// cannot reopen a checkpoint after an operator has taken ownership.
+		state.Judge = nil
```
obey-agent, line 594: ⚠️ [P1] Persist the judge cleanup through the event log. In the event-backed path this assignment only mutates the in-memory state; Navigator.RejectWithDecision emits wf_step_block for a later operator rejection, and the materializer leaves the previous terminal JudgeRejected record intact because no judge-clear event is emitted. After a judge rejection followed by fest workflow reject, the current process shows no Judge but a new fest workflow status/watch reload reports judge_status=rejected and stale judge metadata. Emit an explicit clear/cancel event or materialize non-agent blocks accordingly, and add a cross-process round-trip test.

internal/commands/workflow/approval_guidance.go
```diff
@@ -33,8 +33,8 @@ func approvalRecoveryLinesFor(ctx context.Context, nav *wf.Navigator, step wf.Wo
 		}
 	}
 	return []string{
-		"Re-submit to the judge: " + ui.Accent("fest workflow approve --auto"),
 		"Operator override:       " + ui.Accent("fest workflow approve") + " (interactive)",
+		"Re-run approval judge:   " + ui.Accent("fest workflow judge"),
```
obey-agent, line 46: ⚠️ [P2] Only advertise the judge retry route for judge-owned blocks. This helper is also called immediately after fest workflow reject and by workflow show/advance, but it chooses the judge line solely from configuration. With a configured judge, an ordinary operator rejection therefore tells the user to run fest workflow judge even though IsJudgeRejection correctly refuses to reopen that state. Pass the current StepState into this guidance or gate the retry line on wf.IsJudgeRejection, and add coverage for configured-judge operator rejection guidance.

obey-agent (commented):
Verdict: Request Changes

Overview:
The follow-up revision correctly addresses the earlier review findings: rejudge preflight is now fail-safe, legacy feedback parsing is escape-aware, operator ownership is checked, and the v1 status fields are retained with concise judge detail. The new show/watch feedback opt-in and waiting-judge rendering are also cleanly integrated. One durable event-replay issue and one misleading recovery hint remain before production approval.

Key findings:
- P1: RejectWithDecision clears a prior terminal JudgeRejected record only in memory. The event-backed reload retains stale judge_status and metadata after a later operator rejection, so watch/status can disagree across process boundaries.
- P2: Configured-judge recovery guidance always advertises fest workflow judge, including ordinary operator rejections where that command is intentionally refused.

What is done well:
- Rejudge preflight now runs before mutating blocked state and has regression coverage.
- Legacy audit parsing handles reason= inside free-form reasons.
- JSON v1 compatibility fields and concise feedback are both preserved.
- Show/watch option plumbing and waiting-judge rendering are covered by focused tests.

Validation:
- go test ./... -count=1
- go vet ./...
- just docs
- just build quick
- Async judge integration suite with Docker, -count=1

Staff standard:
The revision is substantially stronger and close to ready, but I would not approve until the event-backed state is consistent after operator takeover and the recovery guidance reflects ownership.

internal/guidance/workflow/navigator.go
```diff
@@ -461,6 +487,14 @@ func runningJudgeRunID(state *StepState) string {
 	return state.Judge.RunID
 }
 
+func terminalJudgeRunID(state *StepState, decision DecisionMetadata) string {
```
obey-agent: ⚠️ [P2] Emit cleanup for legacy terminal judges with an empty RunID. terminalJudgeRunID returns the empty string for an older JudgeRejected/JudgeFailed record, and the callers therefore skip EmitJudgeClearedEvents entirely. The in-memory state is cleared, but event replay reconstructs the old terminal Judge from the empty-ID wf_judge_returned event and leaves stale judge metadata after the operator decision. The materializer explicitly supports empty RunIDs, so return a separate found/should-clear signal (and emit the empty-ID clear event) and add a legacy empty-RunID round-trip test.

obey-agent (commented):
Verdict: Request Changes

Overview:
This revision resolves the two prior blocking findings: operator judge cleanup is now durable across event reloads, and recovery guidance correctly distinguishes operator-owned rejections. The lifecycle hardening, show/watch integration, compatibility fields, and regression coverage are strong, and all reviewed gates pass. One legacy empty-RunID replay edge remains before approval.

Key findings:
- P2: terminalJudgeRunID returns an empty string for legacy terminal judge records without a RunID, causing callers to skip the new judge-cleared event; event replay then restores stale judge metadata after operator takeover.

What is done well:
- Durable judge cleanup uses explicit events and protects against late verdict replay.
- Rejudge preflight remains fail-closed and preserves blocked state on failure.
- Operator-owned recovery guidance no longer advertises the judge command.
- Tests now cover stale leases, event reload cleanup, feedback parsing, status compatibility, and show/watch rendering.

Validation:
- go test ./... -count=1
- go vet ./...
- just docs
- just build quick
- Fresh async judge Docker integration suite with -count=1

Staff standard:
I would be proud of the implementation direction and consider it close to production-ready, but I would fix the legacy empty-RunID replay case before approving.

obey-agent (commented):
Verdict: Approve\n\nOverview: Re-reviewed the latest head after the empty-RunID cleanup update. The prior stale-judge replay and misleading approval guidance issues are resolved, and operator decisions now durably clear both current and legacy terminal judge state across event reloads.\n\nKey Findings: No blocking or non-blocking findings.\n\nWhat is done well:\n- Judge cleanup is event-backed and guarded against late non-matching run IDs.\n- Legacy empty-RunID terminal records now emit cleanup events instead of remaining stale after reload.\n- Recovery guidance distinguishes ordinary operator blocks from judge rejections.\n- Regression coverage exercises event materialization, reload behavior, and async judge flows.\n\nValidation:\n- go test ./... -count=1\n- go vet ./...\n- just docs\n- just build quick\n- Fresh Docker-backed async judge integration suite: all TestAsyncJudge_* tests passed\n\nStaff standard: Yes — ready to merge.

PR #274, opened 2026-07-15
Resolve both dungeon/ and .dungeon/ spellings for the festivals dungeon

lancekrogers:
## What

Every fest code path that resolves the festivals dungeon (festivals/dungeon/ with its completed/archived/someday subtrees) accepts both the visible `dungeon/` spelling and the hidden `.dungeon/` spelling. Resolution rule: whichever spelling exists on disk is used; if neither exists yet, the visible spelling is created by default (unchanged behavior for campaigns that already exist today).

A campaign with both spellings present at once is a broken migration state, not a supported layout. Operations whose job is to actually read or write the dungeon now refuse with a clear error pointing at `camp dungeon migrate` instead of silently preferring one spelling and hiding whatever is filed under the other.

This implements the fest half of intent `hidden-dungeon-by-default-and-20260715-002142` ("Hidden .dungeon by default and rename intents to ideas"), amended by that intent's "Revised model: migrate, do not coexist" section: fest resolves both spellings and errors loudly on conflict; the `dungeon_hidden` setting, `camp init` scaffolding of the hidden directory, and the `camp dungeon migrate` command that consolidates a drifted campaign are camp's half, tracked separately.

## Why

New campaigns will scaffold `.dungeon/` (hidden) instead of `dungeon/` to reduce onboarding friction. Existing campaigns keep working with zero migration. Every fest command that reads or writes the dungeon has to understand both spellings for that to be true.

This PR originally shipped a prefer-visible-and-warn rule for the case where both spellings exist. Verifying that behavior surfaced the problem directly: when both spellings exist, whichever one is preferred wins, and anything already filed under the other spelling silently drops out of `fest list dungeon` (and out of chain discovery, festival lookup by name, and every other dungeon-touching read). A one-time stderr warning was the only signal that data had gone missing from a listing. Coexistence also does not stay rare: a rule that lets new dungeon directories in an existing campaign follow a `dungeon_hidden` setting guarantees a legacy campaign accretes `.dungeon/` directories one at a time alongside its `dungeon/` directories, so the mixed state was the eventual destination, not an edge case.

The fix is to treat two spellings as an error that names the problem and points at the fix, not a state to silently tolerate forever. Migration is cheap here because campaigns are filesystem plus git with no database, so a `git mv` sweep (camp's job) resolves it in one commit.

## How it was implemented

The resolver sweep from the original PR is unchanged. A single resolution point still lives in `internal/workspace/dungeon.go`:

- `ResolveDungeonDir(festivalsRoot)`, `JoinDungeon(festivalsRoot, elem...)`, `JoinStatus(festivalsRoot, status)`, and `IsDungeonDirName(name)` keep their existing signatures and behavior, including the prefer-visible fallback and one-time stderr warning. These stay as a safety net for the many generic, multi-status resolution paths that touch the dungeon only incidentally while resolving something else (for example `fest next`'s best-effort chain-context hint, and navigation-link resolution used to find "the current festival" from cwd). Breaking those in a drifted campaign would fail work that has nothing to do with the dungeon.

New in this amendment: `workspace.CheckDungeonConflict(festivalsRoot) error` reports an error when both spellings exist. It is called explicitly at the entry points whose actual job is to read or write the dungeon, rather than threaded through every `ResolveDungeonDir`/`JoinDungeon`/`JoinStatus` call site (which would ripple across roughly twenty call sites for a condition that is almost always false):

- `fest promote` into a dungeon status and `fest status set` into a dungeon status (two separate code paths, `AtomicStatusChange` and `executeFestivalMove`)
- `fest list dungeon`, `fest list dungeon/<substatus>`, and `fest list --all` (shared board builders in `internal/commands/list/list_board.go`, so one-shot, `--watch`, and `--json` all get the same protection)
- `fest status list --status dungeon...`
- `fest chain complete`, `fest chain status`, and `fest chain list` (chain discovery that explicitly checks the dungeon for completed chains)
- `fest show <name>` and `fest status <path>` festival discovery: these search every status directory including the dungeon buckets, so a plain "not found" at the end of that search now surfaces the real problem when a dungeon conflict exists, instead of quietly reporting a false negative
- the deprecated `fest show dungeon` / `fest show completed` / `fest show all` aliases, so they do not regress relative to their `fest list` equivalents
- the two interactive TUI festival pickers (`internal/commands/tui/hierarchy.go` `HierarchySelector.ListFestivals` and `internal/commands/tui/selector.go` `FestivalSelector.loadFestivals`), both of which scan the dungeon buckets alongside active/planning and previously swallowed the read error with `continue`, the same pattern already fixed in the `fest list` board builders. `FestivalSelector`'s default config backs the interactive picker `fest status set` opens when run without an explicit target, so this closes a gap in one of the entry points already named above, not just an adjacent utility. Both checks are gated on the dungeon buckets actually being in scope, so an active-only selector is unaffected by an unrelated dungeon conflict elsewhere in the campaign. A one-time stderr warning is a weak signal here regardless: once a huh/bubbletea form has painted the alt-screen, an earlier stderr warning is not visible to the user, which is why these two get a hard, pre-render error instead.

The static contract-entries change (obey daemon watcher `FallbackPaths`) from the original PR is unaffected and stands as built.

## Known limitation / follow-up seam

`dungeon_hidden` is still not wired into fest's own creation paths; when fest itself must create a dungeon path and neither spelling exists yet, it defaults to creating the visible `dungeon/`. That is camp's setting to own, per the implementation plan.

`camp dungeon migrate` itself is not part of this PR. Fest's job is only to stop silently hiding data by naming the conflict; camp owns the actual consolidation.

## Verification

- `go build ./...`, `go vet ./...`, and `gofmt -l` on every file this PR touches are clean.
- `just test unit-raw`: the full unit test suite passes, including new coverage in `internal/workspace/dungeon_test.go` (`TestCheckDungeonConflict`, table-driven: both exist errors and names both spellings plus the migrate hint; either alone or neither present returns nil).
- `just build test-binary` then `go test -tags=integration -run TestDungeonDualResolution -v ./tests/integration/...`: all subtests pass, including three that replace the old prefer-visible-and-warn assertions:
  - `BothExist_PromoteErrors`: `fest promote --dungeon completed` refuses when both spellings exist, names both spellings and the migrate hint in the error, and leaves the festival untouched in `planning/`
  - `BothExist_StatusSetErrors`: `fest status set archived` (the separate `executeFestivalMove` code path) refuses the same way
  - `BothExist_ListDungeonErrors`: `fest list dungeon` and `fest list dungeon/completed` both refuse with the migrate hint, while `fest list active` keeps working in the same drifted campaign
- The hidden-only and legacy-visible-only fixture subtests from the original PR are unchanged and still pass, confirming the non-conflict resolution behavior was not affected.
- `internal/commands/tui/dungeon_conflict_test.go` (new, white-box, no pty needed since neither function under test renders a form): `TestHierarchySelector_ListFestivals_DungeonConflict` and `TestFestivalSelector_loadFestivals_DungeonConflict`, both table-driven, covering both-exist-errors, either spelling alone, neither present, and an active-only filter being unaffected by an unrelated conflict.
- `just test integration-raw` (the full containerized integration suite, both the main package and the lifecycle sub-package) passes.
- `just lint all` reports one pre-existing `staticcheck` finding in `internal/campledger/emit.go` (an empty branch, SA9003) that this PR does not touch; confirmed unchanged from before this branch diverged, so it is not a regression introduced here.


internal/workspace/dungeon.go
```diff
@@ -0,0 +1,96 @@
...
+// resolves the dungeon many times (e.g. scanning several status
+// directories) only prints the warning once.
+var dungeonConflictWarned sync.Map
+
+// ResolveDungeonDir returns the on-disk directory name to use for the
+// dungeon under festivalsRoot: whichever of dungeon/ or .dungeon/ actually
+// exists. If both exist, the visible spelling wins and a one-time warning is
+// printed to stderr. If neither exists yet, it defaults to the visible
+// spelling so callers creating a dungeon path get the current behavior.
+func ResolveDungeonDir(festivalsRoot string) string {
```
obey-agent, line 47: **Single resolution choke point is the right design.** Prefer visible when both exist (with process-lifetime deduped warning), hidden when only that exists, default visible when neither — matches tribal-knowledge contract. Callers that previously hard-coded `filepath.Join(..., "dungeon", ...)` now go through JoinDungeon/JoinStatus, which is how dual spelling actually stays correct as new commands are added.

internal/workflow/transition.go
```diff
@@ -50,7 +52,10 @@ func findItemStatus(workflowRoot, item string) (string, error) {
 	}
 
 	for _, entry := range entries {
-		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+		if !entry.IsDir() {
```
obey-agent, line 55: **findItemStatus no longer skips `.dungeon`.** Correct: previously `strings.HasPrefix(name, ".")` treated the hidden dungeon as a non-status directory and would miss items under it. The carve-out via `IsDungeonDirName` is required for dungeon_hidden campaigns.

tests/integration/dungeon_dual_resolution_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// TestDungeonDualResolution exercises fest's status/list/promote-to-dungeon
+// flows against campaigns scaffolded with the visible dungeon/ spelling, the
+// hidden .dungeon/ spelling, and both at once, verifying every flow resolves
+// to the correct on-disk directory per docs/tribal-knowledge dungeon
+// resolution rules: whichever exists; visible wins (with a warning) if both
+// exist; visible is created when neither exists yet.
+func TestDungeonDualResolution(t *testing.T) {
```
obey-agent, line 24: **Integration matrix is complete enough to ship:** hidden promote+list, status set archived, legacy visible untouched, both-exist prefers visible + warns. That is exactly the contract operators care about.

obey-agent (commented):
## Verdict: Approve

## Overview
Centralizes festivals dungeon path resolution so both `dungeon/` and `.dungeon/` spellings work everywhere status/list/promote/repair/watch touch disk. Contract entries gain FallbackPaths for the daemon; unit + containerized integration cover prefer-visible, hidden-only, and both-exist.

## Key Findings
- No blocking issues. Resolution API (`ResolveDungeonDir`, `JoinDungeon`, `JoinStatus`, `IsDungeonDirName`) is the right abstraction.
- Conflict warning is process-deduped via `sync.Map` — good for multi-resolve commands.
- Minor non-blocker: each `JoinDungeon`/`JoinStatus` still stats the filesystem; fine at CLI scale, not a hot path concern.

## What's Done Well
- Mechanical replacement of hardcoded joins across chain/nav/status/repair/watch
- `findItemStatus` and list-root filtering fixed for hidden dungeon
- Integration tests exercise promote, status set, list, and both-spellings conflict

## Staff Standard
Yes. This is how dual-spelling support should have been done — one package, not scattered string compares.

internal/workspace/dungeon.go
```diff
@@ -0,0 +1,124 @@
...
+	hidden := isExistingDir(filepath.Join(festivalsRoot, HiddenDungeonDir))
+
+	switch {
+	case visible && hidden:
+		warnDungeonConflict(festivalsRoot)
+		return DungeonDir
+	case hidden:
+		return HiddenDungeonDir
+	default:
+		return DungeonDir
```
obey-agent, line 58: **fest can read both spellings but can only ever create the visible one**

This `default` is the read fallback, and as a read fallback it is fine. The problem is that `JoinDungeon` is also every **creation** path in fest, so this line decides the spelling of dungeons fest brings into existence — and it always answers `dungeon`.

`HiddenDungeonDir` appears nowhere outside this file:

```
$ git grep -rn "HiddenDungeonDir" -- '*.go' | grep -v _test.go
internal/workspace/dungeon.go:19,21,37,44,55,98,105,118
```

camp#437 hit this exact problem and split it in two: `spelling.Resolve` for reads (defaults visible) and `spelling.NameForNew` for creates (follows the campaign's established spelling). Its doc is explicit that the pair is load-bearing: *"Defaulting Resolve to Visible when neither exists is correct only because creates go through NameForNew."* fest#274 has the `Resolve` half and no `NameForNew` half.

Concretely, on a campaign camp scaffolded with `dungeon_hidden` (so root is `.dungeon`, as are `workflow/*` and `.campaign/intents`):

- camp's init reconciler covers `absDir`, `workflow/reviews`, `workflow/design`, `workflow/explore`, and `.campaign/intents`. `festivals/` is **not** in that list — camp shells out to `fest init` for it and never revisits the spelling.
- Whenever `festivals`' dungeon first comes into existence — `fest init` copying its template tree, or lazily via `chain/complete.go:101`'s `os.MkdirAll`, `status/date_directory.go`, or `repair_execute.go` — it resolves through here, neither spelling exists, and fest creates **`festivals/dungeon`**, visible.

So a brand-new `dungeon_hidden` campaign ends up hidden everywhere except `festivals/`, which is the directory users look at most. That is the one the feature exists to stop people asking about.

It is also silent. `camp status`'s notice (`notice.DungeonLegacy`) resolves **only the root dungeon** and returns nil when it is hidden, so the campaign never gets told to run `camp dungeon migrate`, and the mixed layout is permanent unless someone notices by eye.

fest needs the `NameForNew` equivalent: when creating a dungeon, follow the campaign's established spelling rather than defaulting to visible. The campaign root is above `festivalsRoot`, so this needs a spelling input (resolve the campaign's root dungeon, or take the spelling from config) rather than deciding from `festivalsRoot` alone.

internal/commands/chain/complete.go
```diff
@@ -96,7 +97,7 @@ func runComplete(ctx context.Context, chainID string, force bool, notes string)
 		return err
 	}
 
-	destDir := filepath.Join(root, "dungeon", "completed", "chains")
+	destDir := workspace.JoinDungeon(root, "completed", "chains")
```
obey-agent, line 106: **Chain completion is named in the guard's contract but does not call it**

`CheckDungeonConflict`'s doc says exactly where it must be called:

> Call this at the entry of operations whose actual job is to read or write the dungeon (list dungeon, promote/complete into a dungeon status, **chain completion**, festival discovery that falls through to "not found").

`runComplete` is chain completion, and it has no `CheckDungeonConflict` call. It resolves the dungeon here, `MkdirAll`s into it, writes the chain, and then **removes the original** (line 117). On a campaign holding both spellings that runs to completion silently against whichever spelling `ResolveDungeonDir` happens to prefer, which is the case the guard exists to refuse.

`internal/commands/system/repair*.go` and `internal/commands/list/list_watch.go` are also unguarded. Repair is the one I would look at hardest: `repair_analyze.go:53` scans `JoinDungeon(festivalsRoot)` and classifies what it finds into orphans and unknowns, and `repair_execute.go` moves things based on that classification — all while the other spelling's contents are invisible to the scan.

The deeper issue is the shape rather than these specific misses. An opt-in guard that every dungeon-touching call site must remember to invoke will keep drifting: `b694bc6` ("close two more silent-hide sites in the tui picker package") is that drift already happening once inside this PR. camp#437 avoided it by making the choke point itself refuse — `spelling.Resolve` returns `ConflictError`, so no caller can forget.

I understand why fest split it: generic multi-status helpers touch the dungeon incidentally and should not hard-fail, and `promote.go:501` building `searchDirs` across `id.StatusDirectories` is a fair example of the exemption. But the exemption is the minority of call sites, and the enumeration is inverted — it would be more robust for resolution to fail closed and for the incidental helpers to opt *out* explicitly, than for every read/write site to remember to opt in.

At minimum for this PR: add the guard to `runComplete`, `system/repair*`, and `list_watch`. If you want to keep the opt-in shape, the exemption list belongs in the doc as the enumerated set, so the next command author knows which side they are on.

obey-agent (commented):
## Staff Review (re-review at b694bc6) — Request Changes

### Why this supersedes the earlier Approve

The previous review approved `fd93899d`. Two commits landed after it, and one of them reverses that review's premise: `c976b3f` makes both spellings an **error**, where the approved design preferred visible and warned. `b694bc6` then closed "two more silent-hide sites." Both changes are right, and they align fest with the direction camp#437 took independently. But they change the contract enough that the Approve no longer describes this PR, so I re-reviewed at `b694bc6`.

### Blocking

1. **fest can read both spellings but can only ever create the visible one** (`internal/workspace/dungeon.go:46`). `ResolveDungeonDir` defaults to visible when neither exists, and `JoinDungeon` is also every creation path in fest — `HiddenDungeonDir` is referenced nowhere outside `dungeon.go`. camp#437 split exactly this into `Resolve` (read) and `NameForNew` (create) and documents that the split is what makes the visible default safe. fest has the read half only.

   The consequence is not theoretical. camp's init reconciler covers the root, `workflow/*`, and `.campaign/intents` — **not `festivals/`**, which camp delegates to `fest init`. So on a `dungeon_hidden` campaign, whenever the festivals dungeon first gets created, fest creates `festivals/dungeon` **visible** while the rest of the campaign is `.dungeon`. And it is silent: `notice.DungeonLegacy` only resolves the **root** dungeon, so a campaign whose root is already hidden never gets told to run `camp dungeon migrate`. A brand-new hidden campaign quietly keeps a visible dungeon in the directory users spend the most time in. Inline.

2. **The conflict guard is opt-in, and chain completion — which its own doc names — does not call it** (`internal/commands/chain/complete.go:100`). `runComplete` resolves the dungeon, `MkdirAll`s into it, writes, and deletes the original, with no `CheckDungeonConflict`. `system/repair*` and `list/list_watch.go` are unguarded too; repair is the worrying one, since it classifies orphans/unknowns from a scan that cannot see the other spelling and then moves files based on that classification. `b694bc6` is itself evidence that a remember-to-call guard drifts. Inline.

### Not blocking, but the branch needs updating

`just lint all` is **red** at this head:

```
internal/campledger/emit.go:128:2: SA9003: empty branch (staticcheck)
```

This is **not** this PR's code. The branch is 7 commits behind main, and main already removed that block in `4c016a7` ("fix: harden fest dependency versions"). Merging main in clears it. Flagging only so the red gate is not mistaken for a defect here, and because the branch should be updated before merge regardless.

### Verified at `b694bc6`
- `just test unit` — **pass** (2562/2562)
- `just test integration` — **pass** (68/68, containerized)
- `just lint vet` — **pass**
- `just lint all` — **fail**, stale-branch drift only (above)

(Note for anyone reproducing: the integration harness bind-mounts `bin/linux/fest`, so it only runs from a worktree under `$HOME` — colima does not mount `/private/tmp`.)

### What's done well
- Centralizing on `ResolveDungeonDir`/`JoinDungeon`/`JoinStatus` is the right call, and the mechanical replacement of hardcoded `filepath.Join(..., "dungeon", ...)` across chain/nav/status/repair/watch is what makes dual spelling hold as new commands land.
- `IsDungeonDirName` correctly carves the hidden dungeon out of the `strings.HasPrefix(name, ".")` skip in `findItemStatus` — that would have silently skipped every item under `.dungeon`.
- `CheckDungeonConflict`'s error carries a `WithHint` pointing at `camp dungeon migrate`, so the failure tells the user the fix. The wording lines up with camp's `ConflictError` rather than inventing a second vocabulary.
- The conflict warning being deduped per festivals root via `sync.Map` is the right detail for commands that resolve many times in one run.
- Deciding to error rather than silently prefer one spelling was the correct reversal, and it landed here independently of camp reaching the same conclusion.

### Staff Standard
The resolution layer is well built and the direction change is right. What I cannot sign off on is shipping the half that reads both spellings without the half that creates the correct one — paired with camp#437 that leaves the flagship `festivals/` dungeon visible on exactly the campaigns that asked for it hidden, with no notice to tell anyone. That is the feature quietly not doing its job, which is worse than not shipping it yet.

Also note the ordering constraint I raised on camp#437: `camp dungeon migrate` renames `festivals/dungeon`, and no released fest understands `.dungeon` until this lands **and ships**. These two need to go out together.

**Verdict: Request Changes** (posted as a comment per campaign recipe).

obey-agent:
**Re-review findings addressed (at `2dac394`)**

Both blocking findings from the `b694bc6` re-review are fixed, plus the stale-branch lint.

**1. fest could only ever create the visible dungeon** (`internal/workspace/dungeon.go`)
`ResolveDungeonDir`'s neither-exists fallback now follows the campaign's established spelling — inferred from the campaign root (`filepath.Dir(festivalsRoot)`) — instead of always returning visible. Deciding this at the shared choke point fixes every `JoinDungeon`/`JoinStatus` creation path (chain complete, date-directory, status moves, repair) at once. I deliberately chose this over a parallel `NameForNew`/`JoinDungeonForNew` variant: a create-only helper that every write site must remember to call is the same opt-in fragility finding 2 flags. A read against a not-yet-created dungeon returns a non-existent path under either spelling, so following the campaign spelling is harmless for reads and correct for creates. `NormalizeNewDungeonSpelling` handles the one creation path that bypasses the choke point — `fest init` copying the template tree (always visible) — by renaming to `.dungeon/` on a hidden campaign.

**2. Opt-in guard skipped by key call sites**
`CheckDungeonConflict` is now called by `runComplete` (chain completion), `runRepair` (system repair — analyze + execute), and `runListWatch`. A both-spellings campaign refuses with the `camp dungeon migrate` hint instead of reorganizing or listing against a single-spelling view.

**3. Stale-branch lint** — merged `main` in; `internal/campledger/emit.go` SA9003 is gone (removed in main's `4c016a7`).

**Gates (local, at `2dac394`):** `just test unit` — 2564 pass. `just test integration` — 68/68 (containerized), including a new `HiddenCampaignRoot_CreatesHiddenDungeon` case (hidden campaign creates `.dungeon` on first promote, no stray visible dungeon). `just lint all` — 0 issues. New unit coverage for the campaign-spelling fallback and `NormalizeNewDungeonSpelling`.

Ordering constraint from the camp#437 thread still holds: `camp dungeon migrate` renames `festivals/dungeon`, and no released fest understands `.dungeon` until this ships — these two need to go out together.

PR #278, opened 2026-07-17
fix(workflow): send deliverable evidence to the approval judge

lancekrogers:
## Problem

The approval judge rules on evidence it never receives. `judgeApproval` built a `fest.approval.judge/v1` request carrying only `document` (`WORKFLOW.md` / `GATES.md`), which for a `PRESENT` step is just the step definition. The actual deliverables live in `output_specs/`, so the judge (`ob judge`) read a template:

- rejected real work as *"no evidence, only the workflow definition"* even though the files existed, or
- approved on claims it read from the agent-authored document and could not verify.

`fest` already resolves and validates those deliverable paths in the `--auto` readiness gate (`checkApprovalReadiness`), then threw them away: the request struct had no evidence field.

## Fix

- Add an additive `evidence []string` (phase-relative paths) to `approvalJudgeRequest`. Populated in `judgeApproval` from `resolveExistingEvidencePaths`, which reuses the same `DefaultEvidencePaths` + `inspectApprovalEvidence` the readiness gate uses, so the judge receives exactly the existing, non-empty files readiness confirmed.
- The field is additive to the versioned protocol: an older judge ignores it and sees only `document`, exactly as before. The paired judge-side change is `Obedience-Corp/obey#182`.

## Secondary: readiness hint named the path

The `presentation artifact missing or empty` and `no non-empty evidence files` hints never stated the path they required; it was discoverable only by running `strings` on the binary (`output_specs/PRESENTATION.md`, phase-relative). Both hints now name the expected path(s).

## Tests

- `resolveExistingEvidencePaths` returns present deliverables, skips missing/errored, is nil when a step has no conventional evidence, and honors explicit `**Evidence:**`.
- End-to-end: `judgeApproval` over a real navigator + temp phase captures the request the judge receives and asserts `evidence` carries the two existing `output_specs/` files.
- Full `go test ./...` green; `golangci-lint` clean on the touched package.

## Notes

Scope was deliberately the evidence-bundle fix, not giving the judge live read-tools. That larger direction (a phase-scoped tool source + judge session plumbing in `ob`) is tracked as a follow-up; this closes the live gate failure with a minimal, backward-compatible protocol addition.

### DIFF
--- a/internal/commands/workflow/approval_hardening_test.go
+++ b/internal/commands/workflow/approval_hardening_test.go
@@ -2,7 +2,9 @@ package workflow
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview
Minimal, correct protocol fix: the readiness gate already knew which deliverables existed, and the judge request was throwing them away. Reusing the same path resolution for `evidence` is the right integration move — no parallel implementation.

## Key Findings
- No blocking issues.
- Additive `evidence []string` on `fest.approval.judge/v1` is backward compatible with older judges.
- `resolveExistingEvidencePaths` drops missing/empty/errored paths rather than sending ghosts the judge cannot read.
- Readiness hints now name the expected phase-relative paths — small but high-value operator fix.
- End-to-end test captures the stdin request the judge actually receives.

## What's Done Well
- Reuses `DefaultEvidencePaths` + `inspectApprovalEvidence` instead of inventing a second resolver.
- Explicitly scopes out live read-tools for the judge as a follow-up — right call for this PR.
- Table-style unit tests for present/missing/no-conventional/explicit evidence.

## Staff Standard
Yes. This is how you fix a live gate failure without overbuilding. Pair-merge with obey#182.

PR #284, opened 2026-07-19
chore: standardize VHS evidence recipes

lancekrogers:
## Summary

- wire Fest VHS bundle to the campaign shared optimizer, metadata collector, privacy scanner, and fail-closed validator
- replace the direct termcast/public-capable path with the secret-Gist helper
- keep raw evidence local and make campaign-root discovery explicit

## Verification

- just list for the VHS module
- VHS doctor passed
- dry-run bundle recipe parsed successfully
- shared validator self-test passed from CF0004

Based on current remote main at d753f782.

### DIFF
--- a/.justfiles/vhs.just
+++ b/.justfiles/vhs.just
@@ -4,6 +4,8 @@
 # Build and validate its evidence with: just vhs bundle <evidence-dir> <binary> <tape> <fixture-id> <states...>
 # Create a secret-Gist handoff with: just vhs gist out/demo.gif
 
+set positional-arguments
+
 [private]
 default:
     @just --list --justfile {{source_file()}}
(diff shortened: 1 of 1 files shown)


obey-agent (commented):
Verdict: Approve

Overview: The VHS recipe now composes the shared optimizer, ffprobe metadata collector, privacy scanner, fail-closed evidence validator, and binary-safe secret-Gist helper. The latest revision also preserves quoted handoff arguments and supports the documented VHS_PRIVATE_GIST_ROOT override when camp root discovery is unavailable.

Key Findings: No unresolved correctness findings. The canonical consumer tapes still need their adoption-phase fixture paths sanitized before real recordings can pass the new privacy gate.

What is done well: Raw artifacts remain local; publication is secret-only; validation is repeated after the manifest-inclusive privacy scan; the project changes are small and based on current remote main.

Staff Standard: Yes, I would be proud to submit this recipe integration as my own work, with the pending CI check and downstream tape sanitization treated as merge gates.

PR #286, opened 2026-07-20
feat(workflow): surface judge followups so blocked agents can fix and continue

lancekrogers:
## Problem

When an approval judge auto-rejects a blocking checkpoint, a driving agent was handed a dead-end: the one-line reason plus two human-oriented exits (`fest workflow approve` interactive override, or a blind `fest workflow judge` re-run). The judge's `followups`, the itemized fixes the judge template is explicitly told to produce on rejection, were parsed into `approvalJudgeResponse.Followups` and then **dropped at every layer**: never printed, never recorded on the step, never in the `wf_step_block` / `wf_judge_returned` events, never rendered by `fest workflow status` or `fest next`.

So an autonomous agent got the "why" but never the "what to fix", and stopped waiting for a human. This is why agents kept getting blocked at the judge.

Blocking is the desired behavior. The fix is not to weaken it; it is to make a legitimate block **actionable and loopable** so the agent can pick up the rejection, fix the specific issues the judge found, and re-submit.

## Fix

**Followups become first-class** through the whole path:
- `DecisionMetadata`, `StepState`, and the `WorkflowEvent` -> `ProgressEvent` JSONL round-trip carry `followups`, so they persist in the append-only log and survive replay.
- Printed on auto-reject; rendered by `fest workflow status` and in the `fest next` blocked-gate view, next to the existing "revise, then re-run `fest workflow judge`" loop instruction.
- Included (bounded + sanitized) in the obey session-continuation message, so a resumed obey-managed agent gets the same fix list.
- Cleared on any transition out of blocked (advance, skip, recheck, judge re-run, reset).

The fix-and-continue loop itself already existed (`fest next` says revise + re-run judge; `fest workflow judge` reopens and re-evaluates). The only missing piece was handing the agent the concrete fix list. Blocking is unchanged: substandard work still stops.

## Proven end to end

Driven against a real sandbox festival with a fake judge (not mocked). `fest next` at a blocked phase gate now shows:

```
This step is BLOCKED: The PHASE_GOAL_VERIFICATION.md asserts the suite passes but
attaches no console output; claims about a merged PR are self-report, not proof.

Fixes the judge requires (address every item, then re-submit):
  - Attach the actual test-run console output showing pass/fail counts.
  - Include the ledger transition (409->410) as captured output, not an assertion.
  - Do not cite a merged PR as evidence the judge cannot see.

Address the feedback and try again.

When the revised evidence is ready, re-run the approval judge:
  fest workflow judge
```

The followups persist in the durable event log (`"followups": [...]` in the `wf_step_block` JSONL event), and the full loop closes: reject with fix list -> agent revises -> `fest workflow judge` approves -> gate advances to the next step.

## Tests / gate

- New unit coverage: state round-trip (`RejectWithDecision` stores + emits followups, cleared on reopen), event replay (`wf_step_block` -> `StepState.Followups`, cleared on `wf_judge_recheck`), continuation rendering (rejection with/without followups, bounding + sanitizing).
- `just lint all`: 0 issues. `go vet ./...`: clean. Full module `go test ./...`: 99 packages pass.

### DIFF
--- a/embedded/templates/agent/workflow/step.tmpl
+++ b/embedded/templates/agent/workflow/step.tmpl
@@ -56,7 +56,10 @@ EXAMPLE OUTPUT:
 Progress: Step {{.StepNumber}}/{{.TotalSteps}} | Phase: {{.PhaseName}} | Type: {{.PhaseType}}
 {{if eq .Status "blocked"}}
 This step is BLOCKED: {{.Feedback}}
-
+{{if .Followups}}
+Fixes the judge requires (address every item, then re-submit):
+{{range .Followups}}  - {{.}}
+{{end}}{{end}}
 Address the feedback and try again.
 {{if .JudgeRetryAvailable}}
 When the revised evidence is ready, re-run the approval judge:
--- a/internal/continuation/continuation.go
+++ b/internal/continuation/continuation.go
@@ -46,6 +46,9 @@ const (
 	// maxFeedbackLen bounds rendered feedback so a malformed or oversized
 	// judge reason cannot produce an unbounded agent message.
 	maxFeedbackLen = 500
+	// maxFollowups bounds the number of judge fix items rendered into a
+	// continuation message so a malformed verdict cannot make it unbounded.
+	maxFollowups = 12
 )
 
 // Verdict is the terminal judge outcome carried at the transport layer.
@@ -138,6 +141,9 @@ type JudgeResult struct {
 	StepName      string
 	Verdict       Verdict
 	Feedback      string
+	// Followups carries the judge's itemized fixes for a rejection so the
+	// resumed session receives the concrete revise list, not just the reason.
+	Followups []string
 }
 
 // DeliveryID derives the deterministic, retry-stable delivery id for a run.
@@ -165,8 +171,8 @@ func RenderMessage(r JudgeResult) string {
 	case VerdictApproved:
 		return fmt.Sprintf("Approval judge approved %s.\nContinue with: %s", loc, next)
 	case VerdictRejected:
-		return fmt.Sprintf("Approval judge rejected %s.\nFeedback: %s\nAddress the feedback, then run: %s",
-			loc, sanitizeFeedback(r.Feedback), next)
+		return fmt.Sprintf("Approval judge rejected %s.\nFeedback: %s%s\nAddress the feedback, then run: %s",
+			loc, sanitizeFeedback(r.Feedback), renderFollowups(r.Followups), next)
 	case VerdictFailed:
 		return fmt.Sprintf("Approval judge failed for %s.\nFeedback: %s\nInspect the judge configuration or evidence, then run: %s",
 			loc, sanitizeFeedback(r.Feedback), next)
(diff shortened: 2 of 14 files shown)


internal/continuation/continuation.go
```diff
@@ -197,6 +203,36 @@ func BuildNotification(r JudgeResult) SessionNotification {
...
+func renderFollowups(followups []string) string {
+	if len(followups) == 0 {
+		return ""
+	}
+	var b strings.Builder
+	b.WriteString("\nFixes required:")
+	n := 0
+	for _, f := range followups {
+		item := sanitizeFeedback(f)
+		if item == "(no detail provided)" {
```
obey-agent, line 220: Minor (maintainability, non-blocking): `renderFollowups` drops empty items by matching the exact string `"(no detail provided)"`, which is `sanitizeFeedback`'s empty sentinel (line 248). That's a fragile cross-function coupling — if that sentinel ever changes, this silently starts rendering `- (no detail provided)` bullets instead of skipping them. Cleaner and self-contained: drop blanks *before* sanitizing, the way the other two renderers do — `if strings.TrimSpace(f) == "" { continue }`. That removes the magic-string dependency and keeps empty-handling identical across surfaces.

embedded/templates/agent/workflow/step.tmpl
```diff
@@ -56,7 +56,10 @@ EXAMPLE OUTPUT:
 Progress: Step {{.StepNumber}}/{{.TotalSteps}} | Phase: {{.PhaseName}} | Type: {{.PhaseType}}
 {{if eq .Status "blocked"}}
 This step is BLOCKED: {{.Feedback}}
-
+{{if .Followups}}
+Fixes the judge requires (address every item, then re-submit):
+{{range .Followups}}  - {{.}}
```
obey-agent, line 61: Minor (consistency, non-blocking): this template ranges over `.Followups` raw, while `RenderWorkflowStepLine` (the `fest workflow status` path) and `renderFollowups` (the continuation path) both `TrimSpace` and skip empty items. The same `StepState.Followups` feeds `fest next` (this template) and `fest workflow status`, so a judge that emits a blank or a multi-line followup would render a stray `  - ` bullet or a line-broken list here but a clean list there — the two surfaces disagree for one step. Consider feeding all renderers from one shared trim+skip-empty helper so the fix list looks identical everywhere it appears.

obey-agent (commented):
**Verdict: Comment — merge-ready.**

Gates verified green on the head commit (`17efd8b`): `go build ./...` clean, `go vet ./...` clean, full `go test ./...` = 112 packages pass (incl. the new followup store/emit, JSONL replay, and bounds/sanitize tests).

**Overview:** A tight, correct fix for a real problem. The judge's `followups` — the itemized fixes it is explicitly told to produce on rejection — were parsed into `approvalJudgeResponse.Followups` and then dropped at every layer, so a blocked agent got the "why" but never the "what to fix" and stalled waiting for a human. This threads followups through the existing `DecisionMetadata` → `StepState` → `WorkflowEvent` → `ProgressEvent` JSONL round-trip and renders them in the block surfaces + the continuation message, without weakening the block.

**What's done well:**
- Extends existing structures rather than adding a parallel path — followups ride the same decision/event plumbing as feedback, so they persist and replay for free. Exactly the "improve existing, don't fork" standard.
- Clearing is comprehensive and consistent: every transition out of blocked (start / complete / skip / recheck / reset / reopen / clear-remediation) nils `Followups`, in both the in-memory methods and event materialization. No in-memory vs replay divergence — `StartCurrentStep` only touches `Pending` steps.
- Continuation transport is bounded (`maxFollowups=12`) and per-item sanitized — a hostile or broken judge can't produce an unbounded agent message.
- Tests are meaningful (store+emit, JSONL replay round-trip, clear-on-reopen/recheck, bounds/sanitize), not coverage padding. `omitempty` keeps it backward-compatible with existing events/state.

**Key findings (both minor, non-blocking — details inline):**
1. `renderFollowups` (`continuation.go:220`) drops empty items by matching `sanitizeFeedback`'s exact `"(no detail provided)"` sentinel — a fragile magic-string coupling. Trim-before-sanitize instead.
2. `step.tmpl:61` ranges over `.Followups` raw, while the status and continuation renderers trim + skip empties — so `fest next` and `fest workflow status` can disagree for the same step. One shared render helper would unify all three surfaces.

**Staff standard:** Yes — I'd be comfortable submitting this. It's a focused, well-tested extension of existing machinery that removes a genuine agent-blocking dead-end, with disciplined clearing and bounding. The only step from "good" to "exemplary" is collapsing the ~3 near-duplicate followup renderers into one shared trim/skip helper.

PR #288, opened 2026-07-20
fix: restore fgo completion colors

lancekrogers:
## Summary

- restore colors for explicit `fgo` zsh completion display cells when completion output is piped
- preserve the configured dark, light, or high-contrast theme for completion display cells without re-enabling colors in ordinary piped output
- add regression coverage for piped completion output
- add a VHS recording of the real `fgo <TAB>` menu

## Root cause

`fgo` requests `fest go completions --color`, but the completion command writes through a shell pipe. The theme resolver correctly selected a plain palette for normal non-interactive output, and the completion renderer then inherited that plain palette, removing the ANSI display colors.

The fix keeps ordinary piped output plain while resolving the configured theme explicitly for completion display cells.

## Validation

- `GOCACHE=/tmp/fest-status-list-colors-gocache just test unit` — 98 packages, 2,615 tests
- `GOCACHE=/tmp/fest-status-list-colors-gocache just lint vet`
- `GOCACHE=/tmp/fest-status-list-colors-gocache just build quick`
- verified a real piped `fest go completions --color` invocation emits ANSI status sequences
- recorded and visually inspected the real `fgo <TAB>` menu with VHS

The VHS source is committed at `docs/demos/fgo-completion-colors.tape`; the recorded GIF will be attached in the PR conversation.

### DIFF
--- a/docs/demos/fgo-completion-colors.tape
+++ b/docs/demos/fgo-completion-colors.tape
@@ -1,0 +1,37 @@
+# fgo zsh completion colors — explicit completion output through a pipe.
+#
+# Record with a disposable fixture containing the branch binary:
+# FEST_VHS_ROOT=/path/to/fest-status-colors-fixture vhs docs/demos/fgo-completion-colors.tape
+
+Output out/fgo-completion-colors.gif
+
+Set Shell "zsh"
+Set FontFamily "JetBrainsMono Nerd Font"
+Set FontSize 14
+Set Width 1040
+Set Height 700
+Set Padding 22
+Set Margin 0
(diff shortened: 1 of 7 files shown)


obey-agent:
VHS visual evidence for the completion fix:

`fgo` completion runs `fest go completions --color` through a pipe. The fix keeps ordinary piped output plain while restoring the configured theme colors for explicit zsh display cells.

fgo completion color recording

This is the real `fgo <TAB>` menu: active green, ready amber, planning blue, ritual purple, completed magenta, with archived and someday remaining distinct.

PR #291, opened 2026-07-22
feat(hooks): fest hooks system (FH0001)

lancekrogers:
## Summary

Implements the full fest hooks system from festival **FH0001** (`fest-hooks-system`): three declaration layers, step bindings, sequential runner with fail policies, human gates, evidence contract, audit events, and operator CLI surfaces.

- Add shared `hooks` config types and three-layer resolution (machine JSON, festivals YAML, festival YAML) with by-name whole-definition override and most-specific-wins switches
- Preserve legacy `hooks.approval_judge.command` as an implicit `approval_judge` definition (`timeout: 0`) with deprecation warning
- Parse step bindings from frontmatter and GATES.md `**Hooks:**` markers; skip+warn undeclared names
- Add sequential hook runner (fail open/closed, timeouts) and emit `wf_hook_run` festival-local audit events; route approval judge execution through the runner
- Implement non-bypassable human gates (`approval: human-required`): banner/parking, TTY-required approve/reject, refuse `--auto`/judge
- Document the read-by-approver evidence contract and add opt-in `evidence: embed` with a 256KB cap
- Ship `fest hooks list` (text/`--json`) and non-blocking validate warnings for shadow drift / legacy alias

## Test plan

- [x] `just build quick`
- [x] `just test unit` (all packages green)
- [x] `just lint all` (0 issues)
- [x] Smoke: `fest hooks list` / `fest hooks list --json` from linked project shows legacy `approval_judge` alias
- [ ] CI green on PR
- [x] Manual: human-required gate refuses `fest workflow approve --auto` and shows `HUMAN APPROVAL REQUIRED` on `fest next` (verified with the real binary in an isolated workspace, commit caace9e)

Festival completed and promoted:
`festivals/.dungeon/completed/2026-07-22/fest-hooks-system-FH0001`

## Review fixes (caace9e)

The staff review found the runner/audit/validation layers were built but not wired into production paths. caace9e addresses every finding:

- Step-bound pre/post hooks now execute on all four lifecycle verbs (task/sequence/phase complete, gate approve); fail-closed pre failures block the verb
- `wf_hook_run` audit events are emitted for every hook execution and skip, including the judge's own run
- Human-required steps record automation bindings as `skipped: human-gate`
- `fest validate` warns on undeclared bindings and renders a HOOKS section (hook warnings were previously never displayed)
- Evidence path-safety logic deduplicated (approval_readiness delegates to internal/hooks); festerrors replaces fmt.Errorf in new code
[...]

PR #293, opened 2026-07-23
docs: lifecycle hooks schema and operator guide

lancekrogers:
## Summary

Documents the fest lifecycle hooks system shipped in #291 so operators and agents can learn the schema without reading code.

- Add `docs/concepts/hooks.md` (layers, schema, bindings, fail policies, legacy alias, human gates, audit/validate)
- Document hooks in `docs/configuration.md` with YAML/JSON samples
- Add `fest understand hooks` (`embedded/docs/understand/hooks.txt` + command wiring)
- Update `fest understand workflow` approval section to preferred `definitions` form
- Link from `hook-evidence-contract.md` and refresh README bullet
- Regenerate CLI reference (`fest hooks`, `fest hooks list`, `fest understand hooks`)

## Test plan

- [x] `just build quick`
- [x] `just docs` regenerated `docs/cli-reference/fest_hooks*.md`
- [x] `./bin/fest understand hooks` prints the new topic
- [x] `go test ./internal/commands/understand/`

### DIFF
--- a/docs/concepts/hooks.md
+++ b/docs/concepts/hooks.md
@@ -135,14 +135,20 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 | `phase_complete` | phase completion |
 | `gate_approve` | GATES.md blocking approval checkpoint |
 
-Timings: **pre** (before the verb applies) and **post** (after success).
+Timings: **pre** (before the verb applies) and **post** (after the transition
+has already been applied).
 
 Orchestration:
 
 - Sequential, binding-list order (no parallel v1).
-- `fail: closed` (default): non-zero exit / timeout / spawn error short-circuits
-  remaining hooks and **blocks the verb**.
-- `fail: open`: failure is recorded; remaining hooks still run.
+- **`fail: closed` pre-hooks:** a non-zero exit / timeout / spawn error
+  short-circuits remaining pre-hooks and **blocks the verb** (completion or
+  approval is not applied).
+- **`fail: closed` post-hooks:** run **after** the verb is already applied.
+  Failure short-circuits remaining post-hooks and surfaces an audited warning;
+  it does **not** roll back task completion, sequence/phase completion, or gate
+  approval.
+- `fail: open`: failure is recorded; remaining hooks at that timing still run.
 - Undeclared bound names: **skip + warn** (not an error). Scaffolded templates
   may bind `approval_judge` while config is still empty.
 
--- a/docs/concepts/hooks.md
+++ b/docs/concepts/hooks.md
@@ -1,0 +1,232 @@
+# Lifecycle Hooks
+
+fest hooks are **named commands** bound to festival lifecycle events (task
+complete, sequence complete, phase complete, gate approve). They are the
+generalization of the older single `hooks.approval_judge.command` workspace
+setting.
+
+Related:
+
+- Evidence transport details: [hook-evidence-contract.md](./hook-evidence-contract.md)
+- Inspect resolution: `fest hooks list` / `fest hooks list --json`
+- Learn in-terminal: `fest understand hooks`
+
+## Two primitives
+
+| Primitive | What it is | Controlled by |
+| --- | --- | --- |
+| **Automation hooks** | Commands on lifecycle events (judge, linters, tests, notifications) | `hooks.*` config + step bindings |
+| **Human gates** | Person-required checkpoints | `approval: human-required` on the step |
+
+Human gates are **not** hooks. No `hooks.enabled` / level / per-hook switch can
+disable them.
+
+## Declaration layers
+
+Hook **definitions** live only at three layers. Each layer overrides the
+previous **by hook name** (whole definition replace; no field-level merge). A
+layer that declares nothing inherits everything above it.
+
+| Layer | Location | Format | Scope |
+| --- | --- | --- | --- |
+| 1. Machine | `~/.obey/fest/config.json` | JSON | every campaign on this machine |
+| 2. Festivals | `festivals/.festival/config.yaml` | YAML | every festival in the campaign |
+| 3. Festival | `fest.yaml` | YAML | this festival only |
+
+**Default at every layer is empty:** a fresh machine, campaign, or festival
+runs no hooks until you declare some.
+
+### Schema
+
+```yaml
+hooks:
+  enabled: true            # layer-wide switch; most specific layer wins
+  levels:                  # per-level switches; default all true
+    phase: true
+    sequence: true
+    task: true
+  definitions:
+    approval_judge:
+      command: ob judge    # required; cwd = festival root
+      fail: closed         # closed (default) | open
+      timeout: 120s        # default 120s for newly declared hooks; 0 = no deadline
+      evidence: paths      # paths (default) | embed
+      enabled: true        # per-hook switch
+```
+
+Field notes:
+
+- `command` is the only required definition field.
+- `definitions` is a map keyed by hook name (unit of layer override and binding).
+- `hooks.enabled` and `hooks.levels` resolve most-specific-wins across layers,
+  independently of `definitions`.
+- Commands are split on whitespace (same model as today's judge exec): no shell
+  operators or quoting in the command string.
+
+JSON machine-layer example:
+
+```json
+{
+  "hooks": {
+    "enabled": true,
+    "definitions": {
+      "lint": {
+        "command": "just lint",
+        "fail": "open",
+        "timeout": "60s"
+      }
+    }
+  }
+}
+```
+
+### Resolution (operator view)
+
+```
+effective definitions = merge least → most specific by name (replace whole def)
+enabled  = most_specific_set(hooks.enabled, default true)
+levels   = most_specific_set(hooks.levels.*, default true)
+runnable(name, level) =
+    enabled && levels[level] && definition.enabled
+```
+
+Inspect with:
+
+```bash
+fest hooks list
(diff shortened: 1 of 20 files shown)


embedded/docs/understand/workflow.txt
```diff
@@ -61,19 +61,30 @@ Common commands:
...
 
-Configuring `hooks.approval_judge.command` in `.festival/config.yaml` is the
-operator opt-in that delegates blocking checkpoints (WORKFLOW.md and GATES.md)
-away from human pings.
+Blocking checkpoints can be delegated to an automation hook named
+`approval_judge`. Preferred declaration (festivals layer):
 
-When that hook is set:
+  hooks:
+    definitions:
```
obey-agent, line 68: Blocking: the recommended schema here does not configure the current approval path. `approvalJudgeConfiguredFor` and `resolveApprovalJudgeCommandFor` read only `cfg.Hooks.ApprovalJudge.Command`, which is the legacy flat key; `definitions.approval_judge` is only consulted by generic hook resolution and evidence metadata. Following this documented migration makes `fest next`, `fest workflow judge`, and `fest workflow approve --auto` behave as if no judge is configured. Either wire auto-judge command discovery to the resolved `approval_judge` definition across the supported layers before documenting this as preferred, or retain the flat form as the required auto-judge configuration. Please add a regression test that configures only `definitions.approval_judge`.

docs/concepts/hooks.md
```diff
@@ -0,0 +1,232 @@
...
+| `phase_complete` | phase completion |
+| `gate_approve` | GATES.md blocking approval checkpoint |
+
+Timings: **pre** (before the verb applies) and **post** (after success).
+
+Orchestration:
+
+- Sequential, binding-list order (no parallel v1).
+- `fail: closed` (default): non-zero exit / timeout / spawn error short-circuits
+  remaining hooks and **blocks the verb**.
```
obey-agent: The fail-closed behavior needs to distinguish pre from post timing. `RunPost` executes after the verb is applied, and the task and gate callers explicitly warn while leaving the task completed or checkpoint approved; sequence and phase completion likewise do not roll back. As written, operators can incorrectly rely on a failed post-hook preventing completion. Please document that fail-closed pre-hooks block the verb, while fail-closed post-hooks short-circuit remaining post-hooks and surface an audited warning after the transition remains applied. The same overstatement appears in `embedded/docs/understand/hooks.txt`.

obey-agent (commented):
## Verdict: Request Changes

## Overview
This is a clear, well-structured operator guide and the new `fest understand hooks` surface is integrated cleanly with generated CLI documentation. The focused tests, documentation regeneration, and quick build all pass, but two documented production contracts do not match the shipped execution paths and would mislead operators.

## Key Findings
- The recommended `definitions.approval_judge` configuration is not read by auto-delegation; migrating from the flat key disables `fest next`, `fest workflow judge`, and `fest workflow approve --auto` judge discovery. See the inline comment on `embedded/docs/understand/workflow.txt`.
- Fail-closed post-hooks cannot block or roll back an already-applied lifecycle verb. They short-circuit remaining post-hooks and emit an audited warning while completion or approval remains applied. See the inline comment on `docs/concepts/hooks.md`.

## Done Well
- The layered schema, evidence contract, human-gate distinction, and audit guidance are presented accessibly.
- The embedded understand content, Cobra registration, and generated command reference stay synchronized; `just docs` produced a clean tree.
- Focused hook, workflow, task, progress, validator, and understand tests pass, as does `just build quick`.

## Staff Standard
Not yet. I would be proud to submit this after the auto-judge configuration contract and pre/post failure semantics are made accurate and regression-covered; those are core operator guarantees, not editorial details.

embedded/docs/understand/workflow.txt
obey-agent: Re-review: still blocking on the unchanged 26465cc head. Current `main` also retains flat-key-only command discovery in `approvalJudgeConfiguredFor` and `resolveApprovalJudgeCommandFor`, so the preferred `definitions.approval_judge` example still disables auto-judge behavior when followed.

docs/concepts/hooks.md
obey-agent: Re-review: still blocking on the unchanged 26465cc head. The documentation still says every fail-closed hook blocks the verb, while the implementation continues to apply the transition before post-hooks and only warns on a post failure.

obey-agent (commented):
## Verdict: Request Changes (unchanged)

## Overview
The PR head remains 26465cc, so the prior documentation review still applies without modification. Current `main` has not changed the underlying auto-judge discovery or post-hook behavior, and both published contracts remain inaccurate.

## Key Findings
- `definitions.approval_judge` is still documented as the preferred auto-judge form, but command discovery still reads only the legacy flat `hooks.approval_judge.command` field.
- Fail-closed post-hooks are still documented as blocking the lifecycle verb, although post-hooks run after the transition and failures only short-circuit and warn.
- Follow-up notes were added to both existing inline threads; both remain unresolved.

## Done Well
- The guide remains clear and comprehensive in its layering, evidence, human-gate, and audit explanations.
- The command wiring and generated documentation were previously verified clean.

## Staff Standard
Not yet. I would approve once the recommended auto-judge configuration and pre/post failure semantics match the behavior operators will actually receive.

embedded/docs/understand/workflow.txt
obey-agent: Original finding resolved on dccd30a: definitions-only approval_judge discovery is now wired through the three-layer resolver and has focused regression coverage. I am resolving this thread; a separate follow-up covers disabled-resolution semantics.

docs/concepts/hooks.md
obey-agent: Resolved on dccd30a. Both the concept guide and embedded understand topic now distinguish pre-hook blocking from post-hook warning behavior without implying rollback.

internal/commands/workflow/approve.go
```diff
@@ -315,21 +315,80 @@ func runApproveWithOptions(ctx context.Context, decision wf.DecisionMetadata, op
...
+	// Defensive fallback if alias expansion did not apply for any reason.
+	return strings.TrimSpace(wcfg.Hooks.ApprovalJudge.Command), nil
+}
+
+func approvalJudgeCommandFromEffective(eff *hooks.Effective) string {
+	if eff == nil {
+		return ""
+	}
+	h, ok := eff.Hooks[hooks.ApprovalJudgeName]
+	if !ok || !h.Enabled {
```
obey-agent, line 388: Blocking: this bypasses effective hook enablement. `h.Enabled` covers only the definition-level switch; the layer-wide `hooks.enabled` value lives in `eff.Enabled` and is enforced by `Effective.Runnable`. In addition, when a festival-level definition has `enabled: false`, the empty result here falls through to a machine/workspace-only resolve and resurrects the upper-layer command; the raw flat-key fallbacks have the same problem when an explicit disabled definition wins over the alias. A disabled judge can therefore still auto-run. Please treat a successful three-layer resolution as authoritative, gate the command with `eff.Runnable(approval_judge, gate)`, remove fallbacks that bypass resolved state, and apply the same rule in `Navigator.ApprovalJudgeCommand`. Add regressions for layer-wide and festival-level per-hook disables.

docs/concepts/hooks.md
```diff
@@ -0,0 +1,238 @@
...
+fest hooks list
+fest hooks list --json
+```
+
+Differing upper-layer definitions appear under `shadows` so a local override
+cannot silently hide an updated default without visibility.
+
+## Step bindings
+
+Definitions alone do nothing. Steps **bind** names to pre/post timing:
```
obey-agent, line 105: This is now internally inconsistent with the implemented approval-judge compatibility path: a configured `definitions.approval_judge` triggers `AutoDelegateBlockingCheckpoints` on eligible blocking checkpoints even without an explicit step binding. Please qualify this rule: ordinary lifecycle hooks require bindings, while `approval_judge` is the special operator opt-in that is implicitly applied to eligible approval checkpoints. Otherwise an operator can read this section and add redundant bindings or conclude that the later auto-delegation guidance is wrong.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Re-review of dccd30a confirms that both original findings are fixed: definitions-only judge discovery now works, and pre/post failure semantics are documented accurately. The focused checks pass, but the new resolver does not yet preserve the documented disable and override contract, and the operator guide contradicts the special implicit behavior of `approval_judge`.

## Key Findings
- `approvalJudgeCommandFromEffective` checks only per-definition enablement, ignoring layer-wide `hooks.enabled`; subsequent fallbacks can also resurrect an upper-layer judge after a more-specific disabled override. The same behavior is duplicated in navigator formatting. See the new inline comment on `internal/commands/workflow/approve.go`.
- The guide says definitions alone do nothing, but `approval_judge` now auto-delegates eligible checkpoints without an explicit step binding. Qualify the generic binding rule with this compatibility exception. See the new inline comment on `docs/concepts/hooks.md`.
- The two original review threads are resolved.

## Done Well
- The remediation correctly routes definitions-only configuration through the shared hook resolver and covers both execution and navigator presentation paths.
- Pre-hook versus post-hook failure behavior is now explicit and operationally accurate in both documentation surfaces.
- Focused workflow, guidance, hooks, and understand tests pass; vet, `just docs`, and `just build quick` are clean on the exact PR head.

## Staff Standard
Not yet. I would be proud to submit this once disabled and shadowed judge configurations remain authoritative across every discovery path and the guide states the approval-judge binding exception directly.

PR #294, opened 2026-07-23
fix(system): preserve user-owned .festival/config.yaml on system update

lancekrogers:
## Summary

Two integration tests were failing (`67/69`). Both trace to the two most recent commits, and one exposed a real, dangerous product bug.

### 1. `system update --force` deletes the operator's `.festival/config.yaml` (real bug)

The hooks system (#291) made `fest init` scaffold a **user-owned** `.festival/config.yaml`, written *after* checksums so it stays untracked. But `categorizeChanges` in `internal/commands/system/update.go` classified *any* workspace file absent from source as "orphaned → delete", excluding only `.state/`. So `config.yaml` was flagged for deletion.

This is not just a dry-run reporting quirk: with the fix reverted, `fest system update --force` **actually deletes** `.festival/config.yaml`, wiping operator hooks such as `hooks.approval_judge.command` — the very config #292 relies on for non-interactive judge safety.

**Fix:** exclude the user-owned `config.yaml` from orphan detection, alongside the existing `.state/` exclusion.

### 2. Approval-readiness containment test was stale

#292 TTY-gates `--judge-command`. `TestApprovalReadinessRejectsEvidenceSymlinkOutsidePhase` passed that flag in a non-interactive container, so it errored with *"--judge-command requires an interactive operator tty"* **before** reaching the symlink-containment check it exists to verify.

**Fix:** the test now configures the judge the operator-sanctioned way — via `.festival/config.yaml` (`hooks.approval_judge.command`) — and drops the flag. Same assertion (`escapes the phase directory`), valid non-interactive path.

## Changes

| File | Change |
|---|---|
| `internal/commands/system/update.go` | Exclude user-owned `config.yaml` from orphan deletion |
| `tests/integration/approval_readiness_test.go` | Exercise containment via `config.yaml` judge, not TTY-gated `--judge-command` |
| `tests/integration/system_update_test.go` | New `TestSystemUpdate_PreservesUserConfig` deterministic regression guard |

## Verification

- Full integration suite green locally (`go test -tags=integration ./tests/integration/...`, 179s, 0 failures).
- New `TestSystemUpdate_PreservesUserConfig` **verified to fail without the fix** (config.yaml deleted) and pass with it. It explicitly clears any stray `config.yaml` from the shared container's source dir so the guard is deterministic rather than dependent on test-order pollution.
- Host build + `go vet -tags=integration` clean.

> Note: repo CI is intentionally paused (cost); gates were run locally.

### DIFF
--- a/internal/commands/system/update.go
+++ b/internal/commands/system/update.go
@@ -349,18 +349,13 @@ func categorizeChanges(ctx context.Context, stored, current map[string]fileops.C
 	}
 
 	// Check for orphaned files (exist locally but not in source).
-	// Skip .state/ (local state) and the user-owned config.yaml: init.go
-	// scaffolds config.yaml after checksums so it stays user-owned rather than a
-	// synced methodology resource, so it is legitimately absent from source.
-	// Deleting it would wipe operator hooks such as hooks.approval_judge.command.
+	// The user-owned config.yaml never reaches this map: fileops.shouldSkipFile
+	// excludes it from checksum generation and source listing, so it is never
+	// tracked, updated, or deleted. .state/ is filtered here for the same reason.
 	for path := range current {
-		if sourceFilesSet[path] ||
-			path == config.WorkspaceConfigFileName ||
-			path == workspace.StateDir ||
-			strings.HasPrefix(path, workspace.StateDir+"/") {
-			continue
+		if !sourceFilesSet[path] && !strings.HasPrefix(path, workspace.StateDir+"/") && path != workspace.StateDir {
+			changes.orphaned = append(changes.orphaned, path)
 		}
-		changes.orphaned = append(changes.orphaned, path)
 	}
 
 	return changes
--- a/internal/commands/system/update.go
+++ b/internal/commands/system/update.go
@@ -348,12 +348,19 @@ func categorizeChanges(ctx context.Context, stored, current map[string]fileops.C
 		}
 	}
 
-	// Check for orphaned files (exist locally but not in source)
-	// Skip .state/ directory which contains local state files
+	// Check for orphaned files (exist locally but not in source).
+	// Skip .state/ (local state) and the user-owned config.yaml: init.go
+	// scaffolds config.yaml after checksums so it stays user-owned rather than a
+	// synced methodology resource, so it is legitimately absent from source.
+	// Deleting it would wipe operator hooks such as hooks.approval_judge.command.
 	for path := range current {
-		if !sourceFilesSet[path] && !strings.HasPrefix(path, workspace.StateDir+"/") && path != workspace.StateDir {
-			changes.orphaned = append(changes.orphaned, path)
+		if sourceFilesSet[path] ||
+			path == config.WorkspaceConfigFileName ||
+			path == workspace.StateDir ||
+			strings.HasPrefix(path, workspace.StateDir+"/") {
+			continue
 		}
+		changes.orphaned = append(changes.orphaned, path)
 	}
 
 	return changes
--- a/tests/integration/system_update_test.go
+++ b/tests/integration/system_update_test.go
@@ -152,3 +152,56 @@ func TestSystemUpdate_ShowsOrphanedInDryRun(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, orphanExists, "orphaned file should still exist after dry-run")
 }
+
+// TestSystemUpdate_PreservesUserConfig verifies that `fest system update --force`
+// never deletes the user-owned .festival/config.yaml. init.go scaffolds it after
+// checksums so it is deliberately absent from source templates; it must not be
+// treated as an orphan. Deleting it would wipe operator hooks such as
+// hooks.approval_judge.command, the very config non-interactive judging relies on.
+func TestSystemUpdate_PreservesUserConfig(t *testing.T) {
+	tc := GetSharedContainer(t)
+
+	parentDir := "/sysupdate-preserve-config-test"
+	sourceDir := "/root/.obey/fest/festivals"
+	workspaceDir := parentDir + "/festivals" // init creates this
+	configPath := workspaceDir + "/.festival/config.yaml"
+
+	// Minimal source templates. config.yaml is user-owned and never shipped as a
+	// methodology template, so remove any stray copy from the shared source to
+	// deterministically reproduce a real install where source lacks config.yaml.
+	_, err := tc.runCommand([]string{"sh", "-c", fmt.Sprintf(`
+		mkdir -p %s/.festival/templates/festival
+		echo "# Festival Goal" > %s/.festival/templates/festival/GOAL.md
+		rm -f %s/.festival/config.yaml
+	`, sourceDir, sourceDir, sourceDir)})
+	require.NoError(t, err, "failed to create source templates")
+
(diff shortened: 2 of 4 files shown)


internal/commands/system/update.go
```diff
@@ -348,12 +348,19 @@ func categorizeChanges(ctx context.Context, stored, current map[string]fileops.C
...
+	// Check for orphaned files (exist locally but not in source).
+	// Skip .state/ (local state) and the user-owned config.yaml: init.go
+	// scaffolds config.yaml after checksums so it stays user-owned rather than a
+	// synced methodology resource, so it is legitimately absent from source.
+	// Deleting it would wipe operator hooks such as hooks.approval_judge.command.
 	for path := range current {
-		if !sourceFilesSet[path] && !strings.HasPrefix(path, workspace.StateDir+"/") && path != workspace.StateDir {
-			changes.orphaned = append(changes.orphaned, path)
+		if sourceFilesSet[path] ||
+			path == config.WorkspaceConfigFileName ||
```
obey-agent: Protecting `config.yaml` only in the orphan loop does not keep it user-owned. `GenerateChecksums` at the end of every successful update still records it, so after one update an operator edit lands in `changes.modified`; a second `--force` then attempts `updater.UpdateFile("config.yaml")`. I reproduced this in a disposable container: `.fest-checksums.json` contained `config.yaml`, and the next forced update emitted `Failed to update config.yaml: source file not found`. If a stale source ever contains that path, it can overwrite the operator config. Please exclude this path from every categorization and checksum-persistence step, not only orphan deletion, and add a two-update regression with an edit between runs.

tests/integration/system_update_test.go
```diff
@@ -152,3 +152,56 @@ func TestSystemUpdate_ShowsOrphanedInDryRun(t *testing.T) {
...
+	_, err = tc.runCommand([]string{"sh", "-c", fmt.Sprintf(`cat > %s <<'EOF'
+version: "1.0"
+hooks:
+  approval_judge:
+    command: /bin/false
+EOF`, configPath)})
+	require.NoError(t, err, "failed to write operator config")
+
+	// Run the destructive path.
+	_, err = tc.RunFestInDir(workspaceDir, "system", "update", "--force")
```
obey-agent: The destructive operation under test returns an error here, but the test discards it. Any early failure before orphan cleanup leaves `config.yaml` in place and makes this regression pass without exercising the fix. Please capture the output and require `NoError` with that output in the failure message; this fixture is deterministic. If a non-zero exit is genuinely expected, assert a specific marker that proves the deletion phase actually ran.

obey-agent (commented):
## Verdict: Request Changes

### Overview

This PR addresses a real destructive bug and correctly reroutes the approval-readiness test through the operator-owned judge configuration. The immediate orphan-deletion case is fixed, and both focused integration tests pass, but the user-owned configuration contract is not preserved across repeated updates and the new regression can false-pass on an early command failure.

### Key Findings

- `config.yaml` is exempted only from orphan deletion. A successful update still stores it in `.fest-checksums.json`, so the next operator edit is treated as modified methodology and `--force` attempts to update it from source. The inline comment includes a disposable-container reproduction and the requested two-update coverage.
- `TestSystemUpdate_PreservesUserConfig` discards the `system update --force` error, allowing an early failure before deletion to satisfy the preservation assertions. Require the operation to succeed or prove that the destructive phase ran.

### What Is Done Well

- The PR identifies a high-impact operator-data-loss path and uses the existing `WorkspaceConfigFileName` constant rather than another magic string.
- The approval-readiness test now exercises the intended non-interactive, operator-sanctioned hook path and keeps the symlink-containment assertion focused.
- The new preservation test uses the container harness, consistent with the repository filesystem-safety policy.

### Staff Standard

Not yet. I would be proud to submit the diagnosis and narrow containment fix, but I would not ship the repeated-update checksum behavior or a destructive regression test that ignores the command result. Addressing the two inline findings should bring this to production confidence.

lancekrogers:
Thanks for the review. Both findings addressed in c0f42c9.

**Finding 1 (checksum persistence, `update.go`).** You were right that protecting `config.yaml` only in the orphan loop was insufficient. Once an update recorded it in `.fest-checksums.json`, a later operator edit landed in `changes.modified` and the next `--force` called `UpdateFile("config.yaml")`, which either warned `source file not found` or, with a stale source, overwrote the operator config.

Rather than special-case it per categorization step, I excluded it at the single choke point: `fileops.shouldSkipFile`, which backs both `GenerateChecksums` and `ListFiles`. `config.yaml` is now never checksum-tracked, listed as a source file, categorized, updated, or deleted, exactly mirroring the existing `.state/` treatment. All four `GenerateChecksums` callers walk only the `.festival` tree, so this only ever affects `.festival/config.yaml`. I reverted the now-redundant orphan-loop special-case so the policy lives in one place.

I added `TestSystemUpdate_PreservesUserConfigAcrossUpdates`: two forced updates with an operator edit **and** a stale source `config.yaml` planted between them, asserting the operator's edit is never replaced by source content.

**Finding 2 (test discarded the error, `system_update_test.go`).** Fixed. The fixture is deterministic, so `TestSystemUpdate_PreservesUserConfig` now `require.NoError`s the update with the output in the failure message. It also plants a genuine orphan and asserts that orphan was deleted, proving the destructive phase actually ran rather than short-circuiting before it.

**Verification.** Both new tests were confirmed to **fail without the fix** (config.yaml deleted / overwritten) and pass with it. Full integration suite green (171s), host unit tests green, `just lint all` clean. The cross-test flakiness risk from shared source state is closed with a `t.Cleanup` that removes the planted stale source config even on assertion failure.

internal/commands/system/update.go
obey-agent: Re-review on c0f42c9: resolved. `config.yaml` is now excluded at the shared checksum/listing choke point, so it cannot enter current/source categorization or persisted checksums. The new two-update regression covers an operator edit plus a stale source config and passes in the container harness.

tests/integration/system_update_test.go
obey-agent: Re-review on c0f42c9: resolved. The update now requires success with captured output, and the planted orphan assertion proves the destructive cleanup phase ran before the config-preservation assertion.

obey-agent (commented):
## Verdict: Approve

## Overview
Re-review of c0f42c9 confirms that both prior blockers are resolved. The user-owned workspace configuration is now excluded at the shared checksum and source-listing choke point, and the regressions prove both destructive cleanup execution and preservation across repeated updates.

## Key Findings
- No new blocking findings.
- The checksum-persistence thread is resolved: `config.yaml` no longer enters generated checksums or source file lists, including when a stale source copy exists.
- The false-positive test thread is resolved: update success is required and a genuine orphan deletion proves the destructive phase ran.

## Done Well
- The policy is centralized in the existing `shouldSkipFile` path used by both checksum generation and source listing.
- The two-update regression directly exercises the dangerous operator-edit scenario and cleans shared source state with `t.Cleanup`.
- Focused container tests, affected unit tests, integration-tagged vet, and the quick build all pass on the PR head.

## Staff Standard
Yes. I would be proud to submit this remediation as my own work; it closes the data-loss path at the correct boundary and carries production-relevant regression coverage.

PR #299, opened 2026-07-26
fix(workflow): stop silently dropping evidence bullets the parser cannot read

lancekrogers:
An `**Evidence:**` bullet the parser cannot read as a path disappears without a word, and the judge gets launched against whatever is left.

## What happens today

`evidenceItemRe` is `^\s*[-*]\s+(\S+)\s*$` — one non-whitespace token on its own line. The gate scaffold ships this:

```markdown
**Evidence:**
- PHASE_GOAL.md
- (attach each sequence's SEQUENCE_GOAL.md and task outputs relevant to this gate step)
```

The second bullet has spaces, so it matches nothing and is dropped. What survives is `["PHASE_GOAL.md"]`, which looks exactly like a deliberate one-file evidence list. Readiness passes it. The judge is launched holding a **goal template whose boxes are all unchecked** and no deliverables.

It then rejects, correctly, for missing evidence it was never handed:

> *"The submission contains only planning documents (PHASE_GOAL.md and three SEQUENCE_GOAL.md files) with no actual deliverables; PHASE_GOAL.md shows all required outcomes unchecked."*

That is a wasted judge round-trip on the **first run of every phase**, and it reads as the judge being wrong when the gate was simply never filled in. I hit it three times running festival MN0001 end to end.

## The fix

Record the bullets the parser could not read, and refuse before the judge starts, with the offending line in the message:

```
approval readiness failed: evidence list has entries that are not paths:
(attach each sequence's SEQUENCE_GOAL.md and task outputs relevant to this gate step)
Hint: each **Evidence:** bullet must be a single phase-relative path with no spaces;
replace the listed lines in GATES.md with the real deliverables, then re-submit with
'fest workflow judge'
```

The rule is **general, not placeholder-specific**: it also catches a path containing a space and a typo'd bullet, which failed the same silent way.

Parser output on the real scaffold block, before any readiness check:

```
paths    = ["PHASE_GOAL.md"]
unparsed = ["(attach each sequence's SEQUENCE_GOAL.md and task outputs relevant to this gate step)"]
```

## Ordering

The new guard sits after `checkAutoJudgeAllowed`, so an operator-attestation checkpoint is still refused first regardless of its evidence list. A fully filled list is unaffected, which is pinned by a test.

## Tests

`just gate` green: 2686.

- `TestEvidenceBulletsAreNeverSilentlyDropped`: the scaffold placeholder is reported not dropped; a path with a space is reported; a filled list reports nothing unparsed.
- `TestReadinessRefusesUnfilledEvidence`: an unfilled gate never reaches the judge, and the error names both the offending line and `GATES.md`.
- `TestReadinessAllowsAFilledEvidenceList`: no regression for filled gates.

## Note on the other half

I also filed a separate observation in that festival's feedback about the judge-in-flight notice: `fest next` prints "Judge launched (pid N)" *above* the `EXECUTE THE FOLLOWING INSTRUCTIONS` banner, where an agent reading the tail of a long block never sees it, so it stops at a checkpoint whose verdict is already coming. Not fixed here; this PR is scoped to the evidence defect.

### DIFF
--- a/internal/commands/workflow/approval_readiness_unfilled_test.go
+++ b/internal/commands/workflow/approval_readiness_unfilled_test.go
@@ -1,6 +1,7 @@
 package workflow
 
 import (
+	"context"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


internal/guidance/workflow/parser.go
```diff
@@ -252,6 +259,34 @@ func (p *Parser) parseEvidencePaths(section string) []string {
...
+func (p *Parser) parseUnparsedEvidence(section string) []string {
+	m := evidenceRe.FindStringSubmatch(section)
+	if len(m) < 2 {
+		return nil
+	}
+	var unparsed []string
+	for _, item := range evidenceBulletRe.FindAllStringSubmatch(m[1], -1) {
+		if len(item) < 2 {
+			continue
+		}
```
obey-agent, line 263: **[suggestion]** `parseUnparsedEvidence` reimplements "not a single-token path" as `strings.ContainsAny(text, " \t")` rather than classifying against `evidenceItemRe` (or returning paths and unparsed from one pass). Today the space heuristic is equivalent for current regexes, but if `evidenceItemRe` ever changes (quoted paths, trailing comments, optional anchors), unparsed detection can drift and reintroduce silent drops or false positives without a compile-time link between the two.

**Suggestion:** Prefer a single scan over Evidence bullets: if `evidenceItemRe` matches, append to paths; else append the bullet text to unparsed. That keeps the readiness invariant "every bullet is either a path or reported" tied to one definition of "path."

internal/guidance/workflow/parser.go
```diff
@@ -37,9 +37,15 @@ var (
 	// evidenceRe matches an explicit evidence list for approval judging.
 	evidenceRe = regexp.MustCompile(`(?s)\*\*Evidence:\*\*\s*(.+?)(?:\n\n|\n\*\*|---|\n##|$)`)
 
-	// evidenceItemRe matches bullet evidence paths.
+	// evidenceItemRe matches bullet evidence paths: a single token on its own
```
obey-agent: **[suggestion]** The new comment claims "Anything else under **Evidence:** is captured by evidenceBulletRe instead and reported, rather than vanishing." That overstates coverage: only `[-*]` bullets with internal spaces/tabs are reported. Non-bullet lines, `+` list markers, and empty/malformed bullet lines under `**Evidence:**` still match neither regex and still vanish. The PR’s scaffold defect is fixed; the comment can mislead the next change.

**Suggestion:** Narrow the comment to multi-token (or non-`evidenceItemRe`) bullets, or extend classification to any non-empty line inside the Evidence block if full no-silent-drop is the goal.

internal/commands/workflow/approval_readiness_unfilled_test.go
```diff
@@ -0,0 +1,51 @@
...
+	wf "github.com/Obedience-Corp/fest/internal/guidance/workflow"
+)
+
+// TestReadinessRefusesUnfilledEvidence pins that an unfilled gate never reaches
+// the judge. Before this, the scaffold placeholder was dropped by the parser, the
+// remaining entry looked like a deliberate list, and the judge was launched with
+// a goal template and no deliverables. It rejected, correctly, for missing
+// evidence it was never given, which costs a full round-trip on the FIRST run of
+// every phase and reads as the judge being wrong.
+func TestReadinessRefusesUnfilledEvidence(t *testing.T) {
```
obey-agent, line 17: **[nit]** Readiness tests construct `WorkflowStep` by hand and parser tests call `parseEvidencePaths` / `parseUnparsedEvidence` separately. Nothing asserts the full parse → readiness pipeline for a raw GATES.md section (especially Evidence that is only unparsed bullets and no valid paths).

**Suggestion:** Optional one end-to-end case: `ParseContent` on a scaffold-like section, then `checkApprovalReadinessWithInspector` on the resulting step, so field wiring cannot regress independently of either unit.

obey-agent (commented):
## Summary

PR #299 fixes a real, well-observed failure mode: `evidenceItemRe` only accepts a single non-whitespace token per Evidence bullet, so scaffold placeholders (and paths with spaces) vanished, readiness saw a partial list as complete, and the judge burned a round-trip on missing deliverables it was never given. The approach is sound: record unreadable bullets on `WorkflowStep.EvidenceUnparsed`, refuse in `checkApprovalReadinessWithInspector` after `checkAutoJudgeAllowed`, and pin both parser and readiness with focused tests. Ordering, error text (including `GATES.md` via `WithHint`), and the judge launch paths that call readiness/`checkAutoJudgePreflight` all look correct. No correctness bugs found; remaining notes are maintainability and residual silent-drop edges.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

PR #300, opened 2026-07-26
feat(workflow): send working dirs and campaign root to the approval judge

lancekrogers:
> Pairs with a change in `obey` (`ob judge` must read the new fields). This PR is the sending half; it is additive and safe to merge alone.

A judge evaluating an implementation phase **cannot see the implementation.**

## The gap

`approvalJudgeRequest` carries `festival_path` and `phase_path`. Implementation phases produce code in `projects/*`. Nothing in the request mentions that directory exists.

So the judge is asked *"does the implementation satisfy the PHASE_GOAL objectives?"* while structurally unable to open one file of what was built. It cannot check the branch, read the diff, confirm the gate passed, or see that PRs exist.

What it can read is whatever the executor wrote **into the festival** about its own work. Driving festival MN0001 to completion, every implementation verdict rested on `results-summary.md` and `results/*.md` — prose I authored describing a repository the request never mentioned. The judge approved my account of the work rather than the work. That is a fair judge doing its best with an incomplete request.

## The change

Two additive fields:

```go
CampaignRoot string            `json:"campaign_root,omitempty"`
WorkingDirs  []judgeWorkingDir `json:"working_dirs,omitempty"`
```

`working_dirs` is derived from each sequence's `fest_working_dir` and **tagged with the sequence that declared it**, so a phase spanning two repositories tells the judge which deliverable belongs where. Absolute paths are resolved so the judge can open them directly.

Against the real MN0001 festival:

```
campaign_root = "/Users/lancerogers/Dev/AI/obey-campaign"
004_IMPLEMENT_NAVIGATION working_dirs = [
  {"sequence":"01_origin_threading","path":"projects/camp","absolute_path":"/Users/.../projects/camp"},
  {"sequence":"02_adopt_origin","path":"projects/camp","absolute_path":"/Users/.../projects/camp"},
  {"sequence":"03_skew_visibility","path":"projects/camp","absolute_path":"/Users/.../projects/camp"}
]
```

## Deliberate choices

- **Best-effort collection.** An unreadable or malformed sequence is skipped rather than failing the checkpoint. A judge with partial context beats a gate that will not run.
- **Normalization is a gate, not a formality.** A working dir that is absolute or escapes the campaign root is rejected instead of being handed over as a path to open.
- **A phase that declares none sends none.** Design and docs phases keep today's behavior exactly; the festival is already visible through `phase_path`.
- **Additive to `fest.approval.judge/v1`.** An older judge ignores both fields.

## Tests

`just gate` green: 2687.

`TestCollectPhaseWorkingDirs` (multi-repo phase, sequence tagging, hidden dirs and loose files ignored), `TestCollectPhaseWorkingDirsIsBestEffort` (malformed YAML and a traversing path both skipped, the good one survives), `TestCollectPhaseWorkingDirsUnreadablePhase`, `TestCampaignRootFor` (found, and empty rather than guessed when absent).

## The obey half

`ob judge` currently reads `document` plus `evidence`. To use this it needs to read `working_dirs` and actually inspect them — at minimum branch and dirty state per dir, and the ability to read files there when the step's criteria are about code. I will open that separately.

## Related

[fest#299](https://github.com/Obedience-Corp/fest/pull/299) fixes the adjacent defect where evidence bullets the parser could not read were silently dropped, so the judge was launched holding only an unchecked goal template.

### DIFF
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -86,6 +86,19 @@ type approvalJudgeRequest struct {
 	// EvidenceFiles is optional embedded content when the resolved hook uses
 	// evidence: embed. Omitted in the default paths mode.
 	EvidenceFiles []hooks.EvidenceFile `json:"evidence_files,omitempty"`
+	// CampaignRoot is the campaign directory containing this festival, so a
+	// judge can resolve working dirs and read the repositories the phase
+	// actually changed. Empty when no campaign root is found.
+	CampaignRoot string `json:"campaign_root,omitempty"`
+	// WorkingDirs are the fest_working_dir declarations of the phase's
+	// sequences: where the work landed, as opposed to where the plan lives.
+	//
+	// Without these a judge evaluating an implementation phase can only read
+	// the executor's own account of what it did, because the deliverable is
+	// code in another repository that the request never mentioned. Additive to
+	// fest.approval.judge/v1: an older judge ignores the field and behaves
+	// exactly as before.
+	WorkingDirs []judgeWorkingDir `json:"working_dirs,omitempty"`
 }
 
 type approvalJudgeResponse struct {
@@ -800,6 +813,11 @@ func judgeApproval(ctx context.Context, nav *wf.Navigator, step wf.WorkflowStep,
 		Output:        step.Output,
 		Checkpoint:    step.Checkpoint.String(),
 	}
+	// Where the work landed, not just where the plan lives. Best-effort: a
+	// phase with no declared working dirs (design, docs) simply sends none, and
+	// the judge sees the festival through phase_path as before.
+	req.CampaignRoot = campaignRootFor(nav.Ctx.FestivalPath)
+	req.WorkingDirs = collectPhaseWorkingDirs(nav.Ctx.PhasePath, req.CampaignRoot)
 	if nav.IsGate() {
 		req.Document = "GATES.md"
 	}
--- a/internal/commands/workflow/approve.go
+++ b/internal/commands/workflow/approve.go
@@ -89,6 +89,11 @@ type approvalJudgeRequest struct {
 	// CampaignRoot is the campaign directory containing this festival, so a
 	// judge can resolve working dirs and read the repositories the phase
 	// actually changed. Empty when no campaign root is found.
+	//
+	// It is the one absolute path in the request, and it is here so the
+	// working dirs themselves can stay relative: an absolute path per working
+	// dir would put the operator's home directory and username into every
+	// judge prompt, provider log, ledger entry and transcript.
 	CampaignRoot string `json:"campaign_root,omitempty"`
 	// WorkingDirs are the fest_working_dir declarations of the phase's
 	// sequences: where the work landed, as opposed to where the plan lives.
@@ -816,8 +821,10 @@ func judgeApproval(ctx context.Context, nav *wf.Navigator, step wf.WorkflowStep,
 	// Where the work landed, not just where the plan lives. Best-effort: a
 	// phase with no declared working dirs (design, docs) simply sends none, and
 	// the judge sees the festival through phase_path as before.
-	req.CampaignRoot = campaignRootFor(nav.Ctx.FestivalPath)
-	req.WorkingDirs = collectPhaseWorkingDirs(nav.Ctx.PhasePath, req.CampaignRoot)
+	req.CampaignRoot = campaignRootFor(ctx, nav.Ctx.FestivalPath)
+	workingDirs, skipped := collectPhaseWorkingDirs(nav.Ctx.PhasePath)
+	req.WorkingDirs = workingDirs
+	reportWorkingDirSkips(os.Stderr, skipped, len(workingDirs))
 	if nav.IsGate() {
 		req.Document = "GATES.md"
 	}
--- a/internal/commands/workflow/judge_workdirs.go
+++ b/internal/commands/workflow/judge_workdirs.go
@@ -1,0 +1,107 @@
+package workflow
+
+import (
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/Obedience-Corp/fest/internal/frontmatter"
+	"github.com/Obedience-Corp/fest/internal/pathutil"
+)
+
+// judgeWorkingDir is one place a phase's work actually landed.
+//
+// It exists because a judge that cannot see the working directory is being
+// asked whether an implementation satisfied its goal while structurally unable
+// to look at the implementation. Implementation phases produce code in
+// projects/*, and the festival directory holds only the plan and whatever the
+// executor wrote about the work. Without this the judge can only evaluate the
+// executor's own account of what it did.
+type judgeWorkingDir struct {
+	// Sequence is the sequence directory that declared this working dir, so a
+	// judge can tell which deliverable belongs where when a phase spans repos.
+	Sequence string `json:"sequence"`
+	// Path is relative to the campaign root, exactly as fest_working_dir
+	// declares it (for example "projects/camp").
+	Path string `json:"path"`
+	// AbsolutePath is resolved for the judge's convenience; it is the campaign
+	// root joined with Path, and is empty when the campaign root is unknown.
+	AbsolutePath string `json:"absolute_path,omitempty"`
+}
+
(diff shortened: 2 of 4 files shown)


internal/commands/workflow/judge_workdirs.go
```diff
@@ -0,0 +1,107 @@
...
+		return dirs[i].Path < dirs[j].Path
+	})
+	return dirs
+}
+
+// campaignRootFor walks up from the festival path looking for the campaign
+// marker, so working dirs can be handed to the judge as absolute paths it can
+// actually open. Empty when no campaign root is found, which leaves the
+// relative paths intact and costs the judge only the join.
+func campaignRootFor(festivalPath string) string {
```
obey-agent: **[bug]** `campaignRootFor` reimplements `.campaign` discovery but does not match `workspace.DetectCampaign`: it ignores `CAMP_ROOT` (`workspace.EnvCampaignRoot`) and never `filepath.Abs`s the walk start. Absolute path claims for the judge depend on this root. With `CAMP_ROOT` set (or a relative `FestivalPath`), the request can omit `campaign_root` / leave `absolute_path` non-absolute or pointed at a different tree than the rest of fest’s campaign resolution, while `path` remains campaign-relative to the *other* root semantics.

**Suggestion:** Prefer `workspace.DetectCampaign(ctx, festivalPath)` (or share its walk + `CAMP_ROOT` handling), Abs the start path, and only fall back to empty when detection fails. Add a test that sets `CAMP_ROOT` and asserts the judge root follows it.

internal/commands/workflow/judge_workdirs.go
```diff
@@ -0,0 +1,107 @@
...
+		raw, err := os.ReadFile(goalPath)
+		if err != nil {
+			continue
+		}
+		fm, _, err := frontmatter.Parse(raw)
+		if err != nil || fm == nil {
+			continue
+		}
+		rel, err := pathutil.NormalizeWorkingDir(fm.WorkingDir)
+		if err != nil || rel == "" {
```
obey-agent: **[suggestion]** Best-effort `continue` on missing file, bad YAML, empty working dir, *and* normalization failure treats “declared none” the same as “declared invalid/escaping.” For an implementation sequence with a typo’d or absolute `fest_working_dir`, the bad value is silently dropped. If every sequence fails that way, `working_dirs` is omitted entirely and the judge again sees only festival prose—the exact incomplete-request failure this PR documents—without any stderr, audit, or readiness signal that context was stripped.

**Suggestion:** Keep fail-open for the gate, but distinguish “no working dir declared” from “declared but rejected/unreadable.” At minimum log or attach a non-fatal note (stderr warning, optional request field, or judge audit detail) when a non-empty `fest_working_dir` is skipped, especially when the result is zero working dirs for a phase that has sequences.

internal/commands/workflow/judge_workdirs.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		key := entry.Name() + "\x00" + rel
+		if _, dup := seen[key]; dup {
+			continue
+		}
+		seen[key] = struct{}{}
+
+		wd := judgeWorkingDir{Sequence: entry.Name(), Path: rel}
+		if campaignRoot != "" {
+			wd.AbsolutePath = filepath.Join(campaignRoot, rel)
```
obey-agent: **[suggestion]** `AbsolutePath` is `filepath.Join(campaignRoot, rel)` with no existence check and no post-resolve containment. `NormalizeWorkingDir` only gates the path *string* (no `..`). A campaign-relative entry that is a symlink to outside the campaign still yields an in-root `absolute_path` string; opening it follows the link. Evidence resolution already uses `EvalSymlinks` + within-root checks (`docs/concepts/hook-evidence-contract.md`); working dirs handed for the judge to open do not get the same treatment.

**Suggestion:** Optionally resolve with `EvalSymlinks` and verify the real path stays under the campaign root (same helper pattern as evidence), or document that containment is path-string only and that the obey judge must re-validate before open. Skipping non-existent dirs is optional; documenting that absence is allowed is fine.

obey-agent (commented):
## Summary

PR #300 additively extends `fest.approval.judge/v1` so the approval judge receives `campaign_root` and per-sequence `working_dirs` (relative `path` plus optional `absolute_path`), derived from each sequence’s `fest_working_dir`. That closes a real gap: implementation phases land code under `projects/*` while the request previously only pointed at festival/phase plan paths.

Path-string safety correctly reuses `pathutil.NormalizeWorkingDir` (absolute, `~`, and `..` rejected after clean). Best-effort collection (skip unreadable/malformed sequences rather than fail the checkpoint) matches the stated product tradeoff. Tests cover multi-repo tagging, hidden/non-sequence noise, traversal rejection, unreadable phase, and campaign marker walk.

Main concerns: `campaignRootFor` reimplements campaign detection without `CAMP_ROOT` / `filepath.Abs` (diverges from `workspace.DetectCampaign`), silent drops of declared-but-invalid working dirs can re-empty the payload with no operator signal, and `absolute_path` is a plain join without symlink containment checks that evidence paths already use.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 2

internal/commands/workflow/judge_workdirs_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if got[1].Sequence != "02_other_repo" || got[1].Path != "projects/fest" {
+		t.Errorf("second = %+v", got[1])
+	}
+	if got[0].AbsolutePath != filepath.Join(campaign, "projects/camp") {
+		t.Errorf("absolute path = %q", got[0].AbsolutePath)
+	}
+}
+
+func TestCollectPhaseWorkingDirsIsBestEffort(t *testing.T) {
```
obey-agent, line 86: **[suggestion]** `TestCollectPhaseWorkingDirsIsBestEffort` comment claims “absolute or traversing” is rejected, but the fixture only uses `../../etc`. There is also no coverage for: empty `campaignRoot` leaving `absolute_path` empty while still emitting relative `path`; absolute working dir (`/etc`); home-relative `~`; and no marshal/integration check that `approvalJudgeRequest` JSON includes `campaign_root` / `working_dirs` with the expected snake_case keys.

**Suggestion:** Extend the best-effort table with `/abs` and `~/x` cases; add `TestCollectPhaseWorkingDirsEmptyCampaignRoot`; optionally assert `json.Marshal` of a populated `approvalJudgeRequest` for the public contract the obey consumer will pin.

internal/commands/workflow/judge_workdirs.go
```diff
@@ -0,0 +1,107 @@
...
+		fm, _, err := frontmatter.Parse(raw)
+		if err != nil || fm == nil {
+			continue
+		}
+		rel, err := pathutil.NormalizeWorkingDir(fm.WorkingDir)
+		if err != nil || rel == "" {
+			continue
+		}
+		key := entry.Name() + "\x00" + rel
+		if _, dup := seen[key]; dup {
```
obey-agent: **[nit]** Dedup key `sequence + "\x00" + rel` is effectively dead for normal `ReadDir` input: directory names under a phase are unique, and each sequence contributes at most one `fest_working_dir`. Harmless, but it implies multi-value-per-sequence behavior that does not exist.

**Suggestion:** Drop the `seen` map unless a future multi-working-dir source is planned; or document why collision handling remains.

internal/commands/workflow/approve.go
```diff
@@ -86,6 +86,19 @@ type approvalJudgeRequest struct {
 	// EvidenceFiles is optional embedded content when the resolved hook uses
 	// evidence: embed. Omitted in the default paths mode.
 	EvidenceFiles []hooks.EvidenceFile `json:"evidence_files,omitempty"`
+	// CampaignRoot is the campaign directory containing this festival, so a
```
obey-agent, line 89: **[nit]** Additive JSON fields are correctly `omitempty` and schema version stays `fest.approval.judge/v1`, which is right for backward compatibility. The hook evidence contract doc still only describes `festival_path` / `phase_path` / `evidence` and does not mention `campaign_root` or `working_dirs`, so the public request shape for judge authors is incomplete once this lands.

**Suggestion:** Brief additive note in `docs/concepts/hook-evidence-contract.md` (or the approve command long help) listing the new optional fields and that older judges may ignore them—aligned with how `evidence_files` was documented.

PR #301, opened 2026-07-27
fix(types): discover real create types and fix misleading errors

lancekrogers:
## Demo

`fest types list` / `show` / error paths after this fix (VHS):

fest types command

Tape: `docs/demos/fest-types-command.tape`  
Reproduce: build branch `fest` into a fixture `bin/`, copy methodology templates, set `FEST_VHS_ROOT`, run `vhs docs/demos/fest-types-command.tape`.

## Summary

`fest types list` / `show` were effectively broken on a normal install: they looked under `~/.config/fest/templates` with a legacy `PHASE_*_TEMPLATE.md` filename grammar, while `fest system sync` installs a nested tree at `~/.obey/fest/festivals/.festival/templates`. Help and create errors also pointed at the empty half of the command surface.

This PR makes `fest types` answer “what can I pass to create?”:

- Resolve the real methodology templates root (campaign `festivals/.festival/templates`, then system-synced path; `FEST_TEMPLATES_DIR` still overrides)
- Discover nested packages (`phases/<type>/`, sequences, tasks, quality gates)
- Merge `festival_types.yaml` workflow types into `list` / `show` (festival level)
- Default `fest types` (no subcommand) → list
- Fix doubled `not found not found` and wrong recovery hints
- Drop phantom `quick` / dead examples from help and create flag text
- CLI reference docs aligned

Exploration notes: campaign workitem `workflow/explore/fest-types-command-2026-07-26` (WI-c4e524).

## Before / after

| Command | Before | After |
|---------|--------|--------|
| `fest types list` | “No template types found / run system sync” | Festival + phase + sequence + task types |
| `fest types show implementation --level phase` | not found | Phase package + create example |
| `fest types show feature` | `not found not found` + path hint | single not found + `fest types list` hint |
| `fest create festival --type quick` | hint: run `fest types list` (empty) | hint: `fest types list --level festival` |

## Test plan

- [x] `go test ./internal/types/ ./internal/commands/types/ ./internal/commands/festival/`
- [x] Manual: `fest types list` non-empty after build
- [x] Manual: `fest types show standard` / `show implementation --level phase`
- [x] VHS gif recorded and frames inspected (list populated; phase/standard show; feature error clean)


internal/types/registry.go
```diff
@@ -93,55 +119,234 @@ func (r *Registry) scanDirectory(ctx context.Context, dir string, isCustom bool,
...
+		if entry.IsDir() {
+			subDir := filepath.Join(dir, entry.Name())
+			// Do not re-enter methodology package roots via recursion from a parent.
+			if isMethodologyTemplatesRoot(subDir) {
+				if err := r.scanMethodologyRoot(ctx, subDir, isCustom, countMarkers); err != nil {
+					continue
+				}
+				continue
+			}
+			_ = r.scanLegacyFlatAndRecurse(ctx, subDir, isCustom, countMarkers)
```
obey-agent: **Error handling:** I/O failures from nested scans are discarded here.

```go
if err := r.scanMethodologyRoot(...); err != nil {
    continue // real IO/permission errors vanish
}
_ = r.scanLegacyFlatAndRecurse(...) // same
```

`Discover` already treats missing dirs as non-fatal; that is fine. Once a directory *exists*, a failed `ReadDir` / permission error should propagate (or be wrapped and returned), not silently produce a partial registry.

Suggestion: return non-`IsNotExist` errors from both branches so `fest types list` cannot look empty when the tree is unreadable.

obey-agent (commented):
## Verdict: **Approve with nits**

High-value, correctly scoped fix for a real launch-surface bug: `fest types` now answers “what can I pass to create?” against the actual methodology tree and `festival_types.yaml`, instead of an empty legacy path and phantom `quick` type.

## Overview

PR rewires type discovery from legacy flat `PHASE_*_TEMPLATE.md` under `~/.config/fest/templates` to the nested layout that `fest system sync` installs (`phases/<type>/`, `sequences/`, `tasks/`), merges festival workflow types into `list`/`show`, defaults bare `fest types` → list, fixes doubled `not found`, and aligns create-flag / CLI help with reality. Nested-layout unit tests pass; VHS tape + gif document the user-visible before/after.

## Key Findings

### Should fix before or immediately after merge

1. **Silent I/O error discard in recursive discovery** (`registry.go` scan helpers) — partial empty lists with no error when a present directory is unreadable.
2. **Incomplete docs regen for `quick`** — this PR correctly removes `quick` from create flags and several type docs, but leaves it in:
   - `docs/cli-reference/fest_types_festival.md`
   - `docs/cli-reference/fest_types_festival_show.md`
   - `docs/cli-reference/fest-reference.md` (create `--type` help + festival types synopsis)
   That reintroduces the exact misleading surface the PR is fixing. Re-run `fest gendocs` / full cli-reference regen.
3. **Missing tests for `getBuiltInTemplatesDir` resolution order** — this is the core path fix; registry nested tests alone won’t catch a regression to the wrong install root.

### Non-blocking

4. **`fest types festival show` example still uses positional name** (`fest create festival --type X <name>`) while create requires `--name` and unified `fest types show` was fixed to the flag form — two detail paths disagree.
5. **`registry.go` is 553 lines** (project guideline: &lt;500). New scan helpers are good candidates for `registry_scan.go`.
6. **Magic default** `typeName == "planning"` — matches create today; prefer a shared constant if you touch this again.
7. Path discovery intentionally differs from `template.LocalTemplateRoot` (types works outside `festivals/`); document or centralize so paths cannot drift again.

## What’s Done Well

- Real product diagnosis: wrong root + wrong grammar + wrong recovery hints + phantom `quick`.
- Nested methodology scan correctly ignores `festival/` scaffold as create `--type` values and routes workflow types through `MergeFestivalWorkflowTypes`.
- Context checks on new scan paths; project `errors` package used for I/O wraps.
- `errors.NotFound` double-suffix fixed by passing resource name only.
- Help/list empty-state tips now distinguish missing vs empty template dirs and point at `fest system sync` / `FEST_TEMPLATES_DIR`.
- Table-driven nested layout + merge tests; packages under test are green (`./internal/types/`, `./internal/commands/types/`, `./internal/commands/festival/`).
- Demo tape + gif make the user-visible contract reviewable.

## Staff Standard

**Meets staff bar for shipping the fix**, with docs completeness and error-propagation nits that should not be left to rot. Integration-over-duplication is mostly good (merge into existing Registry rather than a parallel catalog); path resolution is still a second source of truth next to `template.LocalTemplateRoot`/`create` — acceptable if documented, better if shared.

**Suggested follow-up (same PR or fast follow):**
1. Propagate non-`IsNotExist` scan errors.
2. Regenerate remaining cli-reference docs that still advertise `quick`.
3. Add path-resolution table tests.
4. Align `fest types festival show` usage line with `--name`.

internal/types/registry.go
```diff
@@ -93,55 +119,234 @@ func (r *Registry) scanDirectory(ctx context.Context, dir string, isCustom bool,
...
+				IsCustom:  isCustom,
+				IsDefault: typeName == "planning",
+				Templates: files,
+				Source:    pkgDir,
+				Markers:   markers,
+			})
+			// Quality gates under phases/<type>/gates/
+			gatesDir := filepath.Join(pkgDir, "gates")
+			if gInfo, gErr := os.Stat(gatesDir); gErr == nil && gInfo.IsDir() {
+				_ = r.scanLegacyFlatAndRecurse(ctx, gatesDir, isCustom, countMarkers)
```
obey-agent: **Same pattern:** quality-gate discovery under `phases/<type>/gates/` drops scan errors with `_ =`.

If gates are present but unreadable, list/show omit `gate/*` types with no signal. Prefer propagating the error rather than ignoring it — these are the types users need for gate-related flows.

internal/commands/types/paths.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/Obedience-Corp/fest/internal/config"
+	"github.com/Obedience-Corp/fest/internal/workspace"
+)
+
+// getBuiltInTemplatesDir returns the methodology templates tree used by create/sync.
+// Resolution order:
+//  1. FEST_TEMPLATES_DIR (explicit override)
+//  2. Campaign festivals/.festival/templates (when cwd is inside a campaign)
+//  3. System-synced ~/.obey/fest/festivals/.festival/templates
+func getBuiltInTemplatesDir() string {
```
obey-agent, line 22: **Test gap (high value):** this function is the actual product fix (wrong legacy path `~/.config/fest/templates` → real methodology tree). Nested-layout registry tests are good, but there is **no coverage of resolution order**:

1. `FEST_TEMPLATES_DIR` override
2. campaign `festivals/.festival/templates` when present
3. fallback `config.ConfigDir()/festivals/.festival/templates`
4. missing campaign dir falls through (does not hard-fail)

Without these, a future refactor can re-break `fest types list` on a normal install while unit tests stay green. Prefer table-driven tests with `t.Setenv` + `t.Chdir` into a fixture tree.

internal/commands/types/paths.go
```diff
@@ -0,0 +1,56 @@
...
+	if dir := os.Getenv("FEST_TEMPLATES_DIR"); dir != "" {
+		return dir
+	}
+
+	if cwd, err := os.Getwd(); err == nil {
```
obey-agent, line 27: **Integration note (non-blocking):** create paths use `template.LocalTemplateRoot` / `FindFestivalsRoot` (must be *inside* `festivals/`), while types discovery uses `workspace.FindFestivals` + `config.ConfigDir()` fallback.

That is probably intentional so `fest types` works from campaign root and outside any campaign. Worth a one-line comment that the divergence is deliberate, and/or a shared `MethodologyTemplatesDir(cwd)` helper so the three call sites cannot drift again the way `~/.config/fest/templates` did.

internal/types/registry.go
```diff
@@ -93,55 +119,234 @@ func (r *Registry) scanDirectory(ctx context.Context, dir string, isCustom bool,
...
+			pkgDir := filepath.Join(phasesDir, typeName)
+			files, markers := collectMarkdownPackage(pkgDir, countMarkers)
+			if len(files) == 0 {
+				continue
+			}
+			r.addType(&TypeInfo{
+				Name:      typeName,
+				Level:     LevelPhase,
+				IsCustom:  isCustom,
+				IsDefault: typeName == "planning",
```
obey-agent: **Nit / magic string:** `IsDefault: typeName == "planning"` hardcodes the default phase type. Matches `create phase --type` default today, but if defaults ever move to config this will drift silently.

Prefer a named constant shared with create (or derive default from the same source of truth as `create phase`). Low priority if keeping scope tight.

internal/commands/types/festival_types.go
```diff
@@ -22,14 +22,13 @@ func newFestivalCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "festival [type-name]",
 		Short: "Discover festival types",
-		Long: `Discover available festival types and their phase structures.
+		Long: `Discover available festival workflow types and their phase structures.
```
obey-agent, line 25: **Docs completeness:** this Long text correctly drops `quick`, but sibling generated docs were not fully regenerated:

- `docs/cli-reference/fest_types_festival.md` still lists `quick`
- `docs/cli-reference/fest_types_festival_show.md` still shows `show quick`
- `docs/cli-reference/fest-reference.md` still has `quick` in create `--type` help and the festival types synopsis

Please re-run full cli-reference generation so the misleading type does not remain discoverable in docs.

PR #303, opened 2026-07-29
fix(list): refresh watch without polling flashes

lancekrogers:
## Summary

- refresh `fest list --watch` from filesystem events so `fest task completed` updates festival percentages immediately
- watch the festival tree recursively, adding only newly created directory subtrees during normal event handling instead of walking the full tree after every write
- cache the rendered frame and redraw only when visible board content changes, eliminating idle and duplicate flashes
- run a cached 30-second safety reconciliation to recover missed watches without repainting unchanged output, while retaining the same interval as the fallback when filesystem notifications are unavailable
- add containerized TTY coverage for idle stability, the real `[0%]` to `[100%]` task-completion flow, and a valid `ready → active` lifecycle transition

## Demo

Full-color split-pane VHS demo of fest list --watch updating from 0 percent to 100 percent after fest task completed

The upper pane runs `fest list --watch` and holds the active festival at `[0%]`. The lower pane visibly runs `fest task completed --yes`, reports the task completed, and the upper pane immediately redraws that same festival at `[100%]`. The deterministic tape and fixture are included under `docs/demos/`.

## Validation

- `just test integration-dev-list-watch` — verifies idle watch renders once, `fest task completed 01_first_task.md --yes` refreshes `[0%]` to `[100%]`, and `fest status set active --force --no-commit` immediately refreshes ready-list membership
- `go test -race ./internal/watch ./internal/commands/list`
- `just test unit` — 2,702 tests passed
- `just lint vet`
- `just lint all` — 0 issues
- `just vhs doctor`
- recorded and visually inspected the full-color split-pane GIF before and after task completion


internal/commands/list/list_watch.go
```diff
@@ -121,8 +99,7 @@ func renderListFrame(ctx context.Context, festivalsDir, filterStatus string, opt
 }
 
 // listWatchPaths returns non-recursive fsnotify paths covering status buckets
-// where festivals appear, move, or leave. Deep progress file changes are
-// covered by the polling interval in runListWatch.
+// where festivals appear, move, or leave.
 func listWatchPaths(festivalsDir string) []string {
```
obey-agent: **[suggestion]** Removing the 2s hybrid ticker makes refresh fully dependent on non-recursive watches of status buckets (`active`, `ready`, `dungeon/completed`, …). Working-status transitions remain fine because festivals are direct children of those dirs, so a leave/create always fires. Dungeon festivals, however, live under dated buckets (`dungeon/completed/YYYY-MM-DD/<fest>`). Moves that never touch a watched parent’s direct children — e.g. completed↔archived when both date dirs already exist, or a second festival dropped into an existing date dir without a leave from a working status — no longer produce an fsnotify event, so `fest list all --watch` / `fest list dungeon --watch` can stay stale until another event or restart. The old poller papered this over every 2s.

**Suggestion:** On Create under a dungeon status dir, `AddPath` the new date bucket (and/or watch a shallow recursive set of known date dirs). Alternatively document that dungeon-nested lifecycle is best-effort under pure event mode, or keep a much slower safety poll only when `--all`/dungeon filters are active.

tests/integration/fest_list_watch_test.go
```diff
@@ -70,20 +70,32 @@ func TestFestListWatchNonTTYFailsFast(t *testing.T) {
...
+
+func TestFestListWatchRefreshesOnFestivalStatusChange(t *testing.T) {
 	container := GetSharedContainer(t)
 	workspaceRoot, activePath, _ := setupListWatchFixture(t, container)
+	readyPath := filepath.Join(workspaceRoot, "festivals", "ready", filepath.Base(activePath))
 
 	script := fmt.Sprintf(
-		"cd %s && timeout 7s /fest list active --watch & watchpid=$!; sleep 3; (cd %s && /fest progress --task 001_IMPLEMENT/01_core/01_first_task.md --complete >/dev/null 2>&1); wait $watchpid",
-		shellQuote(workspaceRoot), shellQuote(activePath),
+		"cd %s && timeout 5s /fest list active --watch & watchpid=$!; sleep 1; mv %s %s; wait $watchpid",
```
obey-agent: **[suggestion]** The status-change test starts the watch, sleeps only 1s, then `mv`s the festival and requires both an initial `ACTIVE Festivals (1)` frame and exactly two screen clears. The previous progress test used a 3s settle window. Under a slow container/binary start, the move can race the first paint (missing `(1)`, wrong clear count). Separately, `require.Equal(t, 2, strings.Count(...\x1b[H\x1b[2J))` will flake if inotify delivers source/dest events outside the 100ms debounce window (two debounced paints) or if the shared `/workspace` festivals tree emits an extra event.

**Suggestion:** Wait for the first clear/`ACTIVE Festivals (1)` before moving (poll loop), keep a slightly longer settle (2–3s), and soften the clear-count check to something like `require.GreaterOrEqual(t, clears, 2)` plus content assertions, or assert `clears == 2` only after confirming no other writers touch the fixture tree.

tests/integration/fest_list_watch_test.go
```diff
@@ -70,20 +70,32 @@ func TestFestListWatchNonTTYFailsFast(t *testing.T) {
...
 
-func TestFestListWatchProgressUpdatesBetweenTicks(t *testing.T) {
+func TestFestListWatchDoesNotRedrawWhileIdle(t *testing.T) {
+	container := GetSharedContainer(t)
+	workspaceRoot, _, _ := setupListWatchFixture(t, container)
+
+	output, exitCode := runFestListWatchBounded(t, container, workspaceRoot, 4, "active", "--watch")
+
+	require.Contains(t, []int{124, 143}, exitCode, "watch process should be stopped by the bounded timeout")
+	require.Equal(t, 1, strings.Count(output, "\x1b[H\x1b[2J"), "an idle watch should render only its initial frame")
```
obey-agent, line 80: **[nit]** Idle stability is asserted only for a quiet 4s window with no concurrent mutations. That matches the new “no ticker” contract, but the PR’s main user-visible change is that deep task/progress edits must not flash the board. There is no regression test that `fest progress --complete` (or a direct task-file write) leaves the clear count at 1.

**Suggestion:** Add a companion test that completes a task while `list --watch` runs and asserts still a single `\x1b[H\x1b[2J`, mirroring the removed progress-update test as a negative case.

obey-agent (commented):
## Summary

This PR correctly converts `fest list --watch` from a hybrid fsnotify+2s-progress-poll loop into an event-driven lifecycle board: non-recursive status-bucket watches drive redraws, the progress ticker is removed, and the polling fallback is slowed to 30s for degraded environments only. The primary lifecycle path (festivals as direct children of `planning`/`ready`/`active`/`ritual`, and leaves from those dirs into dungeon) is covered well; debounce/`dispatchMu` still coalesce rapid rename pairs and prevent overlapping paints. Dominant residual risk is nested dungeon date-bucket moves (and late-created status dirs) that the old poller used to catch, plus slightly brittle exact-frame-count integration assertions.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

tests/integration/fest_list_watch_test.go
```diff
@@ -70,20 +70,31 @@ func TestFestListWatchNonTTYFailsFast(t *testing.T) {
...
 		shellQuote(workspaceRoot), shellQuote(activePath),
 	)
 
 	output, exitCode := runContainerScriptTTY(t, container, script)
 
 	require.Contains(t, []int{124, 143}, exitCode, "watch process should be stopped by the bounded timeout")
 	require.Contains(t, output, "[0%]", "initial frame should render 0%% before the task is completed")
-	require.Contains(t, output, "[100%]", "a later frame should render 100%% after fest progress --complete")
+	require.Contains(t, output, "[100%]", "task completion should refresh the board to 100%%")
+	require.Equal(t, 2, strings.Count(output, "\x1b[H\x1b[2J"), "one task completion should produce one visible refresh")
```
obey-agent: **[suggestion]** `TestFestListWatchRefreshesOnTaskProgressChange` still requires an exact clear-screen count of 2. Frame caching makes this much safer than the prior status-move test (duplicate events with identical board text no longer clear), and settle is now 2s, but the assertion remains timing-coupled: a missed/late first paint, an extra content-changing intermediate frame (if task completion ever surfaces a mid-progress percentage before 100%), or slow container startup can still fail without improving product confidence beyond the `[0%]`/`[100%]` checks.

**Suggestion:** Keep the percentage assertions as the primary signal; soften the clear count to `>= 2` (and optionally `<=` a small upper bound), or poll until `[100%]` appears with a deadline instead of a fixed `sleep 2` + exact equality.

tests/integration/fest_list_watch_test.go
```diff
@@ -70,20 +70,31 @@ func TestFestListWatchNonTTYFailsFast(t *testing.T) {
...
+	container := GetSharedContainer(t)
+	workspaceRoot, _, _ := setupListWatchFixture(t, container)
+
+	output, exitCode := runFestListWatchBounded(t, container, workspaceRoot, 4, "active", "--watch")
+
+	require.Contains(t, []int{124, 143}, exitCode, "watch process should be stopped by the bounded timeout")
+	require.Equal(t, 1, strings.Count(output, "\x1b[H\x1b[2J"), "an idle watch should render only its initial frame")
+}
+
+func TestFestListWatchRefreshesOnTaskProgressChange(t *testing.T) {
```
obey-agent, line 83: **[suggestion]** The old lifecycle integration case (`mv` active → ready and assert board membership refresh) was fully replaced by the task-progress test. Flag/help text still promise refresh on “festival progress or lifecycle status changes,” and status-bucket paths remain in `listWatchPaths`, but nothing in this suite verifies a status-directory move still redraws after the pure event-driven redesign (no hybrid ticker safety net).

**Suggestion:** Restore a second integration test that moves a festival between lifecycle buckets and asserts the filtered board updates (membership/headers), ideally with the same content-cache-aware clear expectations as the progress test.

internal/commands/list/list_watch.go
```diff
@@ -49,6 +47,10 @@ func runListWatch(ctx context.Context, festivalsDir, filterStatus string, opts *
 			fmt.Fprintf(os.Stderr, "%s file watch error: %v\n", ui.Warning("Warning:"), err)
 		},
 	}, func() {
+		// A create event on an already-watched parent can introduce a new phase,
+		// sequence, or task directory. Reconcile before rendering so subsequent
+		// writes inside that directory are observed too.
+		addListWatchPaths(w, festivalsDir)
```
obey-agent: **[suggestion]** Every debounced event re-runs `listWatchPaths` → full `filepath.WalkDir` over the festivals tree before paint. Correct for discovering new phase/sequence dirs, but on large workspaces or write storms (`MaxWait` 500ms) this is repeated full-tree work on the watcher’s serialized callback path. There is also no low-rate hybrid poll: with the new frame cache, a slow secondary tick would not flash and would recover from rare missed fsnotify events; pure event mode loses that belt-and-suspenders path that the old 2s ticker provided.

**Suggestion:** Consider incremental reconcile (only `AddPath` the created child from the event, or walk only under the event path), and/or a slow optional poll (e.g. 30s) in the event-driven path that reuses `listFrameSession` so missed events still converge without idle flashes.

obey-agent (commented):
## Summary

Follow-up commits intentionally re-scope `fest list --watch` to refresh on **festival progress and lifecycle** changes (docs, flag help, demo, and tests all agree). The head is a solid redesign: drop the hybrid 2s ticker that caused idle flashes, watch the full directory tree (with dynamic `AddPath` reconcile on events), and skip `clear`/repaint unless `formatListBoard` output actually changes via `listFrameSession`. Prior issue 1 (dungeon date-bucket blind spots without the poller) is **fixed** by recursive `WalkDir` + parent status watches + `addListWatchPaths`. Prior issue 3 (negative “progress must not flash” test) is **superseded**—progress-driven refresh is now the product intent, and idle no-flash is covered by `TestFestListWatchDoesNotRedrawWhileIdle`. Prior issue 2 is only **partially fixed** (content cache + longer settle); exact clear-count still brittle. No correctness bugs found in the watch filter / debounce path for the stated design.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

PR #304, opened 2026-07-29
feat(fest): pack and unbundle Festival Bundles

lancekrogers:
## Summary
Adds `fest pack` and `fest unbundle` using `obey-shared/festivalbundle` **v0.4.6**.

- Ritual/festival kind inference from path and fest.yaml
- Subject metadata from fest.yaml
- Optional `--validate` after unbundle
- Unbundle does **not** run the festival

Part of festival FB0001 (festival-bundle-format).

## Test plan
- [x] `go build ./cmd/fest`
- [x] Pack/unbundle client-intake ritual dogfood


internal/commands/pack/pack.go
```diff
@@ -0,0 +1,224 @@
...
+func NewUnbundleCommand() *cobra.Command {
+	var (
+		dest     string
+		force    bool
+		noVerify bool
+		noRecv   bool
+		jsonOut  bool
+		validate bool
+	)
+	cmd := &cobra.Command{
```
obey-agent, line 115: This command has no scope annotation, so scope.Resolve defaults it to Workspace. NewPackCommand has the same omission. As a result, fest unbundle refuses to run outside an existing campaign or festivals tree before it even reads the bundle; I reproduced the not in a fest workspace error from /tmp. A portable unbundle command must be Global so it can create a festival in a fresh location, and pack should likewise derive everything from its explicit source.

internal/commands/pack/pack.go
```diff
@@ -0,0 +1,224 @@
...
+			info, err := festivalbundle.Unbundle(ctx, festivalPath, d, festivalbundle.UnbundleOptions{
+				Force:               force,
+				SkipVerify:          noVerify,
+				WriteReceivedRecord: !noRecv,
+			})
+			if err != nil {
+				return festerrors.Wrap(err, "unbundle")
+			}
+			if validate {
+				c := exec.CommandContext(ctx, "fest", "validate")
```
obey-agent: This launches whichever fest happens to be on PATH, not the executable currently running. A standalone/current build with no installed fest fails after extraction with executable file not found; a stale installed binary validates under different code. Reuse the in-process validation path or intentionally re-exec os.Executable so --validate means this version.

internal/commands/pack/pack.go
```diff
@@ -0,0 +1,224 @@
...
+				c.Dir = d
+				c.Stdout = cmd.OutOrStdout()
```
obey-agent: When --json and --validate are combined, validation writes human output to stdout before info.json. I reproduced a successful command whose output jq rejects at line 2. Preserve stdout as one JSON document, for example by sending validation diagnostics to stderr or returning structured validation in the JSON result.

obey-agent (changes requested):
Requesting changes on the new bundle commands. Focused tests and CI pass, but the implementation has three user-visible integration failures: the portable commands are unavailable outside an existing workspace, validation runs an arbitrary PATH binary rather than this executable, and JSON plus validation does not produce a JSON document. I reproduced all three behaviors. Please cover these flag combinations and global invocation in command tests.

lancekrogers:
Addressed review: pack/unbundle use `scope: global`; `--validate` runs in-process `validator.FullValidate` (not PATH `fest`); validation logs to stderr so `--json` remains a single JSON document. Added command tests.

PR #306, opened 2026-07-29
fix(commit): use FE reference prefix

lancekrogers:
## Summary

- replace Fest's standalone `OBEY-FE` commit-reference prefix with `FE`
- preserve campaign tag consolidation by extracting festival IDs from the new prefix
- update generated CLI references and commit-format documentation
- add regression coverage for `FE-CA0004` formatting

## Why

Fest was still exposing the retired Obey namespace in standalone references and post-commit output even though campaign commits already use consolidated name-style tags. The Festival component prefix should be `FE` directly.

## User impact

Standalone Fest commits and the `Task` field in commit output now use references such as `FE-CA0004`. Campaign commits continue to use consolidated tags such as `[obey-campaign:8deed8b4-FE-CA0004]`.

## Validation

- `just gate`
  - stable and dev builds
  - stable, dev, and integration vet
  - golangci-lint: 0 issues
  - 2,703/2,703 unit tests
  - 71/71 containerized integration tests

### DIFF
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -379,8 +379,7 @@ with only festival-scoped files staged (not git add -A).
 
 Use --no-root to skip the campaign root commit.
 
-Reference format: [OBEY-FE-{id}]
-  - OBEY: Obey workflow tool prefix
+Reference format: [FE-{id}]
   - FE: Festival component identifier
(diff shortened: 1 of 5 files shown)


obey-agent (approved):
Approved at f0abce4. The standalone FE prefix is updated consistently across formatting, campaign-tag extraction, generated references, and documentation; the repository has no remaining OBEY-FE uses. Focused commit tests pass and CI is green.

PR #307, opened 2026-07-29
feat(tui): multi-step festival create wizard and Create what? menu

lancekrogers:
## Summary

Human-first `fest create` TTY redesign for **WI-7acd71** (design: `workflow/design/fest-create-tui-2026-07-28`, explore evidence: `workflow/explore/fest-create-tui-ux-2026-07-24`).

- **Create what?** outer menu keeps Festival / Phase / Sequence / Task and adds **Standalone workflow** as a peer (stub + CLI hint; full WORKFLOW.md wizard deferred to intent).
- **Festival** path is a multi-step wizard: type (from `festival_types.yaml`) → name/goal → project → conditional seed → tags → confirm → create.
- Types are loaded from config (no phantom hardcoded `quick`).
- Seed only when the type has an ingest phase; ritual lands in `ritual/`.
- Success copy is role-split (you vs agent); no longer leads with hand-filling markers or creating auto phases that already exist.
- `shared.CreateFestivalOpts` carries `Project` / `Seed` / `SeedFile` into the create pipeline.
- `fest tui` collapses dual plan/quick festival entries into **Create…** (same menu).
- Agent/CLI flags (`--json`, `--agent`, `--project`, `--seed`, …) unchanged.

## Demos

- [docs/demos/fest-create-menu.gif](docs/demos/fest-create-menu.gif) — Create what? menu + workflow stub  
- [docs/demos/fest-create-festival-wizard.gif](docs/demos/fest-create-festival-wizard.gif) — standard multi-step create  

Tapes: `docs/demos/fest-create-menu.tape`, `docs/demos/fest-create-festival-wizard.tape`.

## Test plan

- [x] `go test ./internal/commands/tui/ ./internal/commands/ ./internal/commands/festival/`
- [x] CLI smoke: standard + project + seed creates `input_specs/seed.md`
- [x] CLI smoke: implementation + seed rejected (no ingest)
- [x] VHS recorded against disposable fixture with branch binary
- [ ] Manual TTY walk: `fest create` menu peers; festival wizard happy path

## Related

- Design workitem WI-7acd71  
- Deferred: standalone WORKFLOW.md create TUI intent (`fest-create-tui-standalone-workflowmd-…`)


internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
+func festivalTypeDest(ft *types.FestivalType) string {
+	if ft != nil && ft.Name == "ritual" {
+		return "ritual"
 	}
+	return "planning"
+}
 
-	opts := &shared.CreateFestivalOpts{Name: name, Goal: goal, Tags: tags, Type: festType, VarsFile: varsFile, Dest: "planning"}
-	return shared.RunCreateFestival(ctx, opts)
+func festivalTypeSupportsSeed(ft *types.FestivalType) bool {
```
obey-agent: **[bug]** `festivalTypeSupportsSeed` treats any phase with `Type == "ingest"` or name `INGEST` (including non-auto phases) as seedable, while `ingestAutoPhaseID` / `resolveAndValidateSeed` only accept an **auto** phase with `Type == "ingest"`. A custom type with a non-auto ingest phase (or a phase named INGEST with a different type) will show the seed step, then fail at submit with "festival type has no ingest phase to seed" after the user has already walked the wizard. The fallback copy at `festivalTypeSupportsSeedFallback` has the same logic.

**Suggestion:** Align TUI gating with the create pipeline—e.g. only return true when there is an auto phase with `Type == "ingest"` (mirror `ingestAutoPhaseID`). Drop the name-only `INGEST` match unless the pipeline also accepts it. Add a unit test for non-auto ingest → false.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
 		huh.NewGroup(
-			huh.NewInput().Title("Festival name").Placeholder("e.g., ecommerce-platform").Value(&name).Validate(notEmpty),
+			huh.NewNote().
+				Title("New Festival · Identity").
+				Description(ctxLine),
+			huh.NewInput().
+				Title("Name *").
+				Placeholder("e.g., ecommerce-mvp").
+				Value(&name).
+				Validate(notEmpty),
```
obey-agent: **[bug]** On the identity step, `Name *` uses `Validate(notEmpty)` in the same huh group as the Next/Back confirm. Submitting with Back still validates the whole group, so a user who has not entered a name yet cannot use the Back control to return to the type step; they must ESC (which is treated as back to step 1) or invent a throwaway name. First-time navigation "type → identity → oops wrong type" is awkward.

**Suggestion:** Split name/goal input from navigation (separate forms), or only validate name when `cont == true` (custom validate that no-ops when backing), or use a select for Next/Back like the confirm step so validation can be conditional.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
+	if len(goal) > 80 {
+		goal = goal[:79] + "…"
+	}
+	return fmt.Sprintf(
+		"Name:     %s\nType:     %s\nDest:     festivals/%s/\nPhases:   %s\nProject:  %s\nSeed:     %s\nTags:     %s\nGoal:     %s",
+		d.Name, ft.Name, festivalTypeDest(ft), autoStr, proj, seed, tags, goal,
+	)
+}
+
+func submitFestivalCreate(ctx context.Context, cfg *types.FestivalTypesConfig, d *festivalDraft) error {
```
obey-agent, line 455: **[suggestion]** Charm `submitFestivalCreate` only writes `festival_name` / `festival_goal` / optional `festival_tags` into the vars file. The previous charm create/plan flows and the current **fallback** path still call `collectRequiredVars` + `defaultFestivalTemplatePaths` and prompt for any extra required template variables. Custom or upgraded festival templates with additional `required_variables` will get those prompts under `no_charm` but silent omission under charm.

**Suggestion:** Restore optional required-var collection in the charm submit path (or a dedicated details step), or intentionally drop it on both surfaces and document that festival create vars are limited to name/goal/tags.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
+		Type:     d.TypeName,
+		VarsFile: varsFile,
+		Project:  d.Project,
+		Seed:     d.Seed,
+		Dest:     dest,
+	}
+	return shared.RunCreateFestival(ctx, opts)
+}
+
+// listCampaignProjects returns relative projects/* paths under the campaign root.
```
obey-agent: **[suggestion]** `listCampaignProjects` only returns top-level `projects/*` directories and skips `worktrees`/dotdirs. Navigation’s `collectProjectDirectories` also expands monorepo submodules (`name@sub`). Campaigns that rely on monorepo project entries get an incomplete pick list and must use "Enter a custom path".

**Suggestion:** Reuse or share monorepo expansion logic with the go-link project picker so create and link surfaces list the same projects.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
-	vars := map[string]interface{}{"festival_name": name, "festival_goal": goal}
-	if strings.TrimSpace(tags) != "" {
-		vars["festival_tags"] = strings.Split(tags, ",")
+func stepFestivalProject(ctx context.Context, projects []string, d *festivalDraft) (bool, error) {
+	const (
+		pick = "pick"
+		path = "path"
+		skip = "skip"
+	)
+	choice := skip
```
obey-agent: **[suggestion]** When re-entering the project step with a non-empty `d.Project`, the choice always defaults to `"pick"` even if the user previously chose a custom path. If that path is not in `projects`, the pick list may not show their prior value as a real option (huh select will fall back to an available option), risking accidental overwrite of a carefully entered custom path on Next.

**Suggestion:** Track last project mode (`pick`/`path`/`skip`) on the draft, or if `d.Project` is set and not in `projects`, default choice to `path`.

internal/commands/tui/fallback_festival.go
```diff
@@ -78,106 +119,35 @@ func tuiCreateFestival(ctx context.Context, display *ui.UI) error {
...
 
-// Wizard: create festival then optionally add phases
-func tuiPlanFestivalWizard(ctx context.Context, display *ui.UI) error {
-	if err := ctx.Err(); err != nil {
-		return err
-	}
-	festivalsRoot, err := festivalsRootFromCtx(ctx)
-	if err != nil {
-		return err
+func festivalTypeSupportsSeedFallback(ft *types.FestivalType) bool {
```
obey-agent: **[suggestion]** Seed-support logic is duplicated as `festivalTypeSupportsSeed` (charm) and `festivalTypeSupportsSeedFallback` (no_charm). Tests only cover the charm function (`festival_wizard_test.go`). Drift already matches today but is easy to reintroduce (see Issue 1).

**Suggestion:** Move seed/dest helpers into a build-tag-free file in package `tui` (e.g. `festival_wizard_logic.go`) so both binaries and tests share one implementation.

internal/commands/hooks_create_festival_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+
+	"github.com/Obedience-Corp/fest/internal/commands/shared"
+)
+
+// TestCreateFestivalOptsCarryProjectAndSeed documents the shared opts surface
+// the TUI must populate so the hook can map into festival.CreateFestivalOptions.
+func TestCreateFestivalOptsCarryProjectAndSeed(t *testing.T) {
```
obey-agent: **[suggestion]** `TestCreateFestivalOptsCarryProjectAndSeed` only constructs a struct and asserts field assignment—it does not call `runCreateFestivalHook` or verify Project/Seed/SeedFile are forwarded into `festival.CreateFestivalOptions`. A future hook edit could drop mapping while this test still passes.

**Suggestion:** Unit-test the hook with a small fake/stub around `festival.RunCreateFestival`, or at least assert a table of shared→festival field copies if the package boundary allows.

internal/commands/tui/festival_wizard_test.go
```diff
@@ -0,0 +1,116 @@
...
+package tui
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/Obedience-Corp/fest/internal/types"
+)
+
+func TestFestivalTypeSupportsSeed(t *testing.T) {
```
obey-agent: **[suggestion]** Tests cover seed support, dest, option labels, and help note text, but not the step machine transitions (skip seed when type has no ingest; clear seed when switching to non-seed type), confirm summary contents, or project empty-list fallback. Highest-risk wizard logic is untested.

**Suggestion:** Add pure-function tests for `buildFestivalConfirmSummary` and table-driven cases for “after project, next step is 4 vs 5” / seed clearing by extracting transition helpers if needed.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
+	tags := d.Tags
+	if tags == "" {
+		tags = "(none)"
+	}
+	goal := d.Goal
+	if goal == "" {
+		goal = "(empty — agents plan better with a goal)"
+	}
+	if len(goal) > 80 {
+		goal = goal[:79] + "…"
```
obey-agent: **[nit]** Confirm summary truncates goal with `goal[:79] + "…"`, which slices by bytes and can split multi-byte UTF-8 code points for non-ASCII goals.

**Suggestion:** Truncate by runes (`[]rune` or `utf8` helpers) or use a shared display-ellipsis utility.

internal/commands/tui/charm_festival.go
```diff
@@ -5,195 +5,629 @@ package tui
...
+	)
+}
+
+func submitFestivalCreate(ctx context.Context, cfg *types.FestivalTypesConfig, d *festivalDraft) error {
+	ft := typeByName(cfg, d.TypeName)
+	dest := festivalTypeDest(ft)
+
+	vars := map[string]interface{}{"festival_name": d.Name, "festival_goal": d.Goal}
+	if strings.TrimSpace(d.Tags) != "" {
+		vars["festival_tags"] = strings.Split(d.Tags, ",")
```
obey-agent: **[nit]** `vars["festival_tags"] = strings.Split(d.Tags, ",")` does not trim whitespace around tags, while the create pipeline’s `parseTags(opts.Tags)` does. Templates reading custom vars may see `" commerce"` vs `"commerce"`.

**Suggestion:** Trim each segment (or reuse `parseTags` / equivalent) before writing the vars file.

internal/ui/theme/forms.go
```diff
@@ -30,6 +35,12 @@ func RunForm(ctx context.Context, form *huh.Form) error {
 		WithKeyMap(huh.NewDefaultKeyMap()).
 		WithShowHelp(true)
 
+	// Constrain form layout to the live terminal width. Option keys that are
+	// wider than the box wrap instead of painting past the right border.
+	if w, _, err := term.GetSize(int(os.Stdout.Fd())); err == nil && w > 4 {
```
obey-agent: **[nit]** Global `RunForm` width clamp (`term.GetSize(stdout) - 2`) is applied to every huh form, not only create menus. Correct fix for the reported border clipping; side effect is all TUI forms inherit terminal width even if a form previously relied on huh’s default sizing in odd environments (e.g. misreported TTY size under some multiplexers).

**Suggestion:** Keep the clamp (good default); if regressions appear, gate with a minimum width floor (e.g. only apply when `w < some threshold` or `w > 20`) and document behavior for CI/non-TTY (already no-ops when GetSize fails).

obey-agent (changes requested):
## Summary

Solid human-first TUI redesign: the multi-step charm festival wizard (type → identity → project → conditional seed → tags → confirm), Create what? menu with a workflow stub, shared `Project`/`Seed`/`SeedFile` plumbing through the create hook, and form-width clamping are well structured and align with the WI goals. Agent/CLI create paths look untouched aside from hook field mapping and friendlier human next-steps copy. A few correctness gaps remain—most importantly seed gating that is looser than the create pipeline, and an identity-step Back path blocked by name validation—plus charm/fallback parity and project-picker gaps.

## Issue counts by severity

- bugs: 2
- suggestions: 7
- nits: 3

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** internal/commands/tui/fallback_festival.go:17 -- Fallback festival create is still a linear prompt chain (type → name → goal → tags → project path yes/no → seed yes/no) with no confirm summary and no projects list picker, while charm is a full multi-step wizard with confirm. Acceptable as a degraded path, but behavior and defaults (e.g. freeform project only) diverge enough that bugs fixed on charm can be missed on fallback.
  - **Suggestion:** Either document fallback as intentionally minimal, or share pure helpers (draft, dest, seed support, confirm summary) in a non-build-tagged file so both paths stay aligned.

lancekrogers:
Addressed CHANGES_REQUESTED from the re-review:

**Bugs**
- Seed step now matches pipeline: only **auto** phases with `Type == \"ingest\"` (not name-only / non-auto ingest).
- Identity **Back** no longer blocked by name validation (input and Next/Back are separate forms).

**Suggestions**
- Shared pure helpers in `festival_wizard_logic.go` (charm + fallback).
- Extra `required_variables` from festival templates prompted on submit again.
- Project mode (`pick`/`path`/`skip`) remembered; monorepo `name@sub` paths expanded like go-link.
- Hook mapping tested via `festivalOptsFromShared`.
- Step-machine / summary / truncate / tags tests added.

**Nits**
- Goal truncate is rune-safe; tags trimmed; form width clamp floors at 40 cols.

Fallback remains intentionally leaner (no full confirm wizard) but uses the same seed/dest/vars helpers so it cannot drift on the critical gate.

internal/ui/theme/forms.go
```diff
@@ -16,32 +21,182 @@ const (
...
+	inner := m.form.View()
+	if m.width <= 0 {
+		return m.frame.Render(inner)
+	}
+	// painted width = Width + horizontal borders; target full terminal.
+	boxW := m.width - m.frameBorder
+	if boxW < 10 {
+		boxW = 10
+	}
+	return m.frame.Width(boxW).Render(inner)
```
obey-agent, line 87: **[bug]** Preserve the empty final view from Huh instead of framing it. Once a form submits, `m.form.View()` returns an empty string, but `frame.Render("")` still paints a two-line rounded box. Because this wizard launches a new Bubble Tea program for every step, those boxes accumulate in scrollback and consume the terminal: the committed wizard GIF visibly has 3 empty frames above Project, 5 above Seed, and 7 above Confirm. Return `""` when `inner == ""` before applying the frame, and add a sequential-form regression test or PTY assertion.

internal/ui/theme/forms.go
```diff
@@ -16,32 +21,182 @@ const (
...
+		m.form = m.form.WithWidth(inner)
+	}
+
+	opts := []tea.ProgramOption{
+		tea.WithOutput(os.Stderr),
+		tea.WithContext(ctx),
+		tea.WithReportFocus(),
+	}
+
+	final, err := tea.NewProgram(m, opts...).Run()
```
obey-agent, line 155: **[bug]** Keep the accessible runner from Huh for `TERM=dumb` and non-TTY environments. `huh.NewForm` marks itself accessible when `TERM=dumb`, and `Form.RunWithContext` routes that state through line-oriented `runAccessible`; constructing a Bubble Tea program directly bypasses it, along with configured form input, output, and options. Reproduction: with piped stdin and `TERM=dumb`, `main` reaches the numbered text prompts, while this head exits immediately with `RunForm: form error: could not open a new TTY`. Preserve the accessible path before wrapping normal TTY forms, and cover it with a subprocess test.

internal/commands/tui/festival_wizard_logic.go
```diff
@@ -0,0 +1,286 @@
...
+	return vars
+}
+
+// resolveProjectMode picks the project step default from the draft.
+func resolveProjectMode(d *festivalDraft, projects []string) string {
+	if d.ProjectMode == projectModePick || d.ProjectMode == projectModePath || d.ProjectMode == projectModeSkip {
+		return d.ProjectMode
+	}
+	if d.Project == "" {
+		return projectModeSkip
```
obey-agent, line 214: **[suggestion]** A fresh draft always defaults the Project step to `Skip for now`, even when `projects` is non-empty. That contradicts the screen copy, `Recommended for implementation work`, and the VHS tape instruction, `Project: pick from list`; the committed demo actually reaches Confirm with `Project: (none)`. Default to `pick` when choices exist and `skip` only when none exist, or make the skip-first behavior explicit and correct the demo and test.

obey-agent (changes requested):
Re-review of the follow-up commits: the prior seed gate, Back navigation, shared fallback logic, required-variable prompts, project-mode persistence and monorepo expansion, hook mapping, and test gaps are addressed. Requesting changes for two regressions introduced by the global form-frame runner: sequential wizard forms leave accumulating empty frames, and the TERM=dumb accessible mode from Huh no longer works. `go test ./...` and `go vet ./...` pass; the existing `no_charm` build failures reproduce unchanged on the base and are not counted against this PR.

lancekrogers:
Both regressions from the re-review are fixed and verified against the real binary, not just the suite.

**1. Accumulating empty frames** (`forms.go`) — `formFrameModel.View` now returns `""` when huh's view is empty.

Verified by driving `fest create festival` in a pty (`pyte`, 100x60) and reading the rendered screen. Before, by the Seed step:

```
  0|╭────────────────────────────────────────────────╮
  1|│                                                │
  2|╭────────────────────────────────────────────────╮
  3|│                                                │
  4|╭────────────────────────────────────────────────╮
  5|│                                                │
  6|╭────────────────────────────────────────────────╮
  7|│                                                │
  8|╭────────────────────────────────────────────────╮
  9|│ New Festival · Seed                            │
```

After: one box per step, nothing left in scrollback.

The regression test that shipped with the fix asserted `RenderFormFrame`, which only tests reach — it still passed with the guard removed. Replaced with `TestFormFrameModelRendersNothingWhenFinished`, which drives `formFrameModel` (the path `RunForm` uses) to both submitted and aborted. With the guard removed it fails with the exact stale box:

```
--- FAIL: TestFormFrameModelRendersNothingWhenFinished/submitted
    finished form left a frame in scrollback:
    "╭───...───╮\n│   ...   │\n╰───...───╯"
```

**2. `TERM=dumb` accessible mode** (`forms.go`) — accessible forms are handed back to `Form.RunWithContext` before the frame model wraps them. `TERM=dumb` is huh's own trigger in `NewForm`, and `Form.accessible` has no getter, so that is the detectable signal.

Reproduced and confirmed on the real CLI. Before:

```
$ printf '\n\n' | TERM=dumb fest create
Error: RunForm: form error: could not open a new TTY: open /dev/tty: device not configured
```

After:

```
$ printf '\n\n' | TERM=dumb fest create
Create what?
1. Festival
2. Standalone workflow
3. Phase
4. Sequence
5. Task
6. Back
Enter a number between 1 and 6:
New Festival · Type
...
```

Covered by the subprocess test `TestRunFormAccessibleTERM`.

**3. Project step default** (`festival_wizard_logic.go`) — defaults to `pick` when the campaign has projects, `skip` only when there are none. Test covers both.

**Demos re-recorded.** Both committed GIFs were recorded before the fixes; the menu GIF still had a stale frame above the workflow stub, and the wizard reached Confirm with `Project: (none)`. The wizard tape also needed a new key sequence, since the `pick` default inserts a project-list form the old sequence walked past — that off-by-one was what left `Project: (none)` even after a fresh recording. New sequence derived by walking the wizard in a pty with wait-for-screen instead of fixed sleeps. Confirm now reads:

```
Name:     ecommerce-mvp
Type:     standard
Dest:     festivals/planning/
Phases:   001-INGEST, 002-PLAN (auto)
Project:  projects/demo-app
Seed:     yes (43 chars)
Tags:     backend,commerce
Goal:     Launch MVP with catalog, cart, and checkout
```

The picker only lists inside a campaign, so the fixture needs `.campaign/` and `projects/*`; `docs/demos/README.md` and the tape header now say so.

**Two things found while verifying, also fixed:**

- The autowrap escapes added with the frame runner were written unconditionally, so redirected stderr collected them: `1b5b 3f37 6c 1b5b 3f37 68` ahead of the error text. Now gated on stderr being a TTY.
- `normalizeFormRunError` took a `ctx` it never used (`unusedparams`). Dropped.

**Gates:** `go build ./...`, `go vet ./...`, `go test ./...` all pass. `golangci-lint run` on the touched packages is clean. Repo-wide lint reports 3 pre-existing `errcheck` issues in `internal/commands/pack/pack.go` from #304, identical on `main`. The `no_charm` build failures are also byte-identical on `main`.

PR #309, opened 2026-07-30
feat(commit): route staging and commits through camp commitkit (CA0004 003/02)

lancekrogers:
## What

Closes fest's raw-git staging and commit holes through the camp v0.4.0-rc.1 bump (CA0004 phase 003, sequence 02):

- **Bump**: `github.com/Obedience-Corp/camp` v0.3.1 -> v0.4.0-rc.1. Version line only, no `replace`.
- **StageAll swap**: `stageAllChanges` no longer runs a raw `git add -A` in whatever cwd fest held. It stages through `commitkit.StageAllWithOutcome` against an explicit repo path, inheriting lock retry, the large-file guard, and bulk protection. Both auto-stage branches (in-submodule and non-campaign fallback) are covered.
- **Fest-rendered guard output**: violations arrive as typed data (`GuardBlockedError`, `StageOutcome`) and fest renders them in its own voice; no camp error-string parsing. A refusal reaches `result.Error` naming the same ways out camp names (gitignore it, `camp artifacts add`, commit it anyway, settings switch), with `fest commit --commit-large` as the retry vehicle. Successful stages report exclusions on stderr so `--json` stdout stays pure.
- **Commit swap**: `executeGitCommit` deleted. The primary commit goes through `commitkit.Commit` + `commitkit.ShortHash`, the same pair `commitCampaignRoot` already used. `ErrNoChanges` maps to a clear "no changes to commit" instead of a generic git failure.

## The `--json` fix (criterion 15a)

`executeGitCommit` wired git's stdout to the process stdout while `--json` encoded to the same stream. Captured failing baseline, pre-swap, from the regression test:

```
[main (root-commit) 381c2e2] json purity
 2 files changed, 3 insertions(+)
 create mode 100644 festivals/active/guard-festival/fest.yaml
 create mode 100644 notes.md
{
  "success": true,
  "hash": "381c2e2",
  ...
```

`commitkit.Commit` captures git's output internally; the test now asserts stdout is exactly one JSON document with nothing before or after it. This resolves intent `fest-commit-json-interleaves-git-20260726-001838`.

## Proofs (containerized integration tests)

- **15z** (`commit_guard_test.go`): an untracked sparse 512 MB generated file in a festival workspace is kept out of git and left on disk, said out loud with its undo, in both auto-stage branches (standalone workspace and linked submodule). Project-side policy applies in each, matching `camp commit`.
- **15a** (`commit_json_test.go`): `fest commit --json` emits exactly one JSON document on stdout; the reported hash names a commit that exists at return.
- **15 and 15b** (`commit_drain_test.go`): `camp intent add` (run with a camp binary built from this module's own pinned dependency version) queues its bookkeeping commit; an immediately following `fest commit` lands that queued commit in history before the festival commit, returns with the queue empty, and reports a hash that is HEAD at return. Fest never defers.

## Not in this PR

`--commit-large` passthrough. The released commitkit has no options-taking stage call; camp PR Obedience-Corp/camp#523 exports `StageAllWithOptions`, and once it merges and ships as v0.4.0-rc.2 this branch gains the flag, threaded to the same override camp's own `--commit-large` uses.

## Found while fixturing (intents filed, out of scope here)

- `fest commit` fails in a fresh campaign: `festivalScopedPaths` lists `.campaign/fest` and `festivals/.festival/.state` unconditionally; git errors on the missing pathspec before fest has ever navigated.
- camp: deferred bookkeeping commits fail silently on an unborn branch (`camp init` leaves no initial commit; `read-tree` cannot resolve HEAD).

## Testing

Unit suite green; new guard rendering unit tests; the four containerized integration tests above; full `just gate` run on this head (result in the PR checks conversation).

### DIFF
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -4,6 +4,7 @@ package commit
 import (
 	"bytes"
 	"context"
+	stderrors "errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -174,7 +175,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	if autoStage {
 		if inSubmodule {
 			// In submodule: stage all changes in the project repo
-			if err := stageAllChanges(ctx); err != nil {
+			if err := stageAllChanges(ctx, primaryRepoPath); err != nil {
 				result.Success = false
 				result.Error = err.Error()
 				return outputResult(result)
@@ -194,7 +195,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 			}
 		} else {
 			// Fallback: stage all (non-campaign workspace)
-			if err := stageAllChanges(ctx); err != nil {
+			if err := stageAllChanges(ctx, primaryRepoPath); err != nil {
 				result.Success = false
 				result.Error = err.Error()
 				return outputResult(result)
@@ -354,15 +355,24 @@ func currentGitRoot(ctx context.Context) (string, error) {
 	return strings.TrimSpace(string(out)), nil
 }
 
-// stageAllChanges runs git add -A to stage all changes
-func stageAllChanges(ctx context.Context) error {
+// stageAllChanges stages everything in the repository at repoPath through
+// camp's staging guard: lock retry with stale-lock cleanup, size and bulk
+// protection, and a typed refusal fest renders in its own voice. The repo is
+// passed explicitly because the guard resolves its thresholds from it; the
+// old form ran `git add -A` in whatever cwd fest happened to hold.
+func stageAllChanges(ctx context.Context, repoPath string) error {
 	if err := ctx.Err(); err != nil {
 		return errors.Wrap(err, "context cancelled")
 	}
-	cmd := exec.CommandContext(ctx, "git", "add", "-A")
-	if err := cmd.Run(); err != nil {
-		return errors.Wrap(err, "git add failed")
+	outcome, err := commitkit.StageAllWithOutcome(ctx, repoPath)
+	if err != nil {
+		var blocked *commitkit.GuardBlockedError
+		if stderrors.As(err, &blocked) {
+			return errors.New(guardRefusalMessage(blocked))
+		}
+		return errors.Wrap(err, "staging changes")
 	}
+	reportStageOutcome(os.Stderr, outcome)
 	return nil
 }
 
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	stderrors "errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -20,6 +21,11 @@ import (
 	"github.com/Obedience-Corp/fest/internal/ui"
 	"github.com/spf13/cobra"
 )
+
+// errAllExcluded is returned when the staging guard kept every changed path
+// out of the index, so a follow-up commit would look like a bare "no changes"
+// failure after the user already saw exclusion lines on stderr.
+var errAllExcluded = stderrors.New("everything changed was kept out of git by the size/bulk guard; nothing left to commit")
 
 // Commit reference format prefix: FE (Festival component)
 const commitRefPrefix = "FE"
@@ -169,32 +175,37 @@ func runCommit(cmd *cobra.Command, args []string) error {
 		return outputResult(result)
 	}
 
+	// Guard outcomes and refusals go to the command's stderr so --json stdout
+	// stays a pure document.
+	report := cmd.ErrOrStderr()
+
 	// Auto-stage changes if enabled (default: true).
 	// When at campaign root (not in submodule), stage only festival-scoped paths.
 	if autoStage {
 		if inSubmodule {
 			// In submodule: stage all changes in the project repo
-			if err := stageAllChanges(ctx, primaryRepoPath); err != nil {
+			if err := stageAllChanges(ctx, primaryRepoPath, report); err != nil {
 				result.Success = false
 				result.Error = err.Error()
 				return outputResult(result)
 			}
 		} else if ws != nil && ws.Type == scope.WorkspaceTypeCampaign && hasFestival {
-			// At campaign root: stage only festival-scoped paths
+			// At campaign root: stage only festival-scoped paths (with outcome
+			// reporting so large/bulk exclusions are not silent).
 			paths, err := festivalScopedPaths(ctx, ws.Root, festivalPath, "")
 			if err != nil {
 				result.Success = false
 				result.Error = err.Error()
 				return outputResult(result)
 			}
-			if stageErr := commitkit.StageFiles(ctx, ws.Root, paths...); stageErr != nil {
+			if stageErr := stageFiles(ctx, ws.Root, report, paths...); stageErr != nil {
 				result.Success = false
-				result.Error = fmt.Sprintf("staging festival files: %v", stageErr)
+				result.Error = stageErr.Error()
 				return outputResult(result)
 			}
 		} else {
 			// Fallback: stage all (non-campaign workspace)
-			if err := stageAllChanges(ctx, primaryRepoPath); err != nil {
+			if err := stageAllChanges(ctx, primaryRepoPath, report); err != nil {
 				result.Success = false
 				result.Error = err.Error()
 				return outputResult(result)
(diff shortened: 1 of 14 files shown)


internal/commands/commit/guard.go
```diff
@@ -0,0 +1,97 @@
...
+			total += v.TotalBytes
+			if v.Count > worst.Count {
+				worst = v
+			}
+		}
+		fmt.Fprintf(&b, "%d untracked files (%s) under %s/ would be committed; nothing was staged\n",
+			files, formatBytes(total), worst.CommonPrefix)
+		fmt.Fprintf(&b, "  gitignore it        echo '%s/' >> .gitignore\n", worst.CommonPrefix)
+		fmt.Fprintf(&b, "  keep and sync it    camp artifacts add %s\n", worst.CommonPrefix)
+		fmt.Fprintf(&b, "  commit it anyway    fest commit --commit-large -m \"...\"\n")
```
obey-agent: **[bug]** Bulk and large-file refusal text (and successful exclusion reports at line 70) tell the user to run `fest commit --commit-large`, and unit/integration tests lock that string in. This PR does not register a `--commit-large` flag or call `StageAllWithOptions` (not present on camp v0.4.0-rc.1; camp PR #523 is the export). Following the printed remedy yields an unknown-flag failure. The PR body acknowledges the deferral, but shipping user-facing guidance that cannot work is still a product bug for anyone blocked or excluded by the guard.

**Suggestion:** Until the flag exists, drop `fest commit --commit-large` from refusal/exclusion copy (and the tests that pin it) and only list remedies that work today (`gitignore`, `camp artifacts add`, `camp settings set …`, or `camp commit --commit-large` where appropriate). When camp ships `StageAllWithOptions`, add the flag and switch `stageAllChanges` to that API in the same change as restoring the message.

internal/commands/commit/commit.go
```diff
@@ -354,15 +331,24 @@ func currentGitRoot(ctx context.Context) (string, error) {
...
+// passed explicitly because the guard resolves its thresholds from it; the
+// old form ran `git add -A` in whatever cwd fest happened to hold.
+func stageAllChanges(ctx context.Context, repoPath string) error {
 	if err := ctx.Err(); err != nil {
 		return errors.Wrap(err, "context cancelled")
 	}
-	cmd := exec.CommandContext(ctx, "git", "add", "-A")
-	if err := cmd.Run(); err != nil {
-		return errors.Wrap(err, "git add failed")
+	outcome, err := commitkit.StageAllWithOutcome(ctx, repoPath)
```
obey-agent: **[bug]** Only `stageAllChanges` uses `StageAllWithOutcome` plus `guardRefusalMessage` / `reportStageOutcome`. The festival-at-campaign-root auto-stage branch still calls `commitkit.StageFiles` (no `WithOutcome`, no `errors.As` on `*GuardBlockedError`), and `commitCampaignRoot` does the same. In camp v0.4.0-rc.1, `StageFiles` runs the guard but discards the outcome—exactly the silent-exclusion hazard commitkit’s docs warn about. A large untracked file under a festival directory can be kept out of the campaign-root commit with no fest stderr line and no fest-rendered refusal on block; bulk refusals surface only as a generic wrapped error string.

**Suggestion:** Route those call sites through `StageFilesWithOutcome` (or a small helper shared with `stageAllChanges`) so GuardBlockedError is re-rendered and StageOutcome is reported on stderr. Optionally mirror camp’s “nothing left to commit after exclusions” handling so exclude-everything does not read as a bare failure.

internal/commands/commit/commit.go
```diff
@@ -471,9 +457,18 @@ func commitWithCampaignSupport(ctx context.Context, ws *scope.WorkspaceInfo, rep
 		}
 	}
 
-	hash, err := executeGitCommit(ctx, repoPath, commitMessage)
+	// commitkit.Commit captures git's output internally, which is the --json
+	// contract: nothing but the encoded document may reach stdout. The same
+	// pair already commits the campaign root in commitCampaignRoot.
+	if err := commitkit.Commit(ctx, repoPath, commitkit.CommitOptions{Message: commitMessage}); err != nil {
+		if stderrors.Is(err, commitkit.ErrNoChanges) {
+			return errors.New("no changes to commit")
```
obey-agent, line 523: **[suggestion]** After a successful guarded stage that excluded every changed path, `commitkit.Commit` returns `ErrNoChanges` and fest maps that to `"no changes to commit"`. Camp pairs the same situation with an explicit “everything changed was kept out of git” path so the user does not think staging/commit failed after they already saw exclusion lines on stderr.

**Suggestion:** After staging, if there is a non-empty outcome with exclusions and `HasStagedChanges` is false, return a clearer no-op message (and success-shaped JSON if that matches camp’s commit JSON contract) instead of a generic no-changes error.

obey-agent (commented):
## Summary

Solid, well-scoped migration: `stageAllChanges` and the primary commit path correctly move onto camp v0.4.0-rc.1 commitkit (`StageAllWithOutcome`, `Commit`, `ShortHash`), fest renders typed guard data without parsing camp error strings, and the new integration tests cover the two auto-stage branches, JSON stdout purity, and deferred-commit drain ordering. The largest ship-time defect is that refusal/exclusion copy advertises `fest commit --commit-large` while that flag is explicitly not implemented (and `StageAllWithOptions` is not yet in the pinned rc). Festival-at-campaign-root staging still goes through bare `StageFiles`, so the common in-festival path does not get fest outcome/refusal rendering.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 1

tests/integration/commit_drain_test.go
```diff
@@ -0,0 +1,174 @@
...
+		campHostPath = filepath.Join(gopath, "bin", "linux_"+runtime.GOARCH, "camp")
+		if runtime.GOOS == "linux" {
+			campHostPath = filepath.Join(gopath, "bin", "camp")
+		}
+	})
+	if campBuildErr != nil {
+		t.Fatalf("building camp from the pinned dependency: %v", campBuildErr)
+	}
+	if _, err := tc.Exec("test", "-x", "/usr/local/bin/camp"); err == nil {
+		return
```
obey-agent: **[suggestion]** `ensureCampInContainer` always builds the pinned camp version on the host, but if `/usr/local/bin/camp` is already executable in the shared container it returns without re-copying. A preinstalled or previously copied binary of a different camp version would drive the drain fixture while fest links against v0.4.0-rc.1, weakening the “exact released version” claim in the test comment.

**Suggestion:** Always copy the freshly built `campHostPath` into the container (or compare versions / reinstall when the host build path is newer) so the fixture cannot silently use a stale camp.

internal/commands/commit/commit.go
```diff
@@ -354,15 +331,24 @@ func currentGitRoot(ctx context.Context) (string, error) {
...
-		return errors.Wrap(err, "git add failed")
+	outcome, err := commitkit.StageAllWithOutcome(ctx, repoPath)
+	if err != nil {
+		var blocked *commitkit.GuardBlockedError
+		if stderrors.As(err, &blocked) {
+			return errors.New(guardRefusalMessage(blocked))
+		}
+		return errors.Wrap(err, "staging changes")
 	}
+	reportStageOutcome(os.Stderr, outcome)
```
obey-agent: **[nit]** `reportStageOutcome` is always written to `os.Stderr` rather than `cmd.ErrOrStderr()`. That matches shell-redirection integration tests and keeps `--json` stdout pure, but it is less flexible than the rest of the CLI’s writer plumbing (e.g. pack’s recent `_, _ = fmt.Fprintf(cmd.ErrOrStderr(), …)` style) and harder to unit-test without capturing the process stderr.

**Suggestion:** Thread an `io.Writer` (or `cmd.ErrOrStderr()`) into `stageAllChanges` / `reportStageOutcome` for consistency; keep the default as stderr.

obey-agent:
All five review findings are addressed as of 4c82f0a. Two sessions serviced this review in parallel; 781dd71 was adopted as the base and the commits on top of it complete the set.

**Bug 1 (advertised flag did not exist)** — resolved in two stages, ending better than either alternative: 781dd71 stripped `fest commit --commit-large` and pointed at camp's flag; then camp merged the `StageAllWithOptions` export (camp #523) and shipped it as **v0.4.0-rc.2**, so 6cf03ec bumps the dependency and registers the real `fest commit --commit-large`, threaded to the same guard override camp's own flag uses. The commit-it-anyway line is now supplied **per call site** so it is truthful on every branch: stage-all branches name fest's flag (real and honored), file-list branches keep `camp commit --commit-large`, because commitkit exports no options form for file-list staging and `camp commit` refuses outside a campaign — the standalone branches must not name it. The advertised retry is proven functionally: the standalone 15z test reruns the refused commit with `--commit-large` and asserts the 512 MB file lands in git.

**Bug 2 (campaign-root staging bypassed rendering)** — 781dd71 routes the festival-scoped branch and `commitCampaignRoot` through `StageFilesWithOutcome` with the same `GuardBlockedError` rendering and stderr outcome reporting as the auto-stage branches.

**Suggestion 1 (exclude-everything reads as failure)** — 781dd71 adds `nothingLeftAfterExclusions` (HasStagedChanges after exclusions, `errAllExcluded` with the guard named). Worth noting: the raw failure it replaces exists because `commitkit.Commit`'s documented ErrNoChanges mapping misses the untracked-files-only case; that camp gap is filed as an intent. New containerized coverage grows only the already-excluded file, commits again, and asserts the no-op is attributed to the guard.

**Suggestion 2 (stale camp binary in the drain fixture)** — the fixture now always installs the freshly built binary; a leftover container binary can never impersonate the pinned release.

**Nit (os.Stderr vs ErrOrStderr)** — `cmd.ErrOrStderr()` is threaded as an `io.Writer` through every staging and reporting path; an unchecked Fprintf that arrived with the threading is also fixed.

Verification on 4c82f0a: full `just gate` green (0 lint issues, 2743 unit + 75 integration tests), including the four commit-path integration families (guard 15z both branches plus exclude-everything plus functional `--commit-large`, JSON purity 15a, drain ordering 15/15b, now against camp v0.4.0-rc.2). CLI reference regenerated in a separate commit (also sweeping never-regenerated `fest pack` pages).

PR #310, opened 2026-07-30
feat(resident): recognize camp lifecycle residents, read-only (WF0001 P4/S1)

lancekrogers:
WF0001 phase 004, sequence 01. Teaches fest to recognize a camp-defined lifecycle resident, without ever writing camp's marker.

**Base:** `main`. First fest-side PR of this festival; phases 003 (camp) shipped in camp #521/#524/#525.

## Read-only marker reader

New `internal/resident` with two functions and no writer:

- `Read(dir)` — `(nil, nil)` when `.workitem` is absent **or** carries another `kind`; error only when the file exists and cannot be read or parsed.
- `IsResident(dir)` — predicate; reports false on an unreadable marker so a listing path cannot be broken by one bad file. Callers needing the error call `Read`.

D007 is in the package doc: camp owns `.workitem`, fest reads it. This mirrors camp reading fest's `.workflow/` runtime through a thin duplicate of the fields it needs.

**The version is deliberately not validated.** Camp owns the schema, so a fest-side allowlist would make fest fail on a camp upgrade it has no say in. A test feeds a `v9alpha1` marker with unknown scalar and nested fields to pin that.

**A wrong `kind` returns `(nil, nil)`, not an error.** A stray sidecar is not a resident, and it is not fest's job to complain — `camp workitem doctor` already owns that judgment (camp #525 added exactly those checks).

## `isValidFestival` consolidated rather than copied again

The task said to reuse fest's existing festival-marker check "rather than inventing a third". There were already **three** byte-equivalent copies:

- `internal/scope/resolve.go:207`
- `internal/commands/show/detect.go:170`
- `internal/commands/shared/festival_path.go:80`

Importing any of them from a leaf package would invert layering, and adding a fourth would make it worse. The canonical implementation now lives in `internal/resident` as `IsFestivalDir`, and all three delegate to it.

`Classify` gives the three-way answer (`neither`/`festival`/`resident`) sequence 02 needs. The workitem marker wins over festival markers, for the same reason camp's discovery prefers it: camp writes `.workitem` only on a real promote, while a leftover `fest.yaml` proves nothing about the current owner.

## `fest validate` reports residents instead of erroring

Before, `fest validate` told the user their camp workitem was a malformed festival. Now:

```
old: Error: path is not a festival directory (expected FESTIVAL_OVERVIEW.md ...)   exit=1
new: lifecycle resident (design) owned by camp; not a festival, nothing to validate   exit=0
```

`--json` reuses the normal `ValidationResult` envelope, so consumers see the same shape with `ok: true`, `valid: true`, and one `level: info` issue coded `lifecycle-resident`.

The intercept sits after `resolveFestivalPath` fails, so that function and `isFestivalDir` are untouched and only the marker-present branch changes. A directory with festival intent but no marker still errors exactly as before.

## No-regression, proved by binary diff

fest has real production users, so this is a diff rather than an assertion. Built fest from `main` and from this branch, ran both against one campaign fixture holding a real festival, a resident, and a plain directory:

```
=========== festivals/active/realfest: old vs new ===========
IDENTICAL to pre-change binary
=========== festivals/active/plain: old vs new ===========
IDENTICAL to pre-change binary
```

Goldens are committed under the festival's `results/golden/`.

## Constraints

```
$ go list -deps ./internal/resident/... | grep -c "Obedience-Corp/camp"
0
$ grep -rn "WriteFile.*\.workitem|Create.*\.workitem" internal/ --include="*.go" | grep -v _test
(empty)
```

No camp dependency in the new package and no `.workitem` write path anywhere in fest. (Fest does import `camp/pkg/ledgerkit` in three pre-existing places — a published contract package, not camp internals, predating this work.)

## Gate

`just test unit`: 2717 passed. `golangci-lint --new-from-merge-base origin/main`: 0 issues.


obey-agent (commented):
## Verdict: Comment (merge-ready) — one finding found and fixed during review

**Merge-ready with one finding fixed.** Posted as a Comment; the Approve verdict is Lance's.

## Finding (fixed): two code paths in this sequence disagreed on precedence

`resident.Classify` lets the workitem marker win outright over festival markers.
`fest validate` consults the marker only *after* `resolveFestivalPath` has already
failed. So a directory carrying both `.workitem` and `FESTIVAL_OVERVIEW.md`
classifies as a resident but still validates as a festival.

Both behaviors are defensible, but the divergence was accidental — I wrote both in
the same sequence without noticing. It is now deliberate, commented at the
intercept, and pinned by `TestValidate_BothMarkersStillValidatesAsFestival`:

- `Classify` drives **labelling** (cards, completion in sequence 02), where
  resident-wins matches camp's own discovery.
- The validate intercept drives **whether to run checks**, and the conservative
  choice is not to take validation away from a directory with real festival
  structure.

## Requirements checked

**Read-only.** No writer exists in `internal/resident`, and
`grep "WriteFile.*\.workitem"` across `internal/` is empty.
`TestPackageIsReadOnly` byte-compares the marker before and after both calls,
so a future writer would have to break a test to land.

**No camp import.** `go list -deps ./internal/resident/...` reports 0 camp
packages. The three pre-existing `camp/pkg/ledgerkit` imports elsewhere in fest are
a published contract package, not camp internals, and predate this work.

**Tolerant parse.** `TestRead/unknown_fields_are_ignored` feeds a `v9alpha1`
marker with unknown scalar and nested fields. No version allowlist exists, which is
correct: camp owns the schema, so validating it here would make fest fail on a camp
upgrade it has no say in.

**Not-found contract mirrored.** `Read` returns `(nil, nil)` for absent, matching
camp's `LoadMetadata`. It also returns `(nil, nil)` for a wrong `kind`, which is a
judgment call worth stating: a stray sidecar is not a resident, and
`camp workitem doctor` already owns complaining about it.

**Zero-resident no-regression.** Demonstrated by binary diff rather than
assertion, which matters because fest has production users. Goldens in
`results/golden/` from a `main` build and a branch build: a real festival and a
plain directory are byte-identical on both.

**Consolidation, not a fourth copy.** The task asked to reuse the existing
festival-marker check. There were three byte-equivalent copies; all three now
delegate to `resident.IsFestivalDir`. A leaf package owning the check is also the
right dependency direction — importing `internal/commands/show` from
`internal/resident` would have inverted it.

## Standards

fest's `errors` package throughout (`errors.IO`, `errors.Wrap`, `.WithField`); no
`fmt.Errorf`. Every added function is small and single-purpose.
`residentResult` was split out of `emitResidentInfo` so the `--json` envelope can be
asserted without capturing stdout.

One pre-existing pattern noted but not changed: `ui.UI.Info` writes via
`fmt.Println` to process stdout rather than `cmd.OutOrStdout()`. That is how the
entire `internal/ui` package works, so matching it is the right call here; changing
it is a separate concern.

## Gate

```
$ just test unit
✓ ALL 2718 TESTS PASSED
$ golangci-lint run --new-from-merge-base origin/main ./...
0 issues.
```

PR #311, opened 2026-07-30
feat(list,go): surface lifecycle residents in status, list, and go (WF0001 P4/S2)

lancekrogers:
WF0001 phase 004, sequence 02. Residents become visible in `fest list`, `fest status list`, and `fest go`. Base `main` (fest #310 merged).

## Cards in list and status

`internal/commands/show/residents.go` — `ResidentCard`, `ListResidentsByStatus`, `Progress()`. Only `ready` and `active` yield residents; every other status returns nil, so callers need no special-casing.

`FormatResidentList` returns `""` for zero residents, which *is* the no-regression guarantee: no header, no separator, no count.

Run progress comes from `ResolveStandaloneWorkflow`, fest's own runtime reader. No camp dependency.

```
ACTIVE Residents (1)
────────────────────────────────────────
  plainres  [design]

READY Residents (1)
────────────────────────────────────────
  runres  [explore]  1/3 steps
```

The `1/3 steps` was produced by creating a real standalone workflow in the resident and advancing a step, not by hand-writing runtime state.

## `fest go`

Both navigation collectors gained a resident fallback for the working stages only — a resident in `festivals/.dungeon/` is terminal and out of v1 scope. `fest go` resolves through the same `CollectNavigationTargets` that feeds completion, so extending the collector fixed completion *and* `fest go <resident>` in one place.

Festival targets keep their priority: a festival and a resident on the same stage get the same value, so residents are additional targets rather than a reordering.

## A JSON contract leak, caught by running the command

`fest status list --json` marshalled `[]*ResidentCard` directly and emitted Go field names, the absolute path, and the entire standalone runtime:

```
"residents":[{"Name":"runres","Title":"Run Resident","Type":"explore",
              "Path":"/private/tmp/.../runres","Run":{"Mode":"standalone-tracked",...
```

while `fest list --json` used a hand-written mapper with clean keys. Two surfaces, two shapes, one leaking paths.

Fixed at the type: `ResidentCard.MarshalJSON` publishes an explicit shape and the duplicate mapper is gone, so both surfaces agree and a new field cannot leak by default. `TestResidentCard_MarshalJSONShape` asserts the exact key set and that `Path`, `Run`, and `RuntimeDir` are absent.

## Zero-resident: byte-identical, by diff

fest is in production, so this is a real diff of two binaries against the same festival-only fixture, not an assertion:

```
  fest list                            : IDENTICAL
  fest list all                        : IDENTICAL
  fest list active                     : IDENTICAL
  fest list --all                      : IDENTICAL
  fest status active                   : IDENTICAL
  fest status ready                    : IDENTICAL
  fest status list --status active     : IDENTICAL
  fest go completions                  : IDENTICAL
```

Goldens committed under the festival's `results/golden/`.

## What the "TUI board" actually is

The task asked to extend and pty-verify a TUI stage-column board. Driving both binaries in a pty established there isn't one, and corrected my own earlier note in the process:

- **`fest list --watch`** is the live-repainting surface this change affects. Verified with `pyte` + `pty.fork()`; residents render in the right stage sections with badges and `1/3 steps`, nothing clipped. Snapshot committed.
- **The bare `fest go` picker** is a real bubbletea TUI, but it **already** listed residents before this change — it uses `FestivalPickerItems` with `IncludeUnmarkedFestivalDirectories: true`. Confirmed by driving the pre-change binary in the same pty and getting a byte-identical screen. Unaffected, verified rather than assumed.
- **`charm_list.go`** is the `fest go list` shortcut/link selector, not a stage board.
- **`list_board.go`** is the text board builder shared by one-shot and `--watch`, so extending it covered both.

## Gate

`just test unit`: 2764 passed. `golangci-lint --new-from-merge-base origin/main`: 0 issues.

### DIFF
--- a/internal/commands/list/list.go
+++ b/internal/commands/list/list.go
@@ -405,6 +405,13 @@ func listDungeon(ctx context.Context, festivalsDir string, opts *listOptions, ca
 			result[status] = festivalsToMapWithProgress(festivals, board.Progress)
 		}
 		result["total"] = board.Total
+		if len(board.Residents) > 0 {
+			residents := make(map[string]interface{}, len(board.Residents))
+			for status, cards := range board.Residents {
+				residents[status] = cards
+			}
+			result["residents"] = residents
+		}
 		return outputJSON(result)
 	}
 	fmt.Print(formatDungeonHuman(board.Festivals, board.Order, board.Progress, board.Total, opts.progress))
@@ -417,13 +424,17 @@ func listByStatus(ctx context.Context, festivalsDir, status string, opts *listOp
 		return err
 	}
 	if opts.json {
-		return outputJSON(map[string]interface{}{
+		payload := map[string]interface{}{
 			"status":    status,
 			"count":     len(board.Festivals),
 			"festivals": festivalsToMapWithProgress(board.Festivals, board.Progress),
-		})
+		}
+		if len(board.Residents) > 0 {
+			payload["residents"] = board.Residents
+		}
+		return outputJSON(payload)
 	}
-	fmt.Print(formatStatusHuman(status, board.Festivals, board.Progress, opts.progress))
+	fmt.Print(formatStatusHuman(status, board.Festivals, board.Residents, board.Progress, opts.progress))
 	return nil
 }
 
@@ -438,9 +449,16 @@ func listAll(ctx context.Context, festivalsDir string, opts *listOptions, campai
 			result[status] = festivalsToMapWithProgress(festivals, board.Progress)
 		}
 		result["total"] = board.Total
+		if len(board.Residents) > 0 {
+			residents := make(map[string]interface{}, len(board.Residents))
+			for status, cards := range board.Residents {
+				residents[status] = cards
+			}
+			result["residents"] = residents
+		}
 		return outputJSON(result)
 	}
-	fmt.Print(formatAllHuman(board.Festivals, board.Order, board.Progress, board.Total, opts.progress))
+	fmt.Print(formatAllHuman(board.Festivals, board.Residents, board.Order, board.Progress, board.Total, opts.progress))
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready) — two findings found and fixed during review

**Merge-ready with two findings fixed.** Posted as a Comment; the Approve verdict is Lance's.

## Findings (both fixed)

### 1. JSON contract leak across two surfaces

`fest status list --json` marshalled `[]*ResidentCard` directly, publishing Go
field names, the absolute filesystem path, and the entire
`StandaloneWorkflowInfo`. `fest list --json` used a hand-written mapper with clean
snake_case keys. Two surfaces, two shapes, one leaking paths.

Caught by running the command, not by reading the diff. Fixed at the type with
`MarshalJSON` and an explicit published struct, so both surfaces marshal the same
card and a future field cannot leak by default. The duplicate mapper is gone.

### 2. `collectAllBoard` crossed the 50-line rule

Adding the resident gather pushed it to 51 lines, over the festival's function-size
rule. Extracted `collectStatusFestivals` (list, filter, sort for one status); the
function is now 50 and the extraction removes a nested error path from the loop
body. Zero-resident identity and resident rendering both re-verified after the
refactor.

## Requirements checked

**Cards render.** Verified against the binary in `fest list <status>`, `fest list`
(all), and `fest status list --status`, with type badges and `1/3 steps` progress
produced by a real standalone run, not hand-written state.

**Zero-resident byte-identical.** Eight invocations diffed between a `main` build
and a branch build; all identical. This is the requirement that matters most given
fest is in production, and it is satisfied by diff rather than assertion. The
mechanism is `FormatResidentList` returning `""` for an empty slice, so no header,
separator, or count is emitted.

**Run progress from fest's own runtime.** `ResolveStandaloneWorkflow`, not camp.
`internal/commands/show` has no camp dependency introduced.

**`fest go`.** Residents offered in completion and resolvable by name, verified with
`--print`. Dungeon residents excluded (`grep -c shelvedres` = 0). Festival priority
unchanged: a festival and a resident on the same stage get the same value, asserted
in `TestCollectNavigationTargets_FestivalsKeepPriority`.

**One resolver.** `go.go:313` calls `CollectNavigationTargets`, the same function
completion uses, so extending it covered both paths with no second wiring.

## The TUI question, answered by running it

The task asked to extend and pty-verify a TUI stage-column board. There isn't one.
Established by driving both binaries in a pty, which corrected my own task-01 note
as well as the naive reading:

- `fest list --watch` is the live-repainting surface this change affects. Verified;
  residents render in the right sections, badges and progress intact, nothing
  clipped. Snapshot in `results/tui-snapshot.txt`.
- The bare `fest go` picker **is** a real bubbletea TUI but already listed residents
  before this change (`FestivalPickerItems` with
  `IncludeUnmarkedFestivalDirectories: true`). Confirmed by driving the pre-change
  binary and getting a byte-identical screen. Unaffected, verified rather than
  assumed either way.
- `charm_list.go` is the `fest go list` shortcut selector, not a stage board.

## Noted, not changed

`ListResidentsByStatus` returns nil when `os.ReadDir` fails, which folds "stage
directory absent" (normal) together with a real I/O error. This matches how every
`ListFestivalsByStatus` caller already behaves (`if err != nil { continue }`), so
matching it is the right call here; diverging would make residents noisier than
festivals for the same condition.

## Gate

```
$ just test unit
✓ ALL 2764 TESTS PASSED
$ golangci-lint run --new-from-merge-base origin/main ./...
0 issues.
```

PR #312, opened 2026-07-30
docs(methodology): introduce lifecycle residents (WF0001 P5/S2)

lancekrogers:
WF0001 phase 005, sequence 02 (fest half). Documentation only. Base `main`.

The camp half is [camp#526](https://github.com/Obedience-Corp/camp/pull/527) — independent, either can merge first.

## Lifecycle residents in the methodology docs

`ready/` and `active/` can hold workitem directories a user moved onto the board without turning them into festivals. The methodology docs described those folders as festivals-only, so a reader had no way to learn what the extra cards in `fest list` are.

Two sections added to the **scaffold source** (`methodology/festivals/.festival/`), not this campaign's generated copy:

- `FESTIVAL_SOFTWARE_PROJECT_MANAGEMENT.md` gains `### Lifecycle Residents` immediately after the lifecycle-flow diagram, where the reader has just been told what `ready/` and `active/` hold.
- `README.md` gains `## Lifecycle Residents` before `## Templates`.

Both describe what a user *sees* — cards in `fest list`/`fest status list`, targets in `fest go`, an informational `fest validate` result — and state the ownership boundary plainly: every resident mechanic is a camp operation, and fest reads the `.workitem` marker it never writes. Mechanics are pointed at, not restated, since fest's methodology docs describe the user's view rather than camp's internals.

## Verified before documenting

Per the festival's anchors-are-claims rule, each described behavior was checked against merged source rather than assumed from the task text: resident recognition and `Classify` (fest #310), resident cards and `fest go` targets (fest #311), and camp's `promote --target ready|active` / `demote` (camp #521, #525).

## No docs regen commit

`just docs` ran clean (exit 0, "Combined reference written") and produced no diff: fest's generated CLI reference was already current, and the methodology docs are not cobra-generated. An empty commit would have been noise, so there isn't one.

## Gate

`just test unit`: 2751 passed.

### DIFF
--- a/methodology/festivals/.festival/FESTIVAL_SOFTWARE_PROJECT_MANAGEMENT.md
+++ b/methodology/festivals/.festival/FESTIVAL_SOFTWARE_PROJECT_MANAGEMENT.md
@@ -108,6 +108,29 @@ festivals/
 
 The lifecycle flow is: `planning/ → ready/ → active/ → dungeon/completed/`
 
+### Lifecycle Residents
+
+`ready/` and `active/` can also hold **residents**: workitem directories that a
+user has moved onto the lifecycle board without turning them into festivals. A
+resident is not a festival and never gains phases, sequences, or tasks. It keeps
+its original workflow type, so a design item parked in `active/` is still a design
+item that simply lives on the board.
+
+Fest treats residents as read-only information:
+
+- `fest list` and `fest status list` show them as cards in the stage column, with a
+  type badge and, when the directory carries a standalone `WORKFLOW.md` run, its
+  step progress.
+- `fest go <name>` navigates into one, and completion offers it alongside festivals.
+- `fest validate` reports a resident and exits cleanly rather than complaining that
+  it is a malformed festival.
+
+Fest never creates, moves, or deletes a resident. Promoting a workitem onto the
+board, advancing it between stages, completing it, and returning it to its original
+home are all camp operations (`camp workitem promote --target ready|active`,
+`camp workitem demote`). Fest reads the `.workitem` marker camp writes; it never
+writes it.
+
 ## Festival Types
 
 Festivals come in four types, each designed for different kinds of work:
--- a/methodology/festivals/.festival/README.md
+++ b/methodology/festivals/.festival/README.md
@@ -44,6 +44,19 @@ Run `fest --help` for the full command reference.
     └── orchestration/                         # Multi-agent orchestration patterns
 ```
 
+## Lifecycle Residents
+
+The `ready/` and `active/` directories can contain more than festivals. A workitem
+directory that a user has moved onto the lifecycle board appears there as a
+**resident**: still its original type, never gaining phases or sequences, and never
+mutated by fest.
+
+You will see residents as cards in `fest list` and `fest status list`, as targets in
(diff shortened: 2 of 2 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready)

**Merge-ready.** One defect was found and fixed during implementation (below);
review found nothing further. Posted as a Comment; the Approve verdict is Lance's.

## The defect worth naming: a field that never reached the contract

`outputDungeonJSON` remaps `DungeonItem` into a **local** `jsonItem` struct. Adding
`WorkitemType` to `DungeonItem` therefore did nothing for `--json`, and the unit
tests for `ListParentItems` passed the whole time. Only running the real command
showed the field missing.

This is the exact trap the requirement warned about ("confirm this command's JSON
contract status before assuming it is casual output"). Fixed, commented at the
struct so the next person adding a field sees the indirection, and locked by
`TestOutputDungeonJSON_IncludesWorkitemType`, which also asserts `omitempty` really
omits rather than emitting `""`.

## Requirements checked

**Everything documented was verified in source first**, per the anchors-are-claims
rule: the `ready`/`active` targets, `camp workitem sweep`, `camp workitem demote`,
and both `fresh.yaml` settings with their real values from `settings.go:214-224`.
The tier-1/tier-2 framing and "inferred evidence never auto-promotes" are quoted
from those doc comments, not invented.

**Additive, non-destructive rendering.** The workitem type goes in parens beside the
generic label rather than replacing it, so a plain directory's line is unchanged;
`omitempty` keeps a plain directory and a file byte-identical in `--json`. Both
pinned by tests.

**Best-effort marker read.** An unreadable marker warns and leaves the field empty,
matching this package's existing posture, tested by
`TestListParentItems_UnreadableMarkerIsNotFatal`.

**Shared-function consequence recorded, not hidden.** `ListParentItems` serves
type-local triage too, so `workflow/design/` gains the same field. Special-casing by
path would have been worse; the requirement asked for this to be stated explicitly
and it is.

**Docs regen isolated.** `git show --stat` on the regen commit shows only
`docs/cli-reference/*`, and every changed line traces to the two help edits. fest
produced no regen commit because `just docs` yielded no diff — recorded rather than
manufacturing an empty one.

**Source of truth respected.** fest's methodology edits are in
`projects/fest/methodology/`, and this campaign's scaffolded `festivals/.festival/`
copy is verified untouched.

## Two discrepancies with doc 04, both recorded

Doc 04's "inner crawl of `festivals/.dungeon/` must present residents correctly" has
no surface to attach to:

1. `ListItems` skips every subdirectory, so the crawl never walks the dated status
   buckets, and `Discover()` excludes anything under a `dungeon/` segment.
2. Triage does not reach `festivals/` either — the excluder lists it by name, and
   `camp dungeon add` from inside it resolves the campaign-root dungeon.

Type-local triage is the one real surface. The integration test asserts that actual
behavior rather than forcing a fixture onto a path that does not exist, which is the
right call: a passing test over an unreachable code path would have been worse than
no test.

## Gate

```
camp:  just gate                                        -> PASSED
       go test -tags=integration -run TestDungeonTriage_ -> ok
fest:  just test unit                                    -> 2751 passed
```

PR #313, opened 2026-07-30
fix(commit): stage only festival paths git can match

lancekrogers:
## The bug

The first `fest commit` in a brand-new campaign fails:

```
Error: staging files: git add failed: fatal: pathspec '.campaign/fest' did not match any files
```

`festivalScopedPaths` listed `.campaign/fest` and `festivals/.festival/.state` unconditionally when
staging festival-scoped paths at a campaign root. Neither directory exists until fest first navigates
in the campaign, and git refuses an entire `git add` over one pathspec it cannot match, so the
unmatched path took the festival directory down with it and nothing was committed.

This is the shipped bundle's first-run path. A user who runs `camp init`, creates a festival, and
commits hits the failure before anything else. Found by CA0004's `commit_drain` integration test,
which had been working around it with a `mkdir -p` in the fixture.

## The fix

`festivalScopedPaths` now returns only pathspecs git can resolve. A candidate qualifies if it is
present in the working tree, or absent but still tracked. The second case matters: `git add` on a
tracked path that was deleted succeeds and records the deletion, so filtering on working-tree
existence alone would have silently dropped deletions of these paths from festival commits. Existence
is checked with `os.Lstat` first, and `git ls-files` runs only for candidates that are missing from
disk, so a lived-in campaign where everything exists makes no extra git calls at all.

Nothing is created on demand. A commit command should not have filesystem side effects, and the
absent directories carry no user data to lose.

Both call sites now skip staging when the filtered list is empty: nothing git can match means nothing
to stage, and passing the empty list on would fail the commit with commitkit's `ErrNoFilesSpecified`
over paths that hold no user data. In practice the list is never empty (the festival directory always
exists), so this is a guard against a future caller, not a live path.

Behavior for existing campaigns is unchanged: when the paths exist, the same list reaches staging in
the same order.

## Tests

New `tests/integration/commit_fresh_campaign_test.go` builds a campaign with the released `camp`
binary fest links against, asserts that `camp init` really leaves both fest-owned state directories
unborn, and proves the first `fest commit` succeeds and carries the festival files. It also leaves an
unrelated dirty file at the campaign root and asserts it stays untracked, which pins the
stage-everything failure mode a fix for this bug can easily introduce.

`tests/integration/commit_drain_test.go` loses its `mkdir -p .campaign/fest festivals/.festival/.state`
workaround, so the drain test now exercises the fix too. Its baseline `camp commit` stays: that is a
separate camp-side unborn-branch bug being fixed independently.

Unit tests in `internal/commands/commit` now build real fixtures under `t.TempDir` instead of
asserting over string paths that never existed, and cover the fresh campaign, a partially populated
campaign, the tracked-but-deleted path, and a cancelled context.

### Verification

Both integration tests were run against a binary built from `origin/main` to confirm they fail with
the exact reported error, then against this branch:

| Command | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./...` | pass (102 packages) |
| `golangci-lint run ./...` (`just lint all`) | 0 issues |
| `golangci-lint run --build-tags integration ./tests/integration/...` | 29 pre-existing issues, none in touched files |
| `go test -tags integration ./tests/integration/ -run TestCommit -v` | 5/5 pass |
| `go test -tags integration ./tests/integration/` (full suite) | pass, 184s |
| Both commit tests against a pre-fix binary | fail with `pathspec '.campaign/fest' did not match any files` |

## Tradeoffs and caveats

- `trackedInGit` adds a `git ls-files` call per missing candidate. That is at most two calls, only in
  campaigns where the directories are absent, and zero in the healthy case.
- If `ls-files` fails for any reason, the path is dropped rather than guessed at. A cancelled context
  reaches this path, which is why the unit test pins that behavior.
- The linked-submodule root commit goes through the same helper and is covered by the unit tests, but
  the new container test exercises the campaign-root branch only. The submodule branch already fails
  soft there (a skipped root commit warning rather than a failed commit), so the first-run path was
  the one worth pinning in a container.
- `internal/commands/status` solves the same problem a different way (stage each path, tolerate the
  ones that no longer exist). This change does not touch it; the batched staging call here needs the
  filter up front to keep guard outcome reporting intact.

### DIFF
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -200,9 +200,10 @@ func runCommit(cmd *cobra.Command, args []string) error {
 				result.Error = err.Error()
 				return outputResult(result)
 			}
-			// An empty list means git cannot match any festival-scoped path;
-			// commitkit reads no files as "stage everything", the opposite of
-			// scoped staging.
+			// An empty list means git can match no festival-scoped path, so
+			// there is nothing to stage. Passing it on would fail the commit
+			// with commitkit's ErrNoFilesSpecified over paths that hold no
+			// user data; the commit proceeds on whatever is already staged.
 			if len(paths) > 0 {
 				if stageErr := stageFiles(ctx, ws.Root, report, paths...); stageErr != nil {
 					result.Success = false
@@ -625,8 +626,8 @@ func commitCampaignRoot(ctx context.Context, campaignRoot string, paths []string
 	}
 
 	// No matchable festival-scoped path is the same silent skip as no changes:
-	// there is nothing for the root commit to carry, and an empty list would
-	// otherwise reach commitkit as "stage everything".
+	// there is nothing for the root commit to carry, and commitkit would
+	// reject the empty list with ErrNoFilesSpecified rather than stage it.
 	if len(paths) == 0 {
 		return "", nil
 	}
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -200,10 +200,15 @@ func runCommit(cmd *cobra.Command, args []string) error {
 				result.Error = err.Error()
 				return outputResult(result)
 			}
-			if stageErr := stageFiles(ctx, ws.Root, report, paths...); stageErr != nil {
-				result.Success = false
-				result.Error = stageErr.Error()
-				return outputResult(result)
+			// An empty list means git cannot match any festival-scoped path;
+			// commitkit reads no files as "stage everything", the opposite of
+			// scoped staging.
+			if len(paths) > 0 {
+				if stageErr := stageFiles(ctx, ws.Root, report, paths...); stageErr != nil {
+					result.Success = false
+					result.Error = stageErr.Error()
+					return outputResult(result)
+				}
 			}
 		} else {
 			// Fallback: stage all (non-campaign workspace)
@@ -550,6 +555,11 @@ func isInSubmodule(campaignRoot string) (bool, string) {
 //   - .campaign/fest/ (navigation state)
 //   - festivals/.festival/.state/ (global festival event log)
 //   - The submodule pointer path (if submoduleRelPath is non-empty)
+//
+// Only paths git can match are returned. The two fest-owned state directories
+// appear the first time fest navigates in a campaign, so a campaign fresh from
+// `camp init` has neither, and a single unmatched pathspec fails the whole
+// `git add` — which made the first `fest commit` in a new campaign impossible.
 func festivalScopedPaths(ctx context.Context, campaignRoot, festivalPath, submoduleRelPath string) ([]string, error) {
 	if err := ctx.Err(); err != nil {
 		return nil, errors.Wrap(err, "context cancelled")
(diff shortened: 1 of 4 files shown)


internal/commands/commit/commit.go
```diff
@@ -200,10 +200,15 @@ func runCommit(cmd *cobra.Command, args []string) error {
 				result.Error = err.Error()
 				return outputResult(result)
 			}
-			if stageErr := stageFiles(ctx, ws.Root, report, paths...); stageErr != nil {
-				result.Success = false
-				result.Error = stageErr.Error()
-				return outputResult(result)
+			// An empty list means git cannot match any festival-scoped path;
+			// commitkit reads no files as "stage everything", the opposite of
```
obey-agent: **[nit]** Comments at both empty-list guards claim that commitkit treats an empty file list as "stage everything". Against fest's pinned camp dependency (`v0.4.0-rc.1`), `commitkit.StageFilesWithOutcome` already rejects `len(files)==0` with `git.ErrNoFilesSpecified`; only the lower-level `git.Stage` / `StageWithGuard` path treats empty as stage-all. The empty-list guards are still correct and useful (especially `commitCampaignRoot`'s silent skip vs a hard "no files specified" failure / warning), but the stated rationale is inaccurate and could mislead a future reader about what would happen if the guards were removed.

**Suggestion:** Rephrase both comments to say the empty list is skipped so fest does not surface `ErrNoFilesSpecified` (and, for the root path, so "nothing to carry" stays a silent no-op like "no changes"), optionally noting that the underlying guard path would stage-all if an empty list ever reached it without commitkit's check. Same wording fix applies to the parallel comment at `commitCampaignRoot` (~line 629: "otherwise reach commitkit as \"stage everything\"").

obey-agent (commented):
## Summary

PR #313 correctly fixes first-run `fest commit` in a fresh campaign by filtering festival-scoped pathspecs to only those git can match (on disk via `os.Lstat`, or tracked-but-deleted via `git ls-files`), so an absent `.campaign/fest` or `festivals/.festival/.state` no longer aborts the whole `git add`. Both staging call sites guard empty lists, unit coverage exercises fresh/partial/deleted/cancelled cases with real fixtures, and the new integration test pins success plus scoped staging against a real `camp init` tree. Residual risk is low; residual note is that a couple of comments misstate commitkit's empty-list contract (they still document the right *fest* behavior for the wrong reason).

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

lancekrogers:
Good catch on the nit, and it was wrong in three places, not two.

`commitkit.StageFilesWithOutcome` rejects `len(files) == 0` with `git.ErrNoFilesSpecified`; the
stage-everything behavior belongs to the lower-level `git.StageWithGuard`. The comments had borrowed
commitkit's own note about that lower layer and attached it to the exported entrypoint fest calls.

`cf85bad` rewrites both guard comments to the real rationale (nothing git can match means nothing to
stage, and passing the empty list on would fail the commit over paths holding no user data), fixes
the same claim in the new integration test's comment, and corrects the PR description. The guards
themselves are unchanged, as are all behaviors. `go build ./...`, `go vet ./...` (plus the
integration build tag), and `go test ./internal/commands/commit/...` pass.

PR #315, opened 2026-08-01
Replace the evidence: embed contract with a judge orientation payload

lancekrogers:
Removes the `evidence: embed` contract and turns the approval judge's request into an orientation payload: fest says where the work is, having already guaranteed it exists, and the judge reads what it needs.

Companion PR: Obedience-Corp/obey#235 (the consumer half). Either order is safe to merge, since every field involved is additive JSON.

## Why

fest computed up to 256KB of file content per checkpoint and shipped it as `evidence_files`. `ob judge`, the only judge in the stack, has no such field on its request struct, so every byte was parsed and discarded.

The knob also never reached users: hooks have not shipped through a festival release, which is the only channel fest is installed from. Same reasoning that removed the flat `approval_judge` key in #297.

## What changed

**Embed contract removed.** `BuildEvidenceFiles`, `EvidenceFile`, `EvidenceEmbedCapBytes`, the `EvidenceMode` type and constants, `ResolvedHook.Evidence`, `HookDefinition.Evidence`, `approvalJudgeRequest.EvidenceFiles`, and the `evidence` column on `fest hooks list`.

`resolveJudgeHookContext` served both embed and the audit source layer while being named for neither. Narrowed to `resolveJudgeHookSource`.

**Readiness gate proven, not assumed.** Three tests drive `prepareAutoJudgeReadiness` against real files with a judge runner that fails the test on any invocation, so a reordering that launches the judge before readiness fails loudly instead of turning a free deterministic block into a paid rejection. Mutation-checked: removing the `info.Size() > 0` term in `hooks.WithinRoot` makes the zero-byte test fail, and it recovers on restore.

**Ledger records what the judge was offered.** New `judge_evidence_offered` and `judge_working_dirs_offered` on `wf_judge_started`.

## Two contract changes worth calling out

**1. `fest hooks list --json` no longer emits `evidence`.** Agents parse this. The package had 0% coverage while this PR changed its contract, so `commands_test.go` now pins the key set, asserting `evidence` is *absent* rather than that it reports `"paths"` — a value assertion would have passed either way.

**2. An unrecognized `evidence:` key is ignored, not rejected.** Deliberate, and asserted by `TestLoadWorkspaceConfig_RetiredEvidenceKeyIsIgnored`. Rejecting would stop every gate for anyone whose config still carries it. The sequence-goal criteria asked for "ignored **with a clear message**"; I took silent, because a warning would fire for exactly two configs that never leave this machine and that get cleaned up anyway. Adding permanent warning machinery for that would recreate the problem this work came from: a deprecation warning for something the public channel never saw. Easy to overrule if you disagree.

## "Offered", not "read"

The audit trail loses something real here. Inlining made judge inputs reproducible as a side effect: the prompt was derivable from the request. An inspecting judge chooses what to open, so two runs of one checkpoint can consult different files and nothing fest observes distinguishes them.

The field names are the mitigation. fest knows what it put in the prompt; it does not know what the judge opened, and a field named `read` would claim provenance the record does not have.

Two alternatives rejected, written up in the contract doc so it is not relitigated:

- **Judge self-reports what it read.** Authored by the component under audit. A judge that fabricates an inspection fabricates the list. Reads like provenance while being model-authored.
- **Capture real tool calls from the daemon session.** Unfakeable, and the most expensive; couples the judge audit trail to session internals. Worth revisiting if a disputed verdict ever needs adjudicating.

Recorded at launch, so a judge that crashes or times out still leaves behind what it was asked to look at.

## Bug found and fixed along the way

The readiness block named the missing deliverable only in a structured field, so the terminal — and the durable rejection reason — read `listed evidence missing or empty` with no indication of which file, sending the operator back to GATES.md.

```
Reason: approval readiness: approval readiness failed: listed evidence missing or empty: output_specs/PRESENTATION.md
Hint: create every file listed under **Evidence:**, then re-submit with 'fest workflow judge': output_specs/PRESENTATION.md
```

## Corrections to the plan

- `ResolvePhaseRelative` was described in the plan and the old docs as the readiness signal. It is not: the gate composes `NormalizeEvidencePath` and `WithinRoot` directly. It was already test-only on `main` before this work. Doc corrected; the function left in place, with a follow-up intent to either delete it or give it the call site the plan assumed.
- `just lint all` reported 0 issues on a file `gofmt -l` flagged. Fixed here; the gate gap is captured as its own intent.

## Verification

- `just lint all` 0 issues, `gofmt -l` clean
- `just test unit` 2776/2776
- `just test all` adds 75 containerized integration
- `tests/integration/approval_readiness_test.go` passes **unmodified**, which is the evidence that removing embed did not weaken the symlink containment guard


obey-agent (commented):
## Summary

This PR cleanly retires the unused `evidence: embed` / `evidence_files` path, narrows hook resolution to orientation-only payload fields, improves readiness error text by naming missing deliverables, and adds `judge_evidence_offered` / `judge_working_dirs_offered` on `wf_judge_started` with deliberately honest naming. Call sites for `BeginJudge` / `EmitJudgeStartedEvents` look complete, YAML/JSON ignore of the retired key is intentional and tested, and the new readiness / hooks-list contract tests pin the important behavior (including absence of `evidence` in `fest hooks list --json`). No correctness bugs found in the production path.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #317, opened 2026-08-04
fix(gates): give artifact_review gate steps default evidence

lancekrogers:
Closes the recurring "the phase gate judge rejects for structural reasons" problem. Intent: `fest-phase-gate-templates-give-20260727-123326` (open since 2026-07-27, hit again in CA0004 phase 001, OI0007, and JE0001).

## The bug

`DefaultEvidencePaths` only recognizes presentation steps, names containing `GOAL`, and names containing `COMPLETE`. Everything else returns `nil`.

Probing the real resolver with every gate step name fest ships:

```
PHASE GOAL           ok   PHASE_GOAL.md
COMPLETENESS         ok   output_specs/context.md
APPROVAL             NIL  <- judge gets GATES.md only
COVERAGE             NIL
DOCUMENTATION        NIL
EVIDENCE             NIL
SEQUENCE OUTCOMES    NIL
ACTIONABILITY        NIL
FOLLOW-UP            NIL
INCORPORATION        NIL
QUALITY              NIL
STRUCTURE            NIL
```

10 of 19 shipped gate steps resolve to nothing. `GATES.md` is the *step definition*, so those judges are asked "did each sequence achieve its stated goal?" while holding a document containing only that question. They can do nothing but reject, and the rejection text is accurate: "the step declared no deliverables."

The `EVIDENCE` step resolving to no evidence is the clearest symptom.

This got sharper after the judge capability gate landed: a judge on a chat provider gets no tools, so it cannot compensate by reading `working_dirs`. Whatever is not inlined does not exist to it.

## The fix, and the one it is deliberately not

Resolve from the declared `CheckpointClass` rather than adding more step-name substrings, so a phase type that adds a gate step gets evidence without anyone extending a list. `operator_attestation` stays excluded, since `checkAutoJudgeAllowed` already refuses to auto-judge it.

**These are defaults, not an `**Evidence:**` declaration, and that distinction is the whole design.** My first instinct was to add `**Evidence:**` blocks to the templates. That would have broken every festival:

```go
// approval_readiness.go
if !wf.IsPresentationStep(step) && len(step.EvidencePaths) == 0 {
    return nil   // readiness skipped
}
```

Declaring evidence makes a step explicit, and explicit evidence requires **every** listed file to exist and be non-empty (`TestCheckApprovalReadiness_ExplicitEvidenceRequiresEveryFile`). Scaffolded phases have none of them, so every gate would hard-block on readiness before the judge ran, which is strictly worse than the current rejection. Defaults skip readiness and are filtered to existing files, so listing a file that does not exist costs nothing.

`TestCheckApprovalReadiness_GateStepDefaultsDoNotBlock` pins that behavior.

## Second half of the gap

Implementation phases scaffold no `output_specs/` at all, unlike ingest, so there was nothing for the resolver to find even once fixed. All six `GATES.md` templates now carry a short section saying what to write and where, as prose rather than as a declaration, for the reason above. It also tells the reader why the conventional files are left implicit, so the next person does not add the `**Evidence:**` block I almost did.

## Risk introduced, and how it is covered

The new prose adds a `##` heading to a file the step parser reads. A heading mistaken for a step would renumber every real gate and desync workflow state. `TestShippedGateTemplatesParseToStepsOnly` parses all six shipped templates and asserts the exact step counts (4/3/3/3/3/3) and that no prose becomes a step.

## Verification

`just gate` passes: stable and dev command surfaces plus the containerized integration suite. `gofmt` clean.

New tests: 4 resolver cases (shipped gate names, attestation stays empty, unclassed steps unchanged, explicit still wins), 1 readiness non-blocking case, 1 template parse case.

### DIFF
--- a/methodology/festivals/.festival/templates/phases/implementation/GATES.md
+++ b/methodology/festivals/.festival/templates/phases/implementation/GATES.md
@@ -10,9 +10,11 @@ This gate verifies the implementation phase achieved its goal and produced worki
 
 ## Before you submit any step below
 
-The judge reads what this phase wrote down. It cannot browse your repositories,
-and on a provider without tool access it cannot open anything at all, so work
-that is real but only visible in a working directory does not reach it.
+Whether the judge can open your repositories depends on its provider:
+CLI-backed judges read the declared working directories themselves, while
+judges without tool access see only what this phase wrote down. Writing the
+essentials here keeps the gate independent of which judge you drew, and leaves
+the festival a durable record after the worktree is gone.
 
 Write the phase's evidence into `output_specs/` before submitting the first
 step. These are picked up automatically; you do not need to list them:
--- a/methodology/festivals/.festival/templates/phases/implementation/GATES.md
+++ b/methodology/festivals/.festival/templates/phases/implementation/GATES.md
@@ -8,6 +8,32 @@ fest_parent: [REPLACE: PHASE_ID]
 
 This gate verifies the implementation phase achieved its goal and produced working deliverables.
 
+## Before you submit any step below
+
+The judge reads what this phase wrote down. It cannot browse your repositories,
+and on a provider without tool access it cannot open anything at all, so work
+that is real but only visible in a working directory does not reach it.
+
+Write the phase's evidence into `output_specs/` before submitting the first
+step. These are picked up automatically; you do not need to list them:
+
+| File | What belongs in it |
+| --- | --- |
+| `output_specs/requirements.md` | each required outcome from PHASE_GOAL.md, with the raw command output that verifies it |
+| `output_specs/context.md` | where the work landed: commits, PRs, branches, build provenance |
+| `output_specs/constraints.md` | each quality standard, with its proof |
+| `output_specs/PRESENTATION.md` | the summary, and anything that needs a human decision |
+
+Paste real terminal output, not summaries of it. "All tests pass" is an
+assertion; the test run is evidence. A gate that rejects for missing evidence is
+usually correct, and the fix is to attach the artifact rather than to re-run the
+judge.
+
+Only add an `**Evidence:**` list to a step when you want to name files beyond
+these. Every path in such a list must exist and be non-empty or the step blocks
+before the judge runs, which is why the conventional files above are left
+implicit.
+
 ---
 
 ## Step 1: PHASE GOAL — Verify Goal Achievement
--- a/internal/commands/workflow/approval_hardening_test.go
+++ b/internal/commands/workflow/approval_hardening_test.go
@@ -502,3 +502,25 @@ func TestEvaluateApprovalJudge_PassesWorkDir(t *testing.T) {
 		t.Fatalf("WorkDir = %q, want /festival/root", gotDir)
 	}
 }
+
+// DefaultEvidencePaths now resolves phase artifacts for artifact_review gate
+// steps. That must not turn into a readiness requirement: defaults are hints
+// filtered to existing files, while an explicit **Evidence:** list is a promise
+// every file is present. If this test ever fails, every phase gate whose
+// artifacts are incomplete will hard-block before the judge runs.
+func TestCheckApprovalReadiness_GateStepDefaultsDoNotBlock(t *testing.T) {
+	step := wf.WorkflowStep{
+		Number:          3,
+		Name:            "QUALITY",
+		Checkpoint:      wf.CheckpointUserApproval,
+		CheckpointClass: wf.CheckpointClassArtifactReview,
+		// No EvidencePaths: the paths come from DefaultEvidencePaths.
+	}
+	// Inspector reports every conventional artifact as absent.
+	err := checkApprovalReadinessWithInspector("/phase", step, func(_, _ string) (bool, error) {
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("gate step with default evidence must not block readiness, got %v", err)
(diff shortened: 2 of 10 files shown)


obey-agent (commented):
## Summary

Clean, well-scoped fix for a real judge failure mode. `DefaultEvidencePaths` only resolved GOAL/COMPLETE/presentation names, so most shipped `artifact_review` gate steps (QUALITY, SEQUENCE OUTCOMES, EVIDENCE, etc.) handed the model only `GATES.md` and structural rejections were inevitable—especially after tool-less judges.

The implementation chooses the right lever: key off declared `CheckpointClassArtifactReview` rather than growing name substrings; keep `operator_attestation` empty; keep defaults distinct from explicit `**Evidence:**` so readiness does not hard-block scaffolded phases. Refactoring the GOAL set into `phaseArtifactEvidence()` is a small readability win without changing that path’s semantics. COMPLETE remains the historical smaller set; intentional and fine.

Template prose correctly avoids declaring `**Evidence:**` (which would flip readiness to require every path). New `## Before you submit…` heading is safe under `stepHeaderRe` (`## Step N:` only), and `TestShippedGateTemplatesParseToStepsOnly` locks step counts. Tests cover the shipped nil names, attestation exclusion, unclassed steps, explicit wins, and readiness non-blocking.

Navigator already normalizes legacy blocking gates to `artifact_review` before judge, so field checks on `CheckpointClass` line up with the live path. Shipped WORKFLOW `artifact_review` steps are PRESENT-named and still hit the presentation branch first.

No high-conviction defects. Approve as-is.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

No blocking issues found.

methodology/festivals/.festival/templates/phases/implementation/GATES.md
```diff
@@ -8,6 +8,32 @@ fest_parent: [REPLACE: PHASE_ID]
 
 This gate verifies the implementation phase achieved its goal and produced working deliverables.
 
+## Before you submit any step below
+
+The judge reads what this phase wrote down. It cannot browse your repositories,
```
obey-agent: **Blocking: this claim is false for the most common judge configuration, in all six templates.** "It cannot browse your repositories" contradicts what fest itself ships: fest sends `working_dirs` and `campaign_root` on every approval request (internal/commands/workflow/approve.go, since #315), and ob renders a "Where the work landed" section telling CLI-backed judges (claude-code, codex, grok) they may read those directories directly. Under obey PR #241 a chat model declared `model_tools: true` inspects too. Baking a capability assertion about another repo's binary into six long-lived production templates guarantees drift.

Suggested wording that stays true under every configuration: "Whether the judge can open your repositories depends on its provider: CLI-backed judges can read the declared working directories themselves; judges without tool access see only what this phase wrote down. Write the essentials into output_specs/ so the gate does not depend on which judge you drew — and so the festival keeps a durable record after the worktree is gone."

That last clause also softens the universal "paste real terminal output" instruction, which otherwise entrenches the transcription tax measured in IQ0001 (feedback/observations/002: four of five rejections cleared by pure re-serialization) as permanent guidance rather than the fallback for non-inspecting judges. Same fix applies to the other five GATES.md templates.

obey-agent (commented):
**Verdict: Request Changes** (one wording fix across the six templates; the code is ready)

**Overview**: This correctly diagnoses and closes a real, recurring failure: 10 of 19 shipped gate steps resolved to zero evidence, so their judges held the question and none of the answer and could only reject. Keying defaults on the declared CheckpointClass instead of more name substrings is the right mechanism, and the defaults-versus-explicit distinction (defaults skip readiness and filter to existing files; explicit lists hard-block) is well understood and well pinned by tests. Independently verified: build clean, race tests green in both touched packages, template step counts pinned (4/3/3/3/3/3), legacy gates normalize to artifact_review so pre-class festivals benefit too, and explicit **Evidence:** still wins.

**Key Findings**:
- Blocking (inline comment on implementation/GATES.md:13): "It cannot browse your repositories" is false for CLI-backed judges today — fest itself sends working_dirs on every approval request and ob orients capable judges to read them — and diverges further under obey#241 (model_tools). All six templates carry the claim; suggested capability-neutral wording in the inline comment. The adjacent universal "paste real terminal output" guidance should stay motivated by the durable-record value, not by a false capability claim, or it entrenches the re-serialization tax IQ0001 measured (feedback/observations/002).
- Non-blocking: the GOAL/COMPLETE name-substring cases now partially duplicate the class branch (COMPLETENESS gets the narrower 3-file list because the name case fires first). Fine as minimal change; a future simplification candidate once WI-014c81's fest-side work lands.

**What's Done Well**: The PR body is a model — the probe table of real resolver outputs, the explicit "the fix it deliberately is not" section explaining why **Evidence:** declarations would hard-block every scaffolded phase, and the named parser risk with a test that pins exact step counts. The operator_attestation exclusion matching checkAutoJudgeAllowed shows the author traced the full execution path.

**Staff Standard**: Yes for the Go changes — I would ship them as my own. The template prose needs the one correction before this merges, because fest is in production and templates outlive binaries.

lancekrogers:
Review finding addressed in ebd3852: all six templates now state judge repository access capability-neutrally (CLI-backed judges read declared working directories themselves; only tool-less judges are limited to the phase's documents), keeping the output_specs guidance motivated by judge-independence and the durable record rather than a false capability claim. TestShippedGateTemplatesParseToStepsOnly still green.

PR #318, opened 2026-08-05
fix(watch): route polling fallback through dispatch serialization [intent: fest-runpolling-fallback-bypasses-dispatchmu-20260717-145842]

lancekrogers:
## Intent

- **ID:** `fest-runpolling-fallback-bypasses-dispatchmu-20260717-145842`
- **Title:** fest: runPolling fallback bypasses dispatchMu serialization (latent, not live)
- **File:** `.campaign/intents/active/fest-runpolling-fallback-bypasses-dispatchmu-20260717-145842.md`

Residual from `fest-serialize-watch-onchange-dispatch-20260712-001747`, which shipped as PR #260.

## This is latent hardening, not a live bug

Nothing is broken in `fest` today. `WatchWithFallback` runs the polling fallback only when `New` fails, so exactly one of the two paths is ever live in a single call, and a lone polling loop dispatches inline on its own ticker goroutine and therefore cannot overlap itself. Every other consumer of the package (`internal/commands/list`, `internal/commands/show`, `internal/commands/progress`, `internal/tui/explore`) uses `watch.New` directly and never reaches the polling path at all.

The problem is that the invariant PR #260 established did not cover the fallback. Before this change, `runPolling` was a free function that called `onChange()` directly with no access to `dispatchMu` or the `stopped` flag, both of which lived as private fields on `Watcher`. So the guarantee "onChange never overlaps itself and never fires after stop" was a property of one code path rather than a property of the package. If a future change ran both paths together, or added a second scheduler sharing a callback, the overlap would return with no test or type to catch it.

I verified the intent is still accurate on this branch before changing anything: `internal/watch/watcher.go:329` (pre-change numbering) called `onChange()` straight from the ticker loop.

## Design

Extract the serialization state into a small unexported `dispatcher` type that owns three things together: the callback, the mutex held across the whole call, and the `stopped` gate.

- `Watcher` now holds a `*dispatcher` instead of the `onChange`, `dispatchMu`, and `stopped` fields. `Watcher.dispatch(gen)` keeps its generation guard unchanged and then calls `d.dispatch()`; `cancelPendingTimer` calls `d.stop()`.
- `runPolling` takes a `*dispatcher` and fires each tick through `d.dispatch()`.
- `WatchWithFallback` builds one dispatcher up front and hands it to whichever path runs, via a new unexported `newWatcher(cfg, d)` that `New` also delegates to.

The reason for binding the callback to the dispatcher rather than passing a `func()` per call is that it makes the pairing unforgeable: you cannot route two different callbacks through one gate, or one callback through two gates, by accident. And because `WatchWithFallback` shares a single dispatcher across both paths, the fix is not just cosmetic uniformity: if the two paths ever do run concurrently, they now serialize against each other rather than overlapping silently, which is precisely the failure mode the intent named.

Ownership of `stop` stays with whoever created the dispatcher. `runPolling` does not stop it, because a dispatcher it shares with a live `Watcher` is not its to tear down. `WatchWithFallback` stops it in the fallback branch; `Watcher.Close` already stops it in the other. This keeps `runPolling` composable while preserving the "nothing fires after return" property, since it dispatches inline and so has no callback in flight when it returns.

Behavior is otherwise identical: debounce and MaxWait semantics, close semantics, and fallback-only-when-New-fails are untouched. **No public API change.** `New`, `Config`, `Watcher`, `AddPath`, `AddTree`, `Watch`, `Close`, and `WatchWithFallback` all keep their existing signatures, so the five call sites outside the package are unaffected.

## Tests

Three new polling tests plus a table-driven unit test for the dispatcher:

- `TestRunPolling_SharedDispatcherNeverOverlapsWatcher` runs a polling loop and a `Watcher` debounce storm against one shared dispatcher with a callback slower than either scheduler's cadence, asserting max concurrent invocations stays at 1. It also increments a deliberately unsynchronized counter so `-race` fails if the two ever overlap.
- `TestRunPolling_ReturnsOnContextCancel` asserts `runPolling` returns nil on cancellation and that the callback count is frozen afterward.
- `TestRunPolling_StoppedDispatcherSilencesTicks` asserts ticks arriving after `stop` are swallowed, which is the gate `Close` relies on.
- `TestDispatcher` covers dispatch, repeated dispatch, dispatch-after-stop, stop idempotence, and mid-sequence stop.

I confirmed the new tests actually catch the old behavior rather than passing vacuously. Temporarily reverting the polling loop to call the callback directly produced both a hard failure and a genuine data race between a polling tick and a watcher debounce timer:

```
--- FAIL: TestRunPolling_StoppedDispatcherSilencesTicks (0.12s)
    watcher_test.go:463: polling fired 6 callbacks after the dispatcher stopped, want 0
WARNING: DATA RACE
Read at 0x00c000188800 by goroutine 14:
  ...watch.(*dispatcher).dispatch()
  ...watch.(*Watcher).dispatch()
Previous write at 0x00c000188800 by goroutine 12:
  ...watch.runPolling()
```

Existing tests were updated only for the new construction shape (`dispatcher: newDispatcher(fn)` in place of the `onChange` field); none of their assertions changed.

## Gates

| Gate | Result |
| --- | --- |
| `go build ./...` | clean |
| `just lint vet` (`go vet ./...`) | clean |
| `just lint all` (`golangci-lint run ./...`) | 0 issues |
| `just test unit-raw` | all packages ok, no failures |
| `go test -race -count=3 ./internal/watch/...` | ok, 11.969s |
| `go test -race` on the four consumer packages | ok |

One note on the lint run: the first `just lint all` reported a stale `staticcheck` finding in `../gate-evidence-defaults/internal/guidance/workflow/navigator_format_judge_test.go`, a file in a sibling worktree that no longer exists (that work merged as PR #317, which is this branch's base). It was golangci-lint cache pollution, not a finding against this change. After `golangci-lint cache clean` the run reports 0 issues. Worth knowing if other worktree branches hit the same phantom failure.

## Tradeoffs and residual risk

- The dispatcher is one shared mutex, so a slow callback on a shared dispatcher backpressures the other scheduler. That is the intended semantics (it is the same backpressure PR #260 introduced for the fsnotify path) but it means a pathologically slow render will now also delay polling ticks if both ever share a dispatcher. Ticks are dropped rather than queued, which matches the existing coalescing behavior.
- `Watcher.dispatcher` is a pointer and is never nil-checked. `New` and `newWatcher` always set it, but a hand-rolled `&Watcher{}` literal that skips it will panic on first dispatch. I chose not to add a defensive nil guard because a missing dispatcher is a programming error, not a runtime condition, and the package's existing style documents invariants rather than defending against internal misuse. Three in-package tests construct `Watcher` literals directly and were updated accordingly.
- The concurrency tests are timing-based, consistent with the rest of the package. They passed `-count=3` under `-race`, but they are sleep-driven and so carry the usual small flake risk on a heavily loaded machine.

### DIFF
--- a/internal/watch/watcher_test.go
+++ b/internal/watch/watcher_test.go
@@ -203,9 +203,9 @@ func TestWatcher_MaxWaitFiresDuringSustainedChanges(t *testing.T) {
 
 	w := &Watcher{
 		config: cfg,
-		onChange: func() {
+		dispatcher: newDispatcher(func() {
 			callCount.Add(1)
-		},
+		}),
 	}
 	defer w.cancelPendingTimer()
 
@@ -242,7 +242,7 @@ func TestWatcher_OnChangeNeverOverlaps(t *testing.T) {
 	}
 
 	w := &Watcher{config: cfg}
-	w.onChange = func() {
+	w.dispatcher = newDispatcher(func() {
 		n := inFlight.Add(1)
 		for {
 			m := maxInFlight.Load()
@@ -255,7 +255,7 @@ func TestWatcher_OnChangeNeverOverlaps(t *testing.T) {
 		// this callback is still running, exercising the serialization path.
 		time.Sleep(15 * time.Millisecond)
 		inFlight.Add(-1)
-	}
+	})
 
 	// Sustained events reset the trailing debounce forever; MaxWait keeps firing
 	// callbacks mid-storm, so without serialization these would overlap.
@@ -286,8 +286,8 @@ func TestWatcher_DispatchIgnoresStaleGeneration(t *testing.T) {
 	// Debounce is long enough that no AfterFunc timer fires on its own; the test
 	// drives dispatch directly to exercise the generation guard deterministically.
 	w := &Watcher{
-		config:   Config{Debounce: time.Hour},
-		onChange: func() { calls.Add(1) },
+		config:     Config{Debounce: time.Hour},
+		dispatcher: newDispatcher(func() { calls.Add(1) }),
 	}
 
 	w.scheduleCallback() // gen == 1
@@ -389,9 +389,9 @@ func TestRunPolling(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 150*time.Millisecond)
 	defer cancel()
 
-	err := runPolling(ctx, 50*time.Millisecond, func() {
+	err := runPolling(ctx, 50*time.Millisecond, newDispatcher(func() {
 		callCount.Add(1)
-	})
+	}))
 
 	if err != nil {
 		t.Errorf("runPolling() error = %v", err)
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
## Review (obey-agent)

**Draft** latent hardening: extract a shared `dispatcher` so the polling fallback and fsnotify path both serialize `onChange` and honor stop-after-teardown.

### Assessment
- `dispatcher.dispatch` holds `mu` for the full callback; `stop` sets `stopped` under the same lock — no post-Close fire, no self-overlap.
- `WatchWithFallback` builds one dispatcher, hands it to `newWatcher` or `runPolling`, and `defer d.stop()` on the poll path.
- PR correctly frames this as residual from #260: not a live bug today (only one path active per call), but the shared invariant is the right long-term shape.

Looks correct → **COMMENT**.

internal/watch/watcher_test.go
```diff
@@ -255,7 +255,7 @@ func TestWatcher_OnChangeNeverOverlaps(t *testing.T) {
 		// this callback is still running, exercising the serialization path.
 		time.Sleep(15 * time.Millisecond)
 		inFlight.Add(-1)
-	}
+	})
 
 	// Sustained events reset the trailing debounce forever; MaxWait keeps firing
 	// callbacks mid-storm, so without serialization these would overlap.
```
lancekrogers, line 261: **[nit]** TestWatcher_OnChangeNeverOverlaps still says cancel waits on dispatchMu. That field was removed; wait is now dispatcher.stop() / d.mu.

**Suggestion:** Update the test comment to match the new dispatcher type.

_(anchored near line 261; reviewer cited 264)_

lancekrogers (commented):
## Summary

Approve. This is clean latent hardening: the polling fallback now shares the same serialization and stop gate as the fsnotify debounce path, without changing public API or live production behavior (`WatchWithFallback` still runs only one path per call).

**What changed well**

- Extracting `dispatcher` (`onChange` + `mu` + `stopped`) makes the package invariant unforgeable: both `Watcher.dispatch` and `runPolling` must go through one gate, and a shared dispatcher serializes two schedulers against each other.
- Lock ordering is safe and unchanged in spirit: `w.mu` is always released before `d.mu` is taken. No nested hold of both locks, so no AB-BA deadlock between generation guards and callback serialization.
- `stop` correctly waits for in-flight callbacks because `dispatch` holds `mu` across the full `onChange` call; once `stop` returns, further dispatches are inert.
- Ownership of `stop` is right: `runPolling` does not tear down a dispatcher it may share; `WatchWithFallback` `defer d.stop()` on the polling branch; `Watcher.Close` / `cancelPendingTimer` cover the fsnotify branch.
- Generation guard remains in front of the dispatcher, so cancel still neutralizes stale `AfterFunc`s even if they race past `timer.Stop()`.
- Tests are purposeful: shared-dispatcher overlap + unsynchronized canary under `-race`, stop silences ticks, cancel freezes count, and table-driven dispatcher unit coverage. The PR description’s temporary-revert evidence (hard fail + data race) is convincing.

**Concurrency / deadlock notes (no issues found)**

| Scenario | Outcome |
| --- | --- |
| Timer mid-`onChange` vs `cancelPendingTimer`/`stop` | `stop` blocks on `d.mu` until callback finishes, then gates off |
| Timer past gen check, then cancel | Either gen mismatch no-ops before dispatch, or `stopped` swallows it |
| Polling tick vs watcher timer on shared `d` | Serialized by `d.mu`; ticks coalesce (block, do not queue) |
| `onChange` → `scheduleCallback` | Safe: only `w.mu`, already released before `d.mu` |
| Double `stop` (Watch cancel then Close) | Idempotent |
| Polling path panic in `onChange` | `dispatch` unlocks via defer; `WatchWithFallback` still runs `defer d.stop()` |

Residual reentrancy footgun (pre-existing, not introduced here): `onChange` must not re-enter the same dispatcher (`dispatch`/`stop`/`Close`/`cancelPendingTimer`), because `sync.Mutex` is not reentrant. Same as old `dispatchMu`.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #319, opened 2026-08-05
fix(commit): flip file-list retry copy to fest commit --commit-large [intent: fest-flip-the-file-list-20260801-011034]

lancekrogers:
## Intent

- **ID**: `fest-flip-the-file-list-20260801-011034`
- **Title**: fest: flip the file-list retry copy from `camp commit --commit-large` to `fest commit --commit-large`
- **File**: `.campaign/intents/inbox/fest-flip-the-file-list-20260801-011034.md`

The intent is **not stale**: the export exists and fest was not using it. But the premise behind it turns out to be incomplete, and that is the most important thing in this PR. Read the "What this does not fix" section before merging.

## The dependency bump

`github.com/Obedience-Corp/camp` v0.4.0-rc.2 to **v0.4.0-rc.3**.

v0.4.0-rc.3 is the newest camp tag and the first one carrying `commitkit.StageFilesWithOptions` (camp PR #530). There is no stable v0.4.0 yet, so the rc is the only release with the export. The bump pulls in nothing else: camp's own `go.mod` hash is byte-identical between rc.2 and rc.3 (`h1:Xko1DyzQ80Yf0lGCLu7ZiYWXTxdxXobRgUzERTMLeVE=`), so `go mod tidy` produced no transitive changes. The full diff outside the two staging call sites is the version line and its two `go.sum` lines.

## What changed

**Call sites flipped** (both stage through a file list that `fest commit` itself builds, and both are re-reachable by rerunning `fest commit --commit-large`):

| Location | Before | After |
| --- | --- | --- |
| `internal/commands/commit/commit.go` `stageFiles` | `StageFilesWithOutcome`, copy named `camp commit --commit-large` | `StageFilesWithOptions` with `StageOptions{CommitLarge: commitLarge}`, copy names `fest commit --commit-large` |
| `internal/commands/commit/commit.go` `commitCampaignRoot` | same | same |

`stageAllChanges` already used the options form and already named fest's flag; it is unchanged apart from the signature change below.

**Call sites deliberately left**: none. The brief anticipated a site where the user genuinely must rerun camp rather than fest. Tracing all three staging call sites, no such site exists: every one of them is staging that `fest commit` performs during a run the user started with `fest commit`, so fest's own flag is the correct retry everywhere.

**Consequent simplification**: because all three call sites now pass the same string, the per-call-site `retryCmd` parameter on `guardRefusalMessage` and `reportStageOutcome` no longer carries information. It collapses into a `commitLargeRetry` constant in `guard.go`. This also removes the comment in `guard.go` explaining the split, which was the CA0004 artifact the intent set out to close.

**Removed a now-false comment** in `stageFiles`: "commitkit exports no options form for file-list staging". That is no longer true as of rc.3.

## What this does not fix, and why you should not read this PR as closing a user-visible gap

The flipped copy is currently **unreachable**. Camp's guard does not run at all on an explicit path list.

`camp internal/git/stageguard.go`, `runStageGuard` returns immediately when the call names paths:

```go
if !isStageEverything(files) {
    return nil, nil, nil
}
```

That check runs *before* the `opts.CommitLarge` check, and the rationale is deliberate and documented in camp:

> The distinction is the design's: an explicit `git add path/to/media.mp4` is an intent, and camp honors it. Only the sweep forms are guarded, because only they can pick up something the user never mentioned.

So for any real path list, `StageFilesWithOptions` and `StageFilesWithOutcome` behave identically: no guard, no `GuardBlockedError`, nil outcome, nothing reported. I verified this rather than inferring it, by calling both forms directly against a fixture repo with a 512 MB sparse file:

```
=== filelist ===   (StageFilesWithOptions, CommitLarge:false, explicit path)
  err=<nil>
  outcome=<nil>
  staged="fest-dir/notes.md\nfest-dir/render.bin\n"

=== sweep ===      (StageAllWithOptions, CommitLarge:false)
  err=<nil>
  outcome=&{Excluded:[{Kind:over_threshold Path:fest-dir/render.bin Size:536870912}] ...}
  staged="fest-dir/notes.md\n"
```

The 512 MB file **was staged** on the file-list path.

**The real gap this uncovered**: fest's campaign-root festival commit is unguarded. The path list fest passes is synthesized by `festivalScopedPaths` (the festival directory, `.campaign/fest`, `festivals/.festival/.state`), not named by the user. Camp's "an explicit path is user intent" premise does not hold for a list fest built on the user's behalf, so an over-threshold file dropped inside a festival directory goes into the campaign root commit with no exclusion, no report, and no way for the user to find out.

Closing that needs a camp-side change (a way to ask for the guard on a synthesized list, since fest cannot switch to a sweep at the campaign root without staging everything there, which is exactly what festival-scoped staging exists to avoid). That is out of scope here and wants its own intent against camp.

I pinned the current behavior in a new integration test, `TestCommitGuard_CampaignRootFileListIsNotGuarded`, written to assert the observed truth and to **fail loudly** if camp starts guarding synthesized lists, with a comment saying to rewrite it as an exclusion assertion at that point.

## Tests

- Rewrote `internal/commands/commit/guard_test.go` as table-driven, error cases first, matching the surrounding style. Added a multi-file over-threshold case and a guard-unavailable case that were not covered before. Every case now also asserts `camp commit --commit-large` is **absent**, so a regression back to the old copy fails rather than passing silently.
- Added `TestCommitGuard_CampaignRootFileListIsNotGuarded` to `tests/integration/commit_guard_test.go`, exercising the campaign-root file-list branch through the container harness. That branch had no integration coverage at all before; the two existing guard tests both cover sweep branches.

## Gate results

```
just fmt          gofmt -w .
just lint vet     go vet ./...                          (clean)
just lint all     golangci-lint run ./...               0 issues.
just test unit-raw                                      all packages ok
                  internal/commands/commit  ok  0.863s
integration (containerized, full suite)
                  tests/integration            ok  204.591s
                  tests/integration/lifecycle  ok   16.772s
```

Guard tests specifically:

```
--- PASS: TestCommitGuard_LargeFileKeptOutInStandaloneWorkspace (4.47s)
--- PASS: TestCommitGuard_CampaignRootFileListIsNotGuarded (2.86s)
--- PASS: TestCommitGuard_LargeFileKeptOutInLinkedSubmodule (1.12s)
```

## Tradeoffs and residual risk

1. **This ships inert copy.** The flipped strings cannot be printed today. It is still worth landing: the old comment was factually false after rc.3, the old copy pointed at the wrong binary if it ever became reachable, and the parameter it justified was dead flexibility. But it does not change anything a user sees, and the PR title should not be read as fixing a reported symptom.
2. **Depending on an rc.** fest now requires camp v0.4.0-rc.3 rather than a stable tag. If v0.4.0 stable lands before this merges, retag the dependency to it.
3. **The new integration test asserts current behavior that is arguably a bug.** That is intentional and commented, but it does mean a reviewer could mistake it for endorsement. It is there to make the gap visible and to break when the gap closes.
4. Test cost: one more 512 MB sparse file in the shared container. Sparse via `truncate`, and the guard measures `lstat` size, so it costs no real disk or time.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.25.6
 
 require (
 	github.com/Masterminds/sprig/v3 v3.3.0
-	github.com/Obedience-Corp/camp v0.4.0-rc.2
+	github.com/Obedience-Corp/camp v0.4.0-rc.3
 	github.com/Obedience-Corp/obey-shared v0.4.6
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
--- a/go.sum
+++ b/go.sum
@@ -14,8 +14,8 @@ github.com/Masterminds/sprig/v3 v3.3.0 h1:mQh0Yrg1XPo6vjYXgtf5OtijNAKJRNcTdOOGZe
 github.com/Masterminds/sprig/v3 v3.3.0/go.mod h1:Zy1iXRYNqNLUolqCpL4uhk6SHUMAOSCzdgBfDb35Lz0=
 github.com/Microsoft/go-winio v0.6.2 h1:F2VQgta7ecxGYO8k3ZZz3RS8fVIXVxONVUPlNERoyfY=
 github.com/Microsoft/go-winio v0.6.2/go.mod h1:yd8OoFMLzJbo9gZq8j5qaps8bJ9aShtEA8Ipt1oGCvU=
-github.com/Obedience-Corp/camp v0.4.0-rc.2 h1:+dydz5vsx1LOiSGDyFv29W+/ML2m/8BaM759D0uqvGI=
-github.com/Obedience-Corp/camp v0.4.0-rc.2/go.mod h1:Xko1DyzQ80Yf0lGCLu7ZiYWXTxdxXobRgUzERTMLeVE=
+github.com/Obedience-Corp/camp v0.4.0-rc.3 h1:XK73U0Vzt5G1qlnCp0Vehm4z8evCoNEsBG3PlJh9PNI=
+github.com/Obedience-Corp/camp v0.4.0-rc.3/go.mod h1:Xko1DyzQ80Yf0lGCLu7ZiYWXTxdxXobRgUzERTMLeVE=
 github.com/Obedience-Corp/obey-shared v0.4.6 h1:nvl2sM/5ri9LfmKpqXyt84T1z2TvNBiv4cpPkRVJyBk=
 github.com/Obedience-Corp/obey-shared v0.4.6/go.mod h1:VZ9rp2YklgdlipSr7aUlOHbmNjrRjif4qHybkLdn6ic=
 github.com/alecthomas/assert/v2 v2.11.0 h1:2Q9r3ki8+JYXvGsDyBXwH3LcJ+WK5D0gc5E8vS6K3D0=
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -369,33 +369,38 @@ func stageAllChanges(ctx context.Context, repoPath string, report io.Writer) err
 	if err != nil {
 		var blocked *commitkit.GuardBlockedError
 		if stderrors.As(err, &blocked) {
-			return errors.New(guardRefusalMessage(blocked, "fest commit --commit-large"))
+			return errors.New(guardRefusalMessage(blocked))
 		}
 		return errors.Wrap(err, "staging changes")
 	}
-	reportStageOutcome(report, outcome, "fest commit --commit-large")
+	reportStageOutcome(report, outcome)
 	return nothingLeftAfterExclusions(ctx, repoPath, outcome)
 }
 
 // stageFiles stages the given paths through commitkit with outcome reporting
 // and fest-rendered GuardBlockedError text — the same contract as stageAllChanges,
 // for festival-scoped and campaign-root path lists.
+//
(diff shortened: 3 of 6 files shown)


obey-agent (commented):
## Review (obey-agent)

**Draft** intent PR: camp `v0.4.0-rc.2` → `rc.3` for `StageFilesWithOptions`, and all fest stage paths now pass `--commit-large` through with a single retry copy: `fest commit --commit-large`.

### Assessment
- `stageFiles` / campaign-root staging correctly switch from `StageFilesWithOutcome` to `StageFilesWithOptions{CommitLarge: commitLarge}`.
- `guardRefusalMessage` / `reportStageOutcome` drop per-call-site retry strings in favor of `commitLargeRetry` — consistent UX.
- PR body honestly documents residual: camp still does not guard synthesized path lists the same way as sweeps, so the list-path refusal branch may remain dead until camp can guard fest-built lists. That is a documented non-goal, not a bug in this flip.

No functional bugs in the fest-side change → **COMMENT**. (Draft is fine to leave draft until camp guard parity if desired.)

lancekrogers (commented):
## Summary

Clean, correctly scoped PR: camp bump + `StageFilesWithOptions` with `CommitLarge` on both file-list sites, unified retry copy to `fest commit --commit-large`. Residual unguarded campaign-root file-list behavior is pinned, not papered over.

**Verdict:** Approve — no open issues.

(Submitted as COMMENT: GitHub does not allow self-APPROVE on own PRs.)

PR #321, opened 2026-08-05
fix(deps): patch reachable goldmark XSS (GO-2026-5320) and bump otel to v1.43.0

lancekrogers:
## What

Two dependency bumps that close real security advisories:

| Dependency | From | To | Advisory |
| --- | --- | --- | --- |
| `github.com/yuin/goldmark` | v1.7.16 | v1.7.17 | GO-2026-5320 (XSS in HTML renderer) |
| `go.opentelemetry.io/otel` (+ `sdk`) | v1.42.0 | v1.43.0 | otel/sdk <= 1.42.0, BSD kenv PATH hijacking |

## Why

The goldmark XSS is **reachable**, not theoretical. `govulncheck` traces it from two call sites:

```
Vulnerability #1: GO-2026-5320
    Cross-site Scripting (XSS) in github.com/yuin/goldmark
    Found in: github.com/yuin/goldmark@v1.7.16
    Fixed in: github.com/yuin/goldmark@v1.7.17
      #1: internal/guidance/selection/output.go:601:32: selection.buildFullTaskContentSection
          calls mdrender.Render, which eventually calls html.Renderer.renderAutoLink
      #2: ... renderImage
      #3: ... renderLink
```

It also reaches `internal/tui/explore` via glamour. Dependabot never surfaced this one; it was found by running `govulncheck` directly.

`otel/sdk` is the only advisory from the Dependabot alert list that was still genuinely open. It is reached solely through `tests/integration` -> `testcontainers-go`, so exposure is test-only, but the bump is free.

v1.7.17 is the minimal patch that closes the advisory and stays on the 1.7 line, so glamour compatibility is unchanged. (v1.8.x exists but is a needless minor bump here.)

## Verification

```
go build ./...                  exit 0
just test unit                  102 packages, 2782/2782 tests passed
just lint all                   0 issues
```

`govulncheck` after the bump no longer reports GO-2026-5320:

```
Vulnerability #1: GO-2026-5856
    Invoking Encrypted Client Hello privacy leak in crypto/tls
    Found in: crypto/tls@go1.26.4
    Fixed in: crypto/tls@go1.26.5
```

The one remaining finding is standard library, not a module dependency, and it is an artifact of the local go1.26.4 toolchain. Release builds pin `go-version: '1.25'`, which resolves to go1.25.12, and 1.25.12 is the patched release for that advisory. No pipeline change needed.

## Notes

No `replace` directives. Both bumps are plain `go get` + `go mod tidy`.


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — same goldmark + otel security bumps as camp #544; go.mod/sum only.

### Assessment
- goldmark 1.7.17 for GO-2026-5320; otel 1.43.0 family for the otel/sdk advisory.
- Sums and require lines match the intended versions; no app code churn.

LGTM from review side once CI is green.

PR #322, opened 2026-08-05
deps: add --ready to expose the execution front

lancekrogers:
## What

Adds `fest deps --ready`: the set of tasks that are unblocked right now.

`fest next` returns a single step by design. Nothing could answer "which tasks
could be started concurrently", even though the dependency DAG already exists.
This exposes the execution front so an orchestrator can fan it out.

```
$ fest deps --ready --all
READY TASKS
───────────
These tasks have all dependencies satisfied and can start now.

  - 01_probe_both_ends 004_PROBE_AND_RESOLVE/01_rsync_probe/01_probe_both_ends.md
  - 01_resolve_command 004_PROBE_AND_RESOLVE/02_artifacts_resolve/01_resolve_command.md
  - 01_migrate_suites 005_TEST_MIGRATION/01_containerize_clone_sync/01_migrate_suites.md
  - 01_transport_docs 006_DOCS_AND_RELEASE/01_docs_and_release/01_transport_docs.md
  - 04_testing 003_COLD_SEED/02_pack_copy_and_bundle/04_testing.md

Ready 5 tasks
```

That output is from a real festival (CS0005) and crosses four phase boundaries,
which is exactly the fan-out `fest next` cannot express.

`--json` emits `{count, tasks}` so a caller can early-exit on zero without
inspecting the array, and keeps absolute paths so it can act on them directly.
Text output trims to festival-relative for readability.

## Why the flag needed three fixes underneath it

`Graph.GetReadyTasks`, `GetParallelGroups` and `TopologicalSort` were already
written and tested. The readiness path was never wired to real state, so all
three of these were latent:

1. **The resolver hardcoded status.** `loadSequenceTasks` set
   `Status: "pending"` on every task and nothing ever updated it, so readiness
   was computed against state that could not change. It now reads
   `progress.ParseTaskStatus`, which is checkbox-based and takes no store, so
   there is no migration side effect on a live navigator.

2. **Status strings did not match.** `GetReadyTasks` skipped tasks whose status
   was `"complete"`, but the progress package writes `"completed"`. Completed
   tasks would never have been filtered out. `Task.IsComplete()` now accepts
   both spellings plus `skipped`, and `blocked` tasks are excluded from the
   ready set.

3. **Soft dependencies gated readiness.** `resolve.go:308` creates edges with
   `required=false` specifically because they express preferred order rather
   than a hard gate, but `GetDependencies` returns every edge regardless.
   `GetRequiredDependencies` filters to hard edges for readiness only; the
   display path still shows all of them.

## Scope note

This extends `fest deps` rather than adding a top-level `fest ready`. The
command already owned graph queries (`--all`, `--critical-path`, `--json`), so
readiness belongs with them.

## Testing

- Table-driven coverage for `IsComplete` across all status spellings, dependency
  status variants, blocked exclusion, and soft-vs-hard dependency gating.
- Existing `TestGraph_GetReadyTasks` passes unchanged (the legacy `"complete"`
  spelling is still honored).
- `just gate` passes, including the containerized integration suite.
- Verified against a real festival, read-only, in both text and `--json` modes.
- `just docs` regenerated; the only drift was the two `fest deps` reference files.

## Follow-on

This is the prerequisite for the fan-out work in WI-3d9b84: the festival plugin
gains a `workflows/` script that consumes `--ready --json` and dispatches the set
through the harness's own concurrency, rather than fest growing a concurrency
runtime of its own.


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — useful orchestrator surface; logic looks right.

### Assessment
- `fest deps --ready` exposes the full execution front (vs single-step `fest next`).
- `GetReadyTasks` now: skips complete **and** blocked; uses **hard** deps only (`GetRequiredDependencies`); treats `completed` / `complete` / `skipped` as done via `IsComplete()` — fixes the progress-package status mismatch.
- Resolve path fills status via `progress.ParseTaskStatus` so readiness is not stuck on placeholder `"pending"`.
- JSON shape `{count, tasks}` with empty array (not null) is a stable orchestrator contract.
- Docs + tests cover soft-deps-ignored, blocked exclusion, multi-status complete.

### Non-blocking
- PR also carries the goldmark/otel dep bumps from #321 — fine if intentional stacking; just be aware of merge overlap with #321.
- `--ready` + `--critical-path` together prefer ready (order in `runDeps`); optional mutual-exclusion message later.

LGTM.

lancekrogers:
Converting to draft. The feature is wrong on its headline behavior.

`--ready` reads task status via `progress.ParseTaskStatus`, which counts
**markdown checkboxes**. But `fest task completed` records completion in the
progress store (`.fest/progress_events.jsonl`) and, when a task file has
parseable frontmatter, in `status:`. It does not tick checkboxes. Most task
documents have no checkboxes at all, in which case `ParseTaskStatus` returns
`pending` unconditionally.

So the readiness filter reads a source that is essentially never populated.

Measured against CT0003, which has 10 completions recorded in its store:

```
completed per store: 10
reported ready     : 15
COMPLETED TASKS LEAKED INTO READY SET: 3
    001_SESSION_STORE/01_schemas_and_run_store/01_triage_schema_types.md
    001_SESSION_STORE/01_schemas_and_run_store/06_fest_commit.md
    001_SESSION_STORE/02_start_and_preflight/01_start_command.md
```

Three of fifteen. The other seven completed tasks are filtered only by accident:
their predecessors are also incomplete-looking, so the dependency check hides
them. The filtering that appears to work is not working.

My "verified against a real festival" claim in the description was hollow. I ran
it, the output looked plausible, and I did not check it against ground truth.

Two further problems in the same change:

1. **Resolver semantics changed for every caller.** Wiring status into
   `loadSequenceTasks` affects `ResolveFestival`/`ResolveSequence` for all
   consumers (`fest deps` display, `internal/guidance/orchestration`,
   `commands/show/roadmap.go`), not just `--ready`. I did not trace those
   callers. The gate passing is not evidence they are unaffected.

2. **A file read per task on every call.** `fest deps` without `--ready` now
   pays N file opens it did not pay before. Not benchmarked, not mentioned.

Fix direction: read status through the progress store
(`progress.ResolveTaskStatus`, which consults the store and then falls back to
markdown) rather than `ParseTaskStatus`, and confirm it against a festival with
known completions before re-opening. Frontmatter status should also count.

lancekrogers:
Fixed and re-opened. Both defects above are closed, and verifying properly turned up a third.

**1. Status source.** Resolution now goes through the progress store first
(`ResolveTaskStatus` over a `LoadReadOnly` store), then frontmatter `fest_status`,
then checkboxes only for documents that have any. That last distinction needed a
new exported `progress.HasCheckboxes`, because `ParseTaskStatus` cannot tell "no
checkboxes ticked" from "no checkboxes at all" and both come back as `pending`.

`LoadReadOnly` matters: the normal `Load` path migrates `workflow_state.yaml` and
deletes it. A read-only query must not do that to a running navigator. There is a
test asserting the file survives.

**2. Resolver semantics.** Reverted. `internal/deps/resolve.go` is now byte-identical
to `main`. The overlay moved into an opt-in `ApplyProgress`, called only from the
`--ready` path, so `fest deps` display, `guidance/orchestration` and `show/roadmap`
are untouched and no longer pay a file read per task.

**3. New, found while verifying: `GetReadyTasks` is phase-blind.** The resolver only
creates edges within a sequence, so on CS0005 it reported tasks from phases 004, 005
and 006 at once while 004 was still open. That is every unblocked leaf in the
festival, not an execution front.

`GetExecutionFront` narrows to the earliest phase with ready work. It is a separate
function on purpose: `GetReadyTasks` has a second caller in
`internal/guidance/selection/selector.go`, and narrowing it there would change task
selection. That is the same mistake as #2 and I was not going to repeat it.

**Still not handled, and now stated in the output.** Phase quality gates are not
evaluated. CS0005 is held at a 003 approval checkpoint with a judge running, which
only the navigator knows about. The text output says so and the JSON carries
`gates_evaluated: false`, so a consumer can distinguish "no gate blocks this" from
"gates were never checked". Making this gate-aware means reaching from a leaf package
into the guidance stack; it should be its own change.

## Verification

Ground truth this time, not eyeballing:

```
festivals/active/camp-triage-CT0003
  completed in store    : 12
  execution front       : 1
  completed leaked      : 0   (must be 0)
  distinct phases       : 1   (must be <=1)
festivals/active/camp-syncclone-fast-cross-machine-CS0005
  completed in store    : 13
  execution front       : 2
  completed leaked      : 0   (must be 0)
  distinct phases       : 1   (must be <=1)
```

CS0005 went from 4 tasks spanning three phases to 2 in phase 004.

The regression tests were confirmed to **fail** against the previous implementation
before being accepted, rather than assumed to be meaningful. Disk-mutation test
confirms `workflow_state.yaml` survives a query. `just gate` passes.

obey-agent (commented):
## Review (obey-agent) — re-review on `e8c7bcd9`

**Verdict: COMMENT** — prior head was incomplete; this status/front fix is the real feature.

### What was wrong / fixed
- Resolver left every task `"pending"` and `ParseTaskStatus` on checkbox-free docs also returns pending, so completed work never left the ready set.
- **`ApplyProgress`** now overlays (1) progress store (LoadReadOnly), (2) frontmatter status, (3) checkbox status only when the doc actually has checkboxes.
- **`GetExecutionFront`** narrows phase-blind `GetReadyTasks` to the earliest open phase (parallel siblings kept) so `--ready` is an execution front, not every unblocked leaf.
- JSON gains `gates_evaluated: false` so consumers don't assume navigator gates were checked.
- Tests pin store completion, frontmatter, checkbox-free stay-ready, execution-front phase cut.

### Non-blocking
- Earliest phase is lexicographic on `PhasePath` — works with zero-padded `001_`…`010_` layout; odd unpadded names could mis-order.
- Phase quality gates still out of band (documented via `gates_evaluated` + `fest next` hint).

No remaining blocking bug. LGTM.

PR #324, opened 2026-08-05
fix(workflow,show): reject blocking checkpoints at standalone creation; teach fest list in show hint

lancekrogers:
## What

Two agent-UX fixes from the 2026-08-05 intent triage, both verified still-broken at origin/main before fixing.

**1. `fest create workflow` no longer mints unrunnable standalone workflows.**
Standalone creation accepted `checkpoint: approval_required` and returned ok, but the standalone runtime refuses to advance past blocking checkpoints (`internal/commands/workflow/advance.go`), so the workflow was dead from birth and the agent only found out at execution time. Creation now rejects it up front with the same message plus an actionable hint (use `verification`/`none`, or target a festival phase with `--path` where `fest workflow approve` exists). Festival-phase creation is unchanged; blocking checkpoints remain valid there.

**2. Bare `fest show` outside a festival now teaches selector discovery.**
The non-TTY hint mentioned `--festival <selector>` but never said how to discover a selector. It now leads with `run 'fest list' to see festivals, then 'fest show <name>'` before the selector and navigation options.

## Verification

- New tests: table-driven `TestRejectBlockingCheckpoints` (rejection plus all non-blocking checkpoint values allowed), `TestNotInFestivalHint_LeadsWithDiscovery` (asserts discovery ordering, not just presence). Existing `TestRunShowCurrent_NonTerminalOutsideFestivalReturnsHint` still passes, so the explicit-selector teaching is preserved.
- Real binary check: creating with an `approval_required` step exits 1 with the hint and writes no `WORKFLOW.md`; bare `fest show` outside a festival prints the new hint.
- Full gate: `go test ./...` clean, `just lint all` 0 issues.

Intents: `fest-create-workflow-accepts-approvalrequired-20260725-215550` (option 1, creation-time validation, as the intent preferred), `fest-show-bare-invocation-outside-20260724-145457`.

### DIFF
--- a/internal/commands/festival/create_workflow_standalone.go
+++ b/internal/commands/festival/create_workflow_standalone.go
@@ -23,6 +23,22 @@ import (
 //
 // Festival-only flags (--position, --path, --festival) are rejected here so
 // users get an explicit error rather than confusing silent ignores.
+// rejectBlockingCheckpoints fails standalone creation for steps the
+// standalone runtime refuses to advance past, so `fest create workflow`
+// cannot mint a workflow that is unrunnable from birth (advance rejects a
+// blocking checkpoint at execution time with the same message).
+func rejectBlockingCheckpoints(input *WorkflowInput) error {
+	for i, step := range input.Steps {
+		if strings.TrimSpace(step.Checkpoint) == "approval_required" {
+			return errors.Validation("standalone workflows do not support blocking checkpoints").
+				WithField("step", i+1).
+				WithField("step_name", step.Name).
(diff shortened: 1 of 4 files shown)


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — two solid agent-UX fixes; no bug found.

### Assessment
1. Standalone `fest create workflow` now rejects `checkpoint: approval_required` at creation (same constraint the standalone advance path already enforces) with a useful hint toward `verification`/`none` or festival-phase workflows.
2. `notInFestivalError` hint leads with `fest list` before `--festival <selector>` so agents can discover selectors.

Tests pin both. LGTM.

PR #326, opened 2026-08-12
feat(hooks): add task_start lifecycle verb with start-stage bindings

lancekrogers:
## Summary

All v1 lifecycle verbs were completion verbs (task_complete, sequence_complete, phase_complete, gate_approve), so no hook could run before work begins. This adds `task_start` to the v1 set, bound on task documents via a nested start stage:

```yaml
hooks:
  pre: [lint]            # unchanged: binds around task_complete
  post: [approval_judge]
  start:                 # new: binds around task_start
    pre: [anchor]
    post: [announce]
```

Bare `pre`/`post` keep their terminal-verb meaning, so existing documents are unaffected.

## Semantics

- Fires inside `Manager.MarkInProgress`, which covers both start surfaces (`fest task in-progress` and `fest status set ... in_progress`) and any future caller.
- First start only: a task with a recorded start time never re-fires it, so resuming a blocked task or re-marking in_progress is not a new start.
- A `fail: closed` pre failure blocks the verb and leaves the task unstarted (no status change, no started event).
- Post failures are audit-only via `wf_hook_run`, matching goal-completion post semantics; the task stays started.
- Human-required steps skip start hooks with `skipped: human-gate` audit records, same as other verbs.
- Numeric progress updates (`fest task update --progress N`) continue to bypass lifecycle hooks, the same rule the terminal verbs already follow.
- `fest validate` warns on undeclared start bindings and warns when `hooks.start` appears on a goal document, where task_start never fires.

## Motivation

Agent campaigns need a lifecycle injection point before work begins (anchor commands, context setup). The gap surfaced while designing commit-trailer anchoring in an agent-economy campaign workitem; the interim workaround there is an out-of-band anchor command, which this verb replaces.

## Testing

- `just test unit`: 2814 passed
- `just test integration`: 77 passed (containerized suite)
- `just lint all`: 0 issues

New coverage: blocked pre leaves task unstarted with no started event, pre+post run around the transition in binding order, no re-fire on second start, human-gate skips with audit records, failed closed post keeps the task started, completion-stage bindings never run at start, absent start stage stays absent through frontmatter inject round-trips, validator warnings for undeclared names and goal-document placement.

### DIFF
--- a/internal/progress/events.go
+++ b/internal/progress/events.go
@@ -198,6 +198,12 @@ func materializeState(events []ProgressEvent) map[string]*TaskProgress {
 		case EventCompleted:
 			task.Status = StatusCompleted
 			ts := e.Timestamp
+			// A completion can be the task's first work signal (for example,
+			// UpdateProgress(100) or a direct MarkComplete). Mirror the live
+			// mutation so replay does not re-arm task_start hooks.
+			if task.StartedAt == nil {
+				task.StartedAt = &ts
+			}
 			task.CompletedAt = &ts
 			task.Progress = 100
 			task.TimeSpentMinutes = e.Minutes
--- a/internal/progress/events.go
+++ b/internal/progress/events.go
@@ -210,6 +210,13 @@ func materializeState(events []ProgressEvent) map[string]*TaskProgress {
 			if e.Percent > 0 && task.Status == StatusPending {
 				task.Status = StatusInProgress
 			}
+			// Mirror the live UpdateProgress mutation so the task_start
+			// first-start predicate (StartedAt == nil) is stable across
+			// process restarts.
+			if e.Percent > 0 && task.StartedAt == nil {
+				ts := e.Timestamp
+				task.StartedAt = &ts
+			}
 
 		case EventBlocked:
 			task.Status = StatusBlocked
--- a/internal/progress/events.go
+++ b/internal/progress/events.go
@@ -107,7 +107,7 @@ type ProgressEvent struct {
 	HookName     string `json:"hook_name,omitempty"`
 	HookLayer    string `json:"hook_layer,omitempty"`  // machine|festivals|festival
 	HookTiming   string `json:"hook_timing,omitempty"` // pre|post
-	HookVerb     string `json:"hook_verb,omitempty"`   // task_complete|sequence_complete|phase_complete|gate_approve
+	HookVerb     string `json:"hook_verb,omitempty"`   // task_start|task_complete|sequence_complete|phase_complete|gate_approve
 	HookOutcome  string `json:"hook_outcome,omitempty"`
 	HookSkip     string `json:"hook_skip,omitempty"`
 	HookExitCode int    `json:"hook_exit_code,omitempty"`
--- a/docs/concepts/hooks.md
+++ b/docs/concepts/hooks.md
@@ -141,7 +141,7 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 
 | Verb | When |
 | --- | --- |
-| `task_start` | first transition into `in_progress` (`fest task in-progress`, `fest status set ... in_progress`) |
+| `task_start` | first transition into work (`fest task in-progress`, `fest status set ... in_progress`, or the first `--progress` above zero) |
 | `task_complete` | `fest task completed` |
 | `sequence_complete` | sequence completion |
 | `phase_complete` | phase completion |
@@ -150,11 +150,15 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 Timings: **pre** (before the verb applies) and **post** (after the transition
 has already been applied).
 
-`task_start` fires only on the **first** start (no recorded start time yet):
-resuming a blocked task or re-marking an in-progress task never re-fires it. A
-`fail: closed` pre failure leaves the task unstarted. Numeric progress updates
-(`fest task update --progress N`) flip status without firing lifecycle hooks —
-the same rule terminal verbs already follow.
+`task_start` fires on the **first** transition into work, whichever surface
+causes it: `fest task in-progress`, `fest status set ... in_progress`, or the
+first `fest task update --progress N` above zero. Resuming a blocked task or
+re-marking an in-progress task never re-fires it; `fest task reset` clears the
+recorded start, so a reset task re-anchors on its next start. A `fail: closed`
+pre failure blocks the transition and leaves the task untouched. A `fail:
+closed` post failure keeps the task started and is recorded in the audit trail
+plus a stderr warning. Completion by progress (`--progress 100`) still bypasses
+`task_complete` hooks; only the terminal completion surfaces run those.
 
 Orchestration:
 
--- a/docs/concepts/hooks.md
+++ b/docs/concepts/hooks.md
@@ -1,8 +1,8 @@
 # Lifecycle Hooks
 
 fest hooks are **named commands** bound to festival lifecycle events (task
-complete, sequence complete, phase complete, gate approve). The approval judge
-is one hook among them, not a special case.
+start, task complete, sequence complete, phase complete, gate approve). The
+approval judge is one hook among them, not a special case.
 
 Related:
 
@@ -116,8 +116,16 @@ fest_type: task
 hooks:
   pre: [lint]
   post: [approval_judge, notify]
+  start: # task documents only; runs around task_start
+    pre: [anchor]
+    post: [announce]
 ---
 ```
+
+Bare `pre`/`post` bind around the step's **terminal** verb (`task_complete`,
+`sequence_complete`, `phase_complete`, `gate_approve`). The nested `start:`
+stage binds around **`task_start`** and is only honored on task documents;
+`fest validate` warns when it appears on a goal document.
 
 ### GATES.md body markers
 
@@ -133,6 +141,7 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 
 | Verb | When |
 | --- | --- |
+| `task_start` | first transition into `in_progress` (`fest task in-progress`, `fest status set ... in_progress`) |
 | `task_complete` | `fest task completed` |
 | `sequence_complete` | sequence completion |
 | `phase_complete` | phase completion |
@@ -140,6 +149,12 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 
 Timings: **pre** (before the verb applies) and **post** (after the transition
 has already been applied).
+
+`task_start` fires only on the **first** start (no recorded start time yet):
+resuming a blocked task or re-marking an in-progress task never re-fires it. A
+`fail: closed` pre failure leaves the task unstarted. Numeric progress updates
+(`fest task update --progress N`) flip status without firing lifecycle hooks —
+the same rule terminal verbs already follow.
 
 Orchestration:
 
--- a/docs/concepts/hooks.md
+++ b/docs/concepts/hooks.md
@@ -141,7 +141,7 @@ Bindings only reference names. They never set `command`, `fail`, or `timeout`.
 
 | Verb | When |
 | --- | --- |
-| `task_start` | first transition into work (`fest task in-progress`, `fest status set ... in_progress`, or the first `--progress` above zero) |
+| `task_start` | first transition into work (`fest task in-progress`, a direct completion, or the first `--progress` above zero) |
 | `task_complete` | `fest task completed` |
 | `sequence_complete` | sequence completion |
 | `phase_complete` | phase completion |
@@ -151,10 +151,12 @@ Timings: **pre** (before the verb applies) and **post** (after the transition
 has already been applied).
 
 `task_start` fires on the **first** transition into work, whichever surface
-causes it: `fest task in-progress`, `fest status set ... in_progress`, or the
-first `fest task update --progress N` above zero. Resuming a blocked task or
-re-marking an in-progress task never re-fires it; `fest task reset` clears the
-recorded start, so a reset task re-anchors on its next start. A `fail: closed`
+causes it: `fest task in-progress`, `fest status set ... in_progress`, a direct
+completion, or the first `fest task update --progress N` above zero. Reporting
+a blocker before work begins leaves the start hook armed. Resuming a task that
+was already started or re-marking it in progress never re-fires the hook;
+`fest task reset` clears the recorded start, so a reset task re-anchors on its
+next start. A `fail: closed`
 pre failure blocks the transition and leaves the task untouched. A `fail:
 closed` post failure keeps the task started and is recorded in the audit trail
 plus a stderr warning. Completion by progress (`--progress 100`) still bypasses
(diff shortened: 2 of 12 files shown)


obey-agent (commented):
## Summary

Hostile review of `task_start` + nested `hooks.start` bindings.

I attacked whether first-start gating is real, whether fail-closed pre leaves the task unstarted, whether progress updates / re-entry bypass hooks, YAML round-trip of absent `start:`, and whether validate placement warnings match the runtime.

**Stated goal holds on the primary start surfaces:** `MarkInProgress` (used by `fest task in-progress` and `fest status set … in_progress`) plans start-stage bindings, runs pre before mutation, blocks without status/`started` event, runs post after apply, and does not re-fire when `StartedAt` is already set. Tests cover the break sequences I care about. Bare `pre`/`post` stay terminal-only; existing documents are unaffected.

## Issue counts
- bugs (blocking): 0
- residual risks: 3

## Attack results

| Claim | Result |
|---|---|
| First start only (`StartedAt == nil`) | Holds; second `MarkInProgress` does not re-exec |
| fail:closed pre leaves unstarted | Holds; no `started` event; blocked audit present |
| Post fail is non-blocking for the verb | Holds; `postBlocked` is not treated as verb failure (aligned with complete CLI’s “task remains completed”) |
| Human-gate skips with audit | Holds via nil-safe `PlanBindings` + `ApplyHumanGate` |
| Goal-doc `hooks.start` warned | Holds (`hooks_start_binding_placement`) |
| Absent `start` not injected | Holds (frontmatter test) |

## Residual risks (non-blocking but product-real)

### 1. `UpdateProgress` permanently silences `task_start`
`Manager.UpdateProgress` still sets `StartedAt` and can flip `pending → in_progress` **without** calling `MarkInProgress` or any lifecycle hooks (documented, same as complete-via-progress). Break for the PR’s own motivation: an agent that does `fest task update --progress 10` before ever calling in-progress will **never** run start anchors — `StartedAt` is already set, so later `MarkInProgress` is a no-op for hooks. If agent campaigns are the customer, this is the main footgun; consider either firing start on first progress flip, or documenting it loudly in the agent-facing guide (not only hooks.md).

### 2. Post fail:closed is quieter than complete
Complete CLI prints a stderr warning when post is blocked; start path inside `MarkInProgress` only records `wf_hook_run`. Operators grepping CLI output will miss failed start post hooks.

### 3. Hook wiring asymmetry vs complete
Complete hooks live in the command layer around `MarkComplete`; start hooks live inside the manager. Any future non-CLI `MarkComplete` caller already skips complete hooks — start is more consistently enforced. Fine, but two patterns increase drift risk.

## Verdict

**COMMENT** — `task_start` is correctly gated and fail-closed on pre for the two public start verbs; no merge-blocking correctness bug found on `438c935`. Residual #1 is the one I would not ignore if the goal is agent pre-work anchoring.

obey-agent (commented):
## Summary

Hostile review of `task_start` + nested `hooks.start` bindings. Attacked: (1) does first-start gating actually prevent re-fire, (2) does fail-closed pre leave the task unstarted with no `started` event, (3) can completion-stage `pre`/`post` accidentally run at start, (4) are start bindings skippable via other status transitions, (5) does the schema round-trip materialize empty `start:` stages.

## Attack results

### Hold: first-start + fail-closed pre
`MarkInProgress` keys first-start on `StartedAt == nil`, plans start-stage bindings only then, and short-circuits before status/event mutation when pre is blocked. Unit tests cover blocked pre (no `started` event), pre→post order, re-entry without re-fire, human-gate skip, and bare `pre` not running at start. That core path looks correct.

### Hold: schema / validator
`StepHooks.Start` as a pointer + `StartBindings()` preserves absent stages through inject (tested). Validator warns on goal-doc placement and still scans start bindings for undeclared names. Good.

### Residual: start hooks are not a hard gate
Docs claim numeric progress updates flip status without lifecycle hooks. That is true in code: `UpdateProgress` sets `StartedAt` + `StatusInProgress` without calling `MarkInProgress`, so fail-closed `hooks.start.pre` anchors are skippable via `fest task update --progress N` (and any other caller of `UpdateProgress`). Same class of bypass terminal completion already had for ungated paths — but if `task_start` is sold as an enforcement point (anchor before work), this is a hole. Document more loudly or route first-progress through the same hooked transition.

### Residual: post fail-closed does not fail the command
`RunLifecycleStage` returns `blocked` for fail-closed post with `err == nil`. `MarkInProgress` only checks `postErr`, so a failed closed start post leaves the task started and returns success (test encodes this). Docs say the caller "should surface the failure"; events do, CLI success does not. Matches goal-completion post handling, but agents will treat the transition as clean.

### Residual: hooks live in Manager vs CLI asymmetry
`task_complete` hooks live in `fest task completed` (CLI). `task_start` hooks live in `Manager.MarkInProgress`. Better for `status set` / `in-progress` coverage, but any future Manager path that mutates start state without `MarkInProgress` silently skips them (already true for `UpdateProgress`).

## Verdict

**COMMENT** on head `438c9356` — feature is coherent and well-tested; no merge-blocking defect found. Residuals above should not be rubber-stamped away if start hooks are meant as policy, not just notification.

## Claims attacked vs held
| Claim | Result |
| --- | --- |
| first start only | held |
| blocked pre leaves unstarted | held |
| bare pre/post unchanged | held |
| all in_progress surfaces fire hooks | partial — MarkInProgress yes; UpdateProgress no |

lancekrogers:
Addressed the review residuals in 36888b5:

**Residual 1 (progress update permanently disarms task_start): fixed, not documented around.** The start stage now runs around the first transition into work on every surface, including the first `fest task update --progress N` above zero. A blocked pre leaves the task fully untouched (no progress event, no status change, no recorded start). Two supporting changes make the first-start predicate sound: `StartedAt` is only stamped when progress is above zero (a zero update, e.g. `fest status set pending`, no longer records a phantom start), and `EventProgress` replay now mirrors the live mutation so the predicate is stable across process restarts. New tests cover first-progress firing, blocked pre leaving the task untouched, no re-fire from later progress/in-progress calls, zero progress keeping the anchor armed, and replay fidelity.

**Residual 2 (quiet post failure): fixed.** A fail-closed `task_start` post failure now warns on stderr with the same wording pattern as the completion path; the task stays started and the failure remains in `wf_hook_run`.

**Residual 3 (wiring asymmetry): captured as follow-up, not in this PR.** Moving `task_complete` hooks from the command layer into the manager (which would also close the `fest status set ... completed` and `--progress 100` completion-hook bypass) is filed as campaign intent `fest-unify-taskcomplete-hook-wiring-20260812-160123`. Doing it here would widen a verb-addition PR into a completion-semantics change.

Gates on 36888b5: 2818 unit tests passed, 77 containerized integration tests passed, lint clean.

obey-agent (changes requested):
## Re-review on head `36888b5`

Prior COMMENT residual #1 (progress updates permanently disarm `task_start`) and #2 (silent fail-closed start post) are **actually attacked** in this commit: `UpdateProgress` runs the shared `beginTaskStartHooks` / `finishTaskStartHooks` stage on first progress > 0; post fail-closed now stderr-warns like complete. Unit tests cover first progress, blocked pre, zero progress, and re-fire protection.

That is real progress. One new hole remains around the restart predicate the PR itself claims to make “stable across process restarts.”

## Issue counts
- bugs: 1 (blocking)
- residuals closed from prior COMMENT: 2

## Findings

### [bug] First work as `--progress 100` leaves `StartedAt` nil after event replay

Live `UpdateProgress` on first `progress > 0` (including **100**) sets `StartedAt` and, when progress is 100, queues only `EventCompleted` — **not** `EventProgress`.

`materializeState` was only taught to stamp `StartedAt` on `EventProgress` with `Percent > 0`. `EventCompleted` still never sets `StartedAt`:

```go
case EventCompleted:
    task.Status = StatusCompleted
    // CompletedAt, Progress=100 — no StartedAt
```

Break sequence:
1. Fresh task, agent does `fest task update --progress 100` (or any path that first-starts via 100).
2. Live state: completed + `StartedAt` set; audit has `completed` only.
3. Process restart → rematerialize → **completed with `StartedAt == nil`**.
4. `UpdateProgress` / `MarkInProgress` first-start predicate is `StartedAt == nil` only → **`task_start` pre/post fire again** on a task that already completed (or on a later progress tweak).

The new `TestMaterializeState_ProgressEventSetsStartedAt` only covers `EventProgress` percent 10/0 — it greenwashes the jump-to-complete path the live code explicitly allows.

**Required fix:** on `EventCompleted` (and likely `EventStarted` already covers in-progress), if `StartedAt == nil`, set it from the event timestamp (or preserve a started stamp when emitting completed-from-progress). Add a materialize test: single `EventCompleted` with no prior progress → `StartedAt != nil`. Prefer also asserting that after rematerialize, `UpdateProgress(10)` / `MarkInProgress` do not re-run start hooks.

## What holds

| Claim | Result |
|---|---|
| First `--progress N>0` fires start hooks | Holds (live path + tests) |
| Blocked start pre leaves progress untouched | Holds |
| Zero progress keeps anchor armed | Holds |
| Fail-closed start post warns on stderr | Holds |
| Restart-stable first-start predicate | **Fails** for first-start-via-100 |

## Verdict

**REQUEST_CHANGES** — progress-first start is the right product fix, but the restart mirror is incomplete for the 100-as-first-signal path. Close the materialize gap before merge.

internal/progress/events.go
```diff
@@ -210,6 +210,13 @@ func materializeState(events []ProgressEvent) map[string]*TaskProgress {
 			if e.Percent > 0 && task.Status == StatusPending {
 				task.Status = StatusInProgress
 			}
+			// Mirror the live UpdateProgress mutation so the task_start
+			// first-start predicate (StartedAt == nil) is stable across
+			// process restarts.
```
obey-agent, line 221: This only mirrors `EventProgress`. A lone `EventCompleted` (first work as progress 100, or MarkComplete) still leaves `StartedAt` nil after rematerialize, so `task_start` re-arms on the next process. Stamp `StartedAt` from `EventCompleted` when nil (same timestamp), or emit a paired start event on the live 100-as-first path. Test: single completed event → StartedAt != nil; MarkInProgress after rematerialize does not re-run start hooks.

obey-agent (changes requested):
## Adversarial re-check on unchanged head `36888b5`

Prior REQUEST_CHANGES (EventCompleted rematerialize) is **still unfixed**. Spot-check found a second start-predicate hole in the same class the PR claims to close for progress updates.

## Issue counts
- bugs: 2 (blocking)
- prior finding status: unfixed

## Findings

### [bug 1 — still open] `EventCompleted` never stamps `StartedAt` on rematerialize

Unchanged from prior review. First work as `--progress 100` / MarkComplete queues only `EventCompleted`. Live path sets `StartedAt`; restart loses it; `StartedAt == nil` re-arms `task_start`.

### [bug 2 — new] `ReportBlocker` (and `MarkComplete`) set `StartedAt` without running `task_start`

The PR’s own rationale for teaching `UpdateProgress` to run start hooks:

> without it a progress update would permanently disarm start anchors (StartedAt set, hooks never fired)

That exact pattern remains on other mutations:

```go
// ReportBlocker
if task.StartedAt == nil {
    task.StartedAt = &now  // time tracking — no beginTaskStartHooks
}
```

**Break sequence:**
1. Task has `hooks.start.pre: [anchor]`.
2. Agent reports a blocker while still pending (`fest task blocked …`) before any start.
3. `ReportBlocker` stamps `StartedAt`, status=blocked, **no task_start**.
4. Blocker cleared; agent `fest task in-progress` / first progress.
5. First-start predicate is `StartedAt == nil` → **hooks skipped forever**.

`MarkComplete` on a never-started task has the same stamp-without-hooks shape (less common than blocker-first, same predicate poison).

The PR body claims start fires inside `MarkInProgress` covering start surfaces. That is true only if nothing else stamped `StartedAt` first. Once `StartedAt` is the exclusive arming bit, **every writer of StartedAt is a start surface** — progress was fixed; blocker/complete were not.

**Required:** either (a) run the same `beginTaskStartHooks`/`finishTaskStartHooks` around any first `StartedAt` write that means work is beginning, or (b) stop using `StartedAt` as the hook arming bit (separate `StartHooksFired` / only arm from explicit start transitions). At minimum: `ReportBlocker` must not set `StartedAt` on a still-unstarted pending task, or must run start hooks first (fail-closed pre blocks the blocker report). Add a unit test: pending → blocked → in_progress still fires `task_start` once.

## What still holds
- MarkInProgress / first progress>0 run start hooks live.
- Blocked start pre leaves task unstarted on those paths.

## Verdict
**REQUEST_CHANGES** — restart mirror still incomplete; `StartedAt` side-writers still disarm the verb this PR exists to provide.

internal/progress/events.go
```diff
@@ -198,6 +198,12 @@ func materializeState(events []ProgressEvent) map[string]*TaskProgress {
 		case EventCompleted:
 			task.Status = StatusCompleted
 			ts := e.Timestamp
+			// A completion can be the task's first work signal (for example,
+			// UpdateProgress(100) or a direct MarkComplete). Mirror the live
+			// mutation so replay does not re-arm task_start hooks.
+			if task.StartedAt == nil {
```
obey-agent, line 204: Closes the rematerialize hole: a lone completed event now preserves the first-start predicate. Paired with the live UpdateProgress(100) / MarkComplete paths that only emit EventCompleted.

obey-agent (commented):
## Re-review on head `2b794fd`

Prior REQUEST_CHANGES: (1) `EventCompleted` rematerialize left `StartedAt` nil → re-arm `task_start`; (2) `ReportBlocker`/`MarkComplete` stamped `StartedAt` without running start hooks.

## Prior findings — fixed

| Finding | Status |
|---|---|
| EventCompleted rematerialize | **Fixed** — stamps `StartedAt` when nil; `TestMaterializeState_CompletedEventSetsStartedAt` + `TestUpdateProgress_CompleteReplayDoesNotRefireStartHooks` |
| ReportBlocker disarms start | **Fixed** — no longer writes `StartedAt`; `TestReportBlockerBeforeStartKeepsStartHooksArmed` |
| MarkComplete first-work skip hooks | **Fixed** — `begin`/`finish` start stage around complete; blocked pre leaves uncompleted; unit coverage |

## Attack results

- Live first-work as `--progress 100` → only `EventCompleted` in audit → reload → `StartedAt` set → `MarkInProgress` does not re-fire. Holds.
- pending → blocked → clear → in_progress: start stays armed until real start surface; ClearBlocker does not stamp `StartedAt`. Holds.
- MarkComplete fail-closed start pre: no completed event, no `StartedAt`. Holds.

## Residual risks (non-blocking)

1. **ClearBlocker can set status `in_progress` without firing `task_start`** when the task was never started (blocker-first). Next real start surface still fires hooks because `StartedAt` is nil — OK — but frontmatter can say in_progress before anchors run.
2. **Post fail-closed start after MarkComplete** — completion is already saved; post failure only warns (same as MarkInProgress post semantics).
3. **Docs still say progress-100 bypasses `task_complete`** — orthogonal; start hooks correctly fire on the first-work path.

## Verdict

**COMMENT** — both production blockers from the prior CR set are actually fixed on `2b794fd` with tests that hit the break scenarios. No remaining merge-blocking gap against the stated `task_start` arming/disarming contract.

PR #327, opened 2026-08-12
fix: commit festival work in linked projects

lancekrogers:
## Summary

- route plain `fest commit` from a sequence through `fest_working_dir`, then fall back to the festival navigation link and legacy `project_path`
- create the project commit first and the scoped campaign-root commit second with the same FE reference
- skip the project commit when the linked repository is clean, allowing festival-only evidence and close-out commits to proceed at the campaign root
- retain `no changes to commit` when neither the linked project nor festival scope changed
- resolve `--festival` paths, names, and IDs before commit routing so absolute paths no longer degrade to campaign-only tags
- document the behavior and add container regressions for all entry paths

## Root cause

`fest commit` selected its primary repository solely from the caller's current Git root. Running it from a festival therefore committed only campaign files, while changing into the linked project discarded sequence-level targeting context. Separately, `--festival /absolute/path` deferred scope resolution but the tag resolver only searched festival names and IDs, silently dropping the FE reference and skipping the root follow-up commit.

After project routing was added, a clean linked repository exposed a second edge case: the project-side `no changes` result stopped execution before festival-only files could be committed at the campaign root. The command now treats a clean linked project as a skipped optional commit while preserving a real no-op error when both scopes are clean.

## Validation

- `go test ./...`
- `go vet ./...`
- `just lint all`
- focused container regressions for sequence, navigation-link, absolute-path, festival-only, and true no-op workflows
- complete container suite: `FULL_INTEGRATION_PASS`
- `just install stable` into an isolated `GOBIN`; installed macOS binary reported commit `246a5c4`
- installed Linux binary at `/usr/local/bin/fest`; festival-only linked-project smoke test produced `[FE-DO0001] fest: docs: installed festival-only evidence`, left the project commit unchanged, and left both repositories clean

### DIFF
--- a/internal/commands/commit/commit.go
+++ b/internal/commands/commit/commit.go
@@ -16,7 +16,10 @@ import (
 	"github.com/Obedience-Corp/fest/internal/config"
 	"github.com/Obedience-Corp/fest/internal/errors"
 	"github.com/Obedience-Corp/fest/internal/frontmatter"
+	"github.com/Obedience-Corp/fest/internal/guidance/selection"
 	"github.com/Obedience-Corp/fest/internal/id"
+	"github.com/Obedience-Corp/fest/internal/navigation"
+	"github.com/Obedience-Corp/fest/internal/pathutil"
 	"github.com/Obedience-Corp/fest/internal/scope"
 	"github.com/Obedience-Corp/fest/internal/ui"
 	"github.com/spf13/cobra"
@@ -60,13 +63,15 @@ The fest commit command wraps git commit and automatically:
   1. Stages changes and prepends the festival reference to the commit message
   2. Creates a campaign root commit for festival-scoped files (task docs, progress, state)
 
-When run from a linked project directory, two commits are created:
+When a festival or sequence has a linked project, two commits are created even
+when this command is run from inside the festival:
   - Project commit: stages all project changes
   - Campaign root commit: stages only festival directory, .campaign/fest/,
     festivals/.festival/.state/, and the submodule pointer
 
-When run from a festival directory, one commit is created at the campaign root
-with only festival-scoped files staged (not git add -A).
+The sequence's fest_working_dir is preferred over the festival navigation link
+and legacy fest.yaml project_path. A festival with no linked project creates one
+campaign-root commit containing only festival-scoped files (not git add -A).
 
 Use --no-root to skip the campaign root commit.
 
@@ -77,13 +82,13 @@ Reference format: [FE-{id}]
 Detection priority:
   1. Explicit --task flag value
   2. Task fest_ref from current directory (if inside festival task)
-  3. Festival ID from fest.yaml metadata
-  4. Explicit --festival flag (name or ID)
+  3. Explicit --festival flag (path, name, or ID)
+  4. Festival ID from fest.yaml metadata
 
 Examples:
   fest commit -m "Implement feature"
-  # In linked project → [FE-CS0001] Implement feature
-  # In festival task  → [FE-FEST-a3b2c1] Implement feature
+  # In linked project or sequence → [FE-CS0001] Implement feature
+  # In festival task              → [FE-FEST-a3b2c1] Implement feature
 
   fest commit --task FEST-b4c5d6 -m "Related work"
   # → [FE-FEST-b4c5d6] Related work
@@ -104,7 +109,7 @@ Examples:
 
 	cmd.Flags().StringVarP(&message, "message", "m", "", "commit message (required unless --auto-write)")
 	cmd.Flags().StringVar(&taskRef, "task", "", "task reference ID to use (e.g., FEST-a3b2c1)")
-	cmd.Flags().StringVar(&festivalFlag, "festival", "", "festival name or ID (overrides auto-detection)")
+	cmd.Flags().StringVar(&festivalFlag, "festival", "", "festival path, name, or ID (overrides auto-detection)")
 	cmd.Flags().BoolVar(&noTag, "no-tag", false, "don't prepend task reference")
 	cmd.Flags().BoolVar(&jsonOut, "json", false, "output result as JSON")
 	cmd.Flags().BoolVar(&autoStage, "stage", true, "auto-stage all changes before commit")
@@ -148,14 +153,19 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	// Resolve workspace for campaign integration (nil-safe: ok if absent).
 	ws, _ := scope.WorkspaceFrom(ctx)
 
-	// Determine if we're in a project submodule vs campaign root.
-	var inSubmodule bool
-	var submoduleRelPath string
-	if ws != nil && ws.Type == scope.WorkspaceTypeCampaign {
-		inSubmodule, submoduleRelPath = isInSubmodule(ws.Root)
-	}
-
 	festivalPath, hasFestival := scope.FestivalFrom(ctx)
+	if festivalFlag != "" {
+		var resolveErr error
+		festivalPath, resolveErr = resolveFestivalFlagPath(ctx, ws, festivalFlag)
+		if resolveErr != nil {
+			result.Success = false
+			result.Error = resolveErr.Error()
+			return outputResult(result)
+		}
+		hasFestival = true
+		ctx = scope.WithFestival(ctx, festivalPath)
+		cmd.SetContext(ctx)
+	}
 
 	// Check if we're in a git repository
 	inRepo, err := isGitRepo(ctx)
@@ -177,6 +187,22 @@ func runCommit(cmd *cobra.Command, args []string) error {
 		return outputResult(result)
 	}
 
+	// A sequence's working directory and a festival's project link are explicit
+	// commit targets. Resolve them while the caller is still in the festival so
+	// the command can preserve festival context without requiring a manual cd.
+	projectCommit := false
+	var submoduleRelPath string
+	if ws != nil && ws.Type == scope.WorkspaceTypeCampaign {
+		primaryRepoPath, projectCommit, submoduleRelPath, err = resolveCommitTarget(
+			ctx, ws.Root, primaryRepoPath, festivalPath, hasFestival,
+		)
+		if err != nil {
+			result.Success = false
+			result.Error = err.Error()
+			return outputResult(result)
+		}
+	}
+
 	// Guard outcomes and refusals go to the command's stderr so --json stdout
 	// stays a pure document.
 	report := cmd.ErrOrStderr()
@@ -184,8 +210,8 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	// Auto-stage changes if enabled (default: true).
 	// When at campaign root (not in submodule), stage only festival-scoped paths.
 	if autoStage {
-		if inSubmodule {
-			// In submodule: stage all changes in the project repo
+		if projectCommit {
+			// In a project: stage all changes in the project repo.
 			if err := stageAllChanges(ctx, primaryRepoPath, report); err != nil {
 				result.Success = false
 				result.Error = err.Error()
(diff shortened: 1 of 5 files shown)


obey-agent (commented):
## Summary

Hostile review of linked-project `fest commit` routing (`fest_working_dir` → nav link → `project_path`) and path-capable `--festival`.

I attacked wrong-repo selection from inside a festival, absolute `--festival` degrading to campaign-only tags, submodule pointer updates, and fail-open when a declared link is broken.

**Stated root-cause fix holds:** when the caller is at campaign root under a festival with a declared project target, `resolveCommitTarget` redirects primary commit to that repo, sets `projectCommit`, and follows with a scoped root commit including the gitlink when tracked. Absolute `--festival` is resolved via `validateFestivalFlagPath` / name-ID search before routing. Container regressions cover sequence working-dir, navigation link, and absolute path entry points.

## Issue counts
- bugs (blocking): 0
- residual risks: 3

## Attack results

| Claim | Result |
|---|---|
| Sequence `fest_working_dir` preferred | Holds — nearest sequence from cwd under festival (`parts[0]/parts[1]`) |
| Nav link / legacy `project_path` fallback | Holds |
| Invalid declared link fails loud | Holds (`declared=true` + resolve/stat errors) |
| Already-in-project stays put | Holds — no silent retarget to a different link |
| Absolute `--festival` keeps FE tag + root commit | Covered by integration test |
| Non-submodule targets skip gitlink staging | Intentional (`gitlinkRelPath` requires mode 160000) |

## Residual risks (non-blocking)

### 1. Partial success if root commit fails after project commit
Project commit lands first; root failure is a warning path (`campaign root commit skipped`). Operator can end with project advanced and festival/gitlink not. Pre-existing multi-commit shape; still the worst partial-success mode of this command.

### 2. Sequence selection is path-depth based only
`fest_working_dir` is read only when cwd is at least two levels under the festival. From a phase directory (`001_IMPLEMENT/`) alone, sequence working dir is skipped and nav/legacy wins — may surprise if multiple sequences under one phase have different projects.

### 3. Nav link key is `filepath.Base(festivalPath)`
Relies on navigation map keys matching the festival directory basename. Consistent with `fest link` usage in the new tests; a hand-edited nav file with mismatched keys silently falls through to `project_path`/none.

## Verdict

**COMMENT** — linked-project commit from festival/sequence cwd and path-based `--festival` meet the claimed fix on `7c7a8a01`; integration tests exercise the three break sequences from the PR body. No merge-blocking correctness bug found.

obey-agent (dismissed):
## Summary

Hostile review of dual-repo `fest commit` routing (`fest_working_dir` → navigation link → `project_path`). Attacked: (1) clean linked project with festival-only dirty files, (2) absolute `--festival` path tag + root commit, (3) non-submodule / external link pointer staging, (4) cwd already inside a different project than the sequence working dir, (5) invalid declared links fail open vs fail closed.

## Attack results

### REQUEST_CHANGES: clean project blocks festival-only commits

When the caller is in the campaign/festival tree and a project is linked, `resolveCommitTarget` sets `projectCommit=true` and redirects the **primary** commit to the project repo.

Primary path:
1. `stageAllChanges(project)`
2. `commitWithCampaignSupport` → `commitkit.Commit` → **`ErrNoChanges` becomes hard error `"no changes to commit"`**
3. Campaign-root commit (`commitFestivalAtRoot`) only runs **after** primary success

Asymmetry: `commitCampaignRoot` treats no staged changes as a **silent skip**, but the primary project commit does **not**. So the common case:

- festival/task docs or `.fest/` dirty
- linked project working tree clean
- `fest commit -m ...` from the sequence/festival

…now fails entirely and never reaches the campaign-root festival commit. **Before this PR**, that case was the normal single campaign-root commit. The PR body sells "two commits when linked"; it does not acknowledge destroying the festival-only path.

Integration tests only cover dirty project files. No regression for clean-project + dirty festival.

**Required fix:** if the project has nothing to commit, fall through to campaign-root-only (or treat project `ErrNoChanges` as non-fatal and still attempt `commitFestivalAtRoot`), instead of aborting the whole command.

### Hold: declared-link fail-closed
Invalid `fest_working_dir` / missing target / non-dir / non-git target returns error with `declared=true` rather than quietly committing the campaign root. Good.

### Hold: absolute `--festival` path
`resolveFestivalFlagPath` at command entry + `validateFestivalFlagPath` (requires loadable fest id) fixes the prior path→campaign-only-tag degradation. Covered by `TestCommitFromProjectResolvesAbsoluteFestivalPath`.

### Hold: gitlink only when mode 160000
`gitlinkRelPath` avoids staging non-submodule worktrees as gitlinks. Project commit still happens; root omits pointer. Correct.

### Residual: already-in-project short-circuit
If cwd is inside *any* non-campaign git root, `resolveCommitTarget` keeps that repo and **never** consults `fest_working_dir`. A sequence linked to `projects/A` while the agent sits in `projects/B` commits B. Documented as intentional ("caller already inside a project stays there") but easy to get wrong with multi-project festivals.

### Residual: sequence detection needs ≥2 path segments under festival
Cwd at phase root (one segment) skips `fest_working_dir` and falls through to festival navigation / `project_path`. Fine for single-project festivals; multi-sequence multi-workdir festivals lose sequence preference when not cwd'd into the sequence.

## Verdict

**REQUEST_CHANGES** on head `7c7a8a01` — the clean-project primary-commit hard-fail is a real regression for festival-scoped commits after linking a project. Fix that path (and add a container test) before merge.

## Claims attacked vs held
| Claim | Result |
| --- | --- |
| fest commit from sequence hits working dir | held when project dirty |
| two commits even from festival | **fails when project clean** |
| absolute --festival preserves FE tag + root | held |
| invalid link fails loudly | held |

obey-agent (commented):
## Re-review on head `246a5c4`

Prior REQUEST_CHANGES: clean linked project made primary `commitkit.Commit` hard-fail with `no changes to commit`, so festival-only dirty trees never reached `commitFestivalAtRoot`.

## Prior finding — fixed

After staging, `HasStagedChanges` gates the project primary commit. Clean project → skip primary, still run scoped root commit (or hard-fail only when root also has nothing / `--no-root`). Container test `TestCommitFromLinkedFestivalAllowsFestivalOnlyChanges` locks festival-only success and dual-clean failure.

I re-attacked order: stage project first → check staged → optional primary → root. Festival paths are not left unstaged on the clean-project path (root path stages them). Docs now say “up to two commits” / “skipped when the project is clean.”

## Residual risks (non-blocking)

1. **Already-in-project short-circuit** still ignores `fest_working_dir` when cwd is any non-campaign git root.
2. **Sequence detection** still needs ≥2 path segments under the festival.
3. **Partial success** when project commits and root fails remains a warning (only when primary committed).

## Verdict

**COMMENT** — clean-project regression is closed on `246a5c4` with an integration lock. No remaining merge-blocking correctness bug found.

internal/commands/commit/commit.go
```diff
@@ -270,28 +296,59 @@ func runCommit(cmd *cobra.Command, args []string) error {
...
-	// When at the campaign root, the primary commit above already handled festival files.
-	if !noRoot && inSubmodule && ws != nil && ws.Type == scope.WorkspaceTypeCampaign && hasFestival {
+	// Campaign root commit: after a project commit, stage festival-scoped files
+	// and any tracked submodule pointer separately. Without a project target,
+	// the primary campaign-root commit above already handled festival files.
+	if noRoot && projectCommit && !primaryCommitted {
+		result.Error = "no changes to commit"
+		return outputResult(result)
+	}
+	if !noRoot && projectCommit && ws != nil && ws.Type == scope.WorkspaceTypeCampaign && hasFestival {
```
obey-agent, line 331: **Break:** `projectCommit=true` is set for any non-root campaign repo (`currentRepo != campaignRoot`), not only festival-linked work. When the project is clean and `hasFestival` is false, this entire root block is skipped.

Then execution falls through to `Success=true` with zero commits. On main, `commitWithCampaignSupport` would have returned `no changes to commit`.

Concrete: `cd projects/foo && fest commit -m wip` on a clean submodule without festival scope → exit success, empty result. That is a regression introduced by the skip gate.

internal/commands/commit/commit.go
```diff
@@ -270,28 +296,59 @@ func runCommit(cmd *cobra.Command, args []string) error {
...
+				result.Hash = rootHash
+				result.Message = festivalRootCommitMessage(result.CampaignTag, ref, rawMsg)
+			}
+		} else if !primaryCommitted {
+			result.Error = "no changes to commit"
+			return outputResult(result)
 		}
 	}
 
+	result.Success = true
```
obey-agent, line 355: Unconditional success after optional primary + optional root. When both are skipped (`!primaryCommitted` and root `if` false because `!hasFestival` or non-campaign), this reports success without a commit hash.

Fix: only set `Success` when something actually committed, or explicitly error `no changes to commit` when `!primaryCommitted` and the root path did not produce a hash.

obey-agent (changes requested):
## Adversarial re-review on head `246a5c4` (latent bug on previously COMMENT head)

Prior COMMENT closed the festival-only clean-project regression with `HasStagedChanges` gating. That path still works when `hasFestival` is true. A different gate gap fails open.

## Issue counts
- bugs: 1 (blocking)
- residuals from prior COMMENT: still open (partial root failure after primary success)

## Findings

### [bug] Clean project + no festival context reports Success with zero commits

The new skip path:

1. `projectCommit && !HasStagedChanges` → skip `commitWithCampaignSupport` (`primaryCommitted=false`)
2. Hard-fail `no changes` only if:
   - `--no-root`, **or**
   - root path runs (`hasFestival`) and root also empty/errors
3. Otherwise fall through to `result.Success = true`

**Break sequence (regression vs main):**
1. Campaign workspace, cwd inside a clean submodule (`projects/foo`).
2. No festival in scope (`hasFestival=false`) — common for ad-hoc `fest commit` from a project.
3. `resolveCommitTarget` still sets `projectCommit=true` because `currentRepo != campaignRoot`.
4. Stage finds nothing → skip primary → root `if` is false (needs `hasFestival`) → **Success=true, no hash, no error.**

On `main`, the same inputs always call `commitWithCampaignSupport` and return `no changes to commit`. The comment at the skip site claims *"All other modes retain the ordinary no-changes error from commitWithCampaignSupport"* — that is **false** for project-mode without festival context.

Integration coverage only locks festival-present dual-clean and festival-only dirty. Nothing locks clean-project / no-festival.

**Required:** after skipping primary, if the root path is not entered (or is disabled), fail closed with `no changes to commit` whenever nothing was committed — e.g. track `committedAnything` / always require primary-or-root success, not merely `Success=true` fallthrough. Add a container test: clean submodule, no festival scope, expect error.

## Residual (non-blocking, restated)
- Project commits then root fails → warn + Success (partial). Unchanged.

## Verdict
**REQUEST_CHANGES** — false success greenwashes a clean tree; breaks the PR's own "retain ordinary no-changes error" claim outside the festival-only happy path.

internal/commands/commit/commit.go
```diff
@@ -270,28 +296,63 @@ func runCommit(cmd *cobra.Command, args []string) error {
...
+			if !primaryCommitted {
+				result.Hash = rootHash
+				result.Message = festivalRootCommitMessage(result.CampaignTag, ref, rawMsg)
+			}
+		} else if !primaryCommitted {
+			result.Error = "no changes to commit"
+			return outputResult(result)
 		}
 	}
+	if !primaryCommitted && result.Hash == "" {
```
obey-agent, line 350: This closes the prior fail-open: clean project + no festival no longer falls through to `Success=true`. Residual (non-blocking): if `primaryCommitted` and root staging/commit fails, the code still warns and succeeds — intentional partial, but gates that assume dual-commit completeness should not treat that as green.

obey-agent (commented):
## Re-review on head `ec5d7b5`

Prior REQUEST_CHANGES: clean project + no festival → `Success=true` with zero commits.

## Prior finding — fixed

New fail-closed gate before success:

```go
if !primaryCommitted && result.Hash == "" {
    result.Error = "no changes to commit"
    return outputResult(result)
}
```

Break sequence re-run:
1. Campaign workspace, clean submodule cwd, `hasFestival=false` → `projectCommit=true`.
2. Skip primary (`primaryCommitted=false`), root `if` not entered.
3. Guard trips → error, no Success.

Unit test `TestRunCommit_CleanProjectWithoutFestivalReturnsNoChanges` locks the scenario (submodule under campaign workspace, no festival scope).

The comment at the skip site is now true for the no-festival project-clean path.

## Residual risks (non-blocking)

1. **Partial success when primary commits and root fails** — still warns and `Success=true`. Festival files / submodule pointer can lag the project commit. Pre-existing shape; not the false-empty success bug.
2. **Package-level flag globals in the unit test** — mutates `message`/`autoStage`/… without restore; fine under sequential package tests, fragile if the package goes parallel.
3. **`--json` + error** — still emits a document with `success:false` and process exit 0 via `outputResult` (pre-existing JSON contract).

## Verdict

**COMMENT** — prior false-success regression is actually closed with a targeted test. No remaining merge-blocking gap against the stated clean-tree fail-closed claim.

PR #328, opened 2026-08-13
fix(hooks): run task_complete bindings on every completion surface

lancekrogers:
## Summary

Completion hooks previously lived only in the `fest task completed` command layer, so `fest status set <task> completed` and programmatic `UpdateProgress(100)` completed tasks without ever consulting their bindings. This was flagged as residual risk 3 in the hostile review of #326, which wired `task_start` at the manager layer and left completion behind.

This change moves `task_complete` planning and execution into `progress.Manager`, mirroring the #326 `task_start` pattern:

- `Manager.MarkComplete` always runs the task document's bare `pre`/`post` bindings around the transition.
- `Manager.UpdateProgress` runs them when a 100% update actually transitions the task into completed.
- `fest task completed` now delegates to the manager and only renders the blocked-hook outcome. Its `--json` contract is unchanged (`success`/`task`/`blocked`/`failed_hook`/`message`).
- The start and completion stages share one implementation (`beginTaskHooks`/`finishTaskHooks`), replacing the start-only helpers.

## Semantics

- Pre ordering preserves the historical `fest task completed` flow exactly: `task_complete` pre, then first-start `task_start` pre, transition, `task_start` post, `task_complete` post.
- A blocked `task_complete` pre leaves the task untouched; no completed event is recorded.
- A fail-closed post failure keeps the task completed, recorded in the audit trail plus a stderr warning.
- Replaying `UpdateProgress(100)` on an already-completed task never re-fires completion hooks. Explicitly re-completing via `fest task completed` still re-fires them, matching its historical behavior.
- Behavior change: infrastructure errors from post hooks (runner failure, event save) now fail the command with the task already completed, matching `MarkInProgress` semantics. Previously `fest task completed` only warned on stderr and exited 0.

## Surfaces covered

- `fest task completed`: unchanged UX, hooks now run inside the manager.
- `fest status set <task> completed`: bypass closed; regression test drives the real handler with a real fail-closed hook.
- Programmatic `UpdateProgress(100)`: both CLI update paths already reject 100%, so this is defense in depth for library callers.

Quality gates remain command-layer and are out of scope here; `fest status set completed` still does not evaluate them. That asymmetry predates this change and is worth its own decision.

## Testing

- Full unit suite: 2833 tests pass. `go vet` and `golangci-lint` clean.
- New manager tests: hook order around the transition, blocked pre leaves task incomplete, fail-closed post keeps task completed, 100% transition fires, replay guard, mid-progress fires nothing.
- New status handler regression test for the closed bypass (blocked and passing variants).
- Command-layer hook tests now drive the real runner with real commands (`true`/`false`) since the command exec seam was removed with the code it faked.
- Containerized integration gate (`just gate`) run before opening this PR.

### DIFF
--- a/internal/progress/manager.go
+++ b/internal/progress/manager.go
@@ -167,18 +167,21 @@ func (m *Manager) UpdateProgress(ctx context.Context, taskID string, progress in
 		return err
 	}
 	m.SyncFrontmatterStatus(taskID, task.Status)
-	if err := m.finishTaskHooks(ctx, startStage, "started"); err != nil {
-		return err
-	}
-	if err := m.finishTaskHooks(ctx, completeStage, "completed"); err != nil {
-		return err
+	// The transition is applied; every remaining side effect must still be
+	// attempted when a post stage fails. A start post failure must not skip
+	// the task_complete post stage, and neither may skip parent propagation:
+	// otherwise a completed task leaves its sequence goal permanently stalled.
+	// The first post error is returned after all side effects have run.
+	postErr := m.finishTaskHooks(ctx, startStage, "started")
+	if err := m.finishTaskHooks(ctx, completeStage, "completed"); postErr == nil {
+		postErr = err
 	}
 	if progress == 100 {
 		// Propagation is best-effort: a failure here should not block
 		// the progress update that already succeeded above.
 		_ = m.PropagateCompletion(ctx, taskID)
 	}
-	return nil
+	return postErr
 }
 
 // MarkComplete marks a task as complete
@@ -247,11 +250,16 @@ func (m *Manager) MarkComplete(ctx context.Context, taskID string) error {
 		return err
 	}
 	m.SyncFrontmatterStatus(taskID, task.Status)
-	if err := m.finishTaskHooks(ctx, startStage, "started"); err != nil {
-		return err
-	}
-	if err := m.finishTaskHooks(ctx, completeStage, "completed"); err != nil {
-		return err
+	// The completion is applied; every remaining side effect must still be
+	// attempted when a post stage fails. A start post failure must not skip
+	// the task_complete post stage, and neither may skip the ledger emit or
+	// parent propagation: otherwise a completed task leaves its sequence goal
+	// permanently stalled. The first post error is returned after all side
+	// effects have run, preserving the historical contract that ledger and
+	// propagation always follow a successful completion mutation.
+	postErr := m.finishTaskHooks(ctx, startStage, "started")
+	if err := m.finishTaskHooks(ctx, completeStage, "completed"); postErr == nil {
+		postErr = err
 	}
 	// Campaign ledger: high-intent task completion (D003/D006). Does not
 	// alter progress_events.jsonl content beyond the save already done.
@@ -261,7 +269,7 @@ func (m *Manager) MarkComplete(ctx context.Context, taskID string) error {
 	// Propagation is best-effort: a failure here should not block
 	// the task completion that already succeeded above.
 	_ = m.PropagateCompletion(ctx, taskID)
-	return nil
+	return postErr
 }
 
 // MarkInProgress marks a task as in progress. On the first start (no
--- a/internal/progress/manager.go
+++ b/internal/progress/manager.go
@@ -87,13 +87,27 @@ func (m *Manager) UpdateProgress(ctx context.Context, taskID string, progress in
 		}
 	}
 
+	// A 100% update is a completion transition, so the task_complete verb
+	// runs here too: without it UpdateProgress(100) would complete a task
+	// with its completion bindings silently skipped. Hooks fire only on an
+	// actual transition into completed; replaying 100 on an already-completed
+	// task never re-fires them.
+	var completeStage *taskHookStage
+	if progress == 100 && task.Status != StatusCompleted {
+		var err error
+		completeStage, err = m.beginTaskHooks(ctx, taskID, hooks.VerbTaskComplete)
+		if err != nil {
+			return err
+		}
+	}
+
 	// The first progress above zero is the first start signal, so the
 	// task_start stage runs here too: without it a progress update would
 	// permanently disarm start anchors (StartedAt set, hooks never fired).
-	var stage *startHookStage
+	var startStage *taskHookStage
 	if task.StartedAt == nil && progress > 0 {
 		var err error
-		stage, err = m.beginTaskStartHooks(ctx, taskID)
+		startStage, err = m.beginTaskHooks(ctx, taskID, hooks.VerbTaskStart)
 		if err != nil {
 			return err
 		}
@@ -153,7 +167,10 @@ func (m *Manager) UpdateProgress(ctx context.Context, taskID string, progress in
 		return err
 	}
 	m.SyncFrontmatterStatus(taskID, task.Status)
-	if err := m.finishTaskStartHooks(ctx, stage); err != nil {
+	if err := m.finishTaskHooks(ctx, startStage, "started"); err != nil {
+		return err
+	}
+	if err := m.finishTaskHooks(ctx, completeStage, "completed"); err != nil {
 		return err
 	}
 	if progress == 100 {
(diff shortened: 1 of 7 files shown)


internal/progress/manager.go
```diff
@@ -219,7 +247,10 @@ func (m *Manager) MarkComplete(ctx context.Context, taskID string) error {
 		return err
 	}
 	m.SyncFrontmatterStatus(taskID, task.Status)
-	if err := m.finishTaskStartHooks(ctx, stage); err != nil {
+	if err := m.finishTaskHooks(ctx, startStage, "started"); err != nil {
```
obey-agent: If `finishTaskHooks(startStage)` returns an infrastructure error after the completion Save above, the function returns here: `task_complete` post never runs, and `emitLedger` / `PropagateCompletion` below never run. Historical `fest task completed` always reached complete post after a successful MarkComplete, and always propagated before that post. Move ledger+propagate (and preferably complete post) so a start-post infra failure cannot leave a completed task with no sequence propagation and no completion post.

internal/progress/manager.go
```diff
@@ -153,7 +167,10 @@ func (m *Manager) UpdateProgress(ctx context.Context, taskID string, progress in
...
+		return err
+	}
+	if err := m.finishTaskHooks(ctx, completeStage, "completed"); err != nil {
```
obey-agent: Same hole on UpdateProgress(100): complete/start post infra errors return before `PropagateCompletion` (lines 176–179). A library caller completing via 100% can mark the task done while parent sequence stays open.

obey-agent (changes requested):
## Adversarial review on head `a920a6a`

Stated goal: run `task_complete` on every completion surface (`MarkComplete`, `status set completed`, `UpdateProgress(100)`), preserving historical pre order and “post fail-closed keeps the task completed.”

Manager-layer wiring for the missing surfaces is real and tested. One post-transition ordering bug breaks the historical completion contract and can leave the campaign half-updated after a successful completion mutation.

## Issue counts
- bugs: 1 (blocking)
- residuals: 2

## Findings

### [bug] Post-hook infrastructure errors skip ledger, sequence propagation, and later post stages

After the completion mutation is saved, `MarkComplete` / `UpdateProgress(100)` do:

1. `finishTaskHooks(startStage)`
2. `finishTaskHooks(completeStage)`
3. `emitLedger` (MarkComplete only)
4. `PropagateCompletion`

`finishTaskHooks` returns a hard error on runner / `SaveEvents` failure (fail-closed *blocked* posts only warn and return nil).

**Break sequence A — start post infra fails on first-work complete:**
1. Task has both `hooks.start` and bare `hooks.pre/post` completion bindings.
2. `MarkComplete` (or `UpdateProgress(100)`) runs complete pre + start pre, saves **completed** state.
3. `finishTaskHooks(start)` hits `SaveEvents` error (or runner err).
4. Function returns immediately.
5. **`task_complete` post never runs**, ledger never emits, **`PropagateCompletion` never runs**.

Historical `fest task completed` order was: complete pre → MarkComplete (start + mutation + start post + **ledger + propagate**) → **complete post**. Complete post and parent-sequence propagation always ran after a successful MarkComplete return. This PR inverts that: complete post can be skipped entirely, and sequence/phase completion can stall while the task JSONL says completed.

**Break sequence B — complete post infra fails:**
1. Task already started; only completion bindings fire.
2. Mutation saves completed.
3. `finishTaskHooks(complete)` returns infra error.
4. Ledger + `PropagateCompletion` skipped again — same half-updated tree.

The PR text says fail-closed **post** keeps the task completed (true for blocked posts) and that infra post errors fail the command with the task already completed — but it does **not** say parent propagation is optional. Leaving sequence goals incomplete after a completed task is “partial success that leaves the system broken.”

**Required:** after a successful completion Save/SyncFrontmatter:
- always run ledger + `PropagateCompletion` (best-effort as today) **before** returning for post-hook errors, **or** run them before post finishes; and
- still run remaining post stages (at least `task_complete` post) even if an earlier post stage had a non-blocking/infra failure — or document and test that start-post failure deliberately drops complete-post (that would still be a product bug vs historical UX).

Add a unit test: inject a post-stage SaveEvents/runner failure after a successful completion mutation; assert sequence propagation still attempted and/or complete post still planned to run.

## What holds

| Claim | Result |
|---|---|
| `status set completed` runs completion hooks | Holds — regression test with real `true`/`false` |
| Blocked complete pre leaves task incomplete | Holds — manager + status tests |
| UpdateProgress(100) fires complete hooks only on transition | Holds — replay guard |
| Pre order complete→start / post order start→complete | Holds — `TestMarkComplete_StartAndCompletionHookOrder` |
| Fail-closed complete post keeps task completed | Holds — returns nil after stderr warn |
| `fest task completed` blocked JSON shape | Holds for **pre** blocked path via `BlockedHookName` |

## Residuals (non-blocking)

1. **`--json` + post infrastructure error** — command returns raw `err` without the structured `{success,task,...}` document. PR claims JSON contract unchanged; only the blocked-pre path is structured. Align docs or emit `success:false` JSON for all MarkComplete failures.
2. **Quality gates still only on `fest task completed`** — acknowledged in PR; not introduced here.

## Verdict

**REQUEST_CHANGES** — surface coverage is the right design, but post-finish error handling can drop `task_complete` post, ledger, and sequence propagation after the task is already completed. Fix the post-transition side-effect ordering before merge.

lancekrogers:
Addressed the review in 0e8f8a3:

**Blocking bug (post-hook error skips ledger, propagation, and later post stages): fixed with the ordering you proposed.** `MarkComplete` and `UpdateProgress(100)` now collect post-stage errors instead of returning at the first one. After a successful completion mutation, the task_complete post stage, the ledger emit, and best-effort `PropagateCompletion` always run regardless of an earlier post-stage failure; the first post error is returned only after every side effect has been attempted. Both break sequences from the review are covered by new tests that inject a real `SaveEvents` infrastructure failure (the start post hook shadows `.fest` with a plain file, the complete post hook restores it) and assert the complete post stage still executed and the sequence goal still reached `completed`. Against the pre-fix code both tests fail on both assertions.

**Residual 1 (`--json` + post infrastructure error): fixed.** `fest task completed --json` now emits a structured `success:false` document for every `MarkComplete` failure. The blocked-pre path keeps its exact prior shape (`blocked:true`, `failed_hook`, fixed message); other failures emit `blocked:false` with the error in `message`, so agents can distinguish a blocked (not completed) outcome from a post failure (completed) without parsing stderr.

**Residual 2 (quality gates only on `fest task completed`):** pre-existing asymmetry, acknowledged in the PR body, untouched here.

Gates on 0e8f8a3: 2835 unit tests passed, 81 containerized integration tests passed, lint clean.

obey-agent (commented):
## Re-review on head `0e8f8a3` (prior REQUEST_CHANGES)

Prior blocking finding: after a successful completion Save, `finishTaskHooks` hard-errors returned **before** the later post stage, ledger, and `PropagateCompletion`, leaving a completed task with a permanently stalled sequence.

### Attack: did the new commit actually close that hole?

| Prior break | Status on this head |
|---|---|
| Start post infra/`SaveEvents` fails → complete post never runs | **Fixed** — both `finishTaskHooks` calls always run; first error kept in `postErr` |
| Post infra fails → `emitLedger` skipped (`MarkComplete`) | **Fixed** — ledger runs after both post stages |
| Post infra fails → `PropagateCompletion` skipped | **Fixed** — propagate runs after posts on both `MarkComplete` and `UpdateProgress(100)` |
| No test for post-infra after successful mutation | **Fixed** — `TestMarkComplete_PostInfraFailureStillRunsRemainingSideEffects` and `TestUpdateProgress_HundredPostInfraFailureStillPropagates` inject a real SaveEvents failure (shadow `.fest` with a file), assert error surfaces, task stays completed, complete post still exec'd, sequence goal reaches `completed` |
| `--json` only structured on blocked-pre | **Fixed** — all `MarkComplete` failures emit `{success:false,task,blocked,message}` |

Verified the new tests pass on this head (`go test ./internal/progress/ -run PostInfra`).

I could not re-open the original half-updated-tree break with the post-transition ordering on this head.

### Residuals (non-blocking)

1. **JSON post-failure shape** — `success:false` + `blocked:false` implies “mutation applied, post infra failed,” but the document does not set an explicit `status: completed` / `applied: true`. Agents that treat any non-zero exit as “task incomplete” will mis-handle until they re-query progress. Acceptable if callers already treat `blocked` as the only refuse-mutation signal.
2. **`UpdateProgress(100)` still has no `emitLedger`** — only `MarkComplete` emits campaign ledger. Pre-existing asymmetry; CLI 100% paths force `MarkComplete` / reject 100, so library-only.
3. **`PropagateCompletion` errors still discarded (`_ =`)** — if both post stages leave `.fest` unusable *and* the sequence goal has hooks that need `SaveEvents`, goal completion can still fail silently while the command returns the post error. Main path (no goal hooks / healthy store) is covered by the new tests.
4. **`MarkInProgress` still early-returns on post infra** — different surface; not claimed by this PR.

### Verdict

**COMMENT** — prior REQUEST_CHANGES root cause is actually fixed with tests that hit the break shape. No new merge-blocking correctness gap found against the stated completion-surface / post-transition contract on `0e8f8a3`.

PR #329, opened 2026-08-13
fix(create): make festival dry-run a no-write preview

lancekrogers:
## Summary

- make `fest create festival --dry-run` return before any scaffold writers run
- render the planned festival file tree for human output and structured preview JSON for agents
- share core and gate template definitions between preview and real creation

## Root cause

`--dry-run` entered the normal create pipeline, wrote the festival directory and templates, then returned early only when template markers were found. Marker-bearing previews left unregistered festival shells that consumed the apparent ID on the next create; marker-free previews could continue into registration.

## User impact

Dry runs now leave the workspace, registry, navigation state, ledger, and campaign contract untouched. A real create after a preview reuses the previewed festival ID.

## Validation

- `just test unit-raw`
- `go test -short -tags=dev ./...`
- `just lint all`
- `go vet -tags=integration ./...`
- `go vet -tags=dev ./...`
- `just build quick-stable`
- `just build quick-dev`
- `TESTCONTAINERS_RYUK_DISABLED=true go test -v -tags=integration -run 'TestCreateFestival(DryRun|AgentValidationFailure)' -timeout 5m ./tests/integration`

### DIFF
--- a/docs/cli-reference/fest-reference.md
+++ b/docs/cli-reference/fest-reference.md
@@ -962,7 +962,7 @@ fest create festival [flags]
 ```
       --agent                 Strict mode: process markers, auto-validate, rollback on blocking errors, JSON output
       --dest string           Destination under festivals/: planning or ritual (use 'fest promote' to advance to active) (default "planning")
-      --dry-run               Show template markers without creating file
+      --dry-run               Preview the festival file tree without writing anything
       --goal string           Festival goal
   -h, --help                  help for festival
       --json                  Emit JSON output
--- a/docs/cli-reference/fest_create_festival.md
+++ b/docs/cli-reference/fest_create_festival.md
@@ -11,7 +11,7 @@ fest create festival [flags]
 ```
       --agent                 Strict mode: process markers, auto-validate, rollback on blocking errors, JSON output
       --dest string           Destination under festivals/: planning or ritual (use 'fest promote' to advance to active) (default "planning")
-      --dry-run               Show template markers without creating file
+      --dry-run               Preview the festival file tree without writing anything
       --goal string           Festival goal
   -h, --help                  help for festival
       --json                  Emit JSON output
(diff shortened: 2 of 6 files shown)


obey-agent (commented):
## Adversarial review on head `3216b58`

Stated goal: `--dry-run` on `fest create festival` must be a true no-write preview (no shell dir, no registry, no ledger, no navigation), and a subsequent real create must reuse the previewed ID.

## Attack results

| Claim | Result |
|---|---|
| Dry-run returns before any scaffold writer | **Holds** — `RunCreateFestival` calls `previewCreateFestival` immediately after config + seed validation, before `scaffoldFestivalDirectory` |
| Zero-marker dry-run no longer registers | **Holds** — old path only early-returned when `markersTotal > 0`, so marker-free runs fell through to registration; that branch is gone |
| Workspace/registry untouched | **Holds in tests** — integration snapshot (`find` + sha256 of every file) equal before/after human and agent dry-run; events file absent after zero-marker preview |
| Real create reuses previewed ID | **Holds in tests** — dry-run then create yields sole `preview-only-PO0001` |
| ID generation itself is no-write | **Holds** — `GenerateID` / `FindNextCounter` only scan existing dirs |
| Shared core/gate template defs prevent preview drift | **Partial** — core list and gate phase types are shared; phase *contents* still come from a separate `ReadDir`/`WalkDir` of phase templates vs `RunCreatePhase` |

I could not re-open the “dry-run left an unregistered shell / burned an ID” failure mode on this head with the early-return + integration snapshot.

## Residuals (non-blocking)

1. **Sibling create surfaces still use the old dry-run shape.** `create phase` / `sequence` / `task` still create then optionally dump markers (`DryRun && markersTotal > 0`). Out of PR scope, but the same class of bug remains elsewhere in the CLI.
2. **Preview tree ≠ guaranteed create tree.** Auto-phases are projected by walking phase template dirs (skipping `GOAL.md`/`gates`) plus a hard-coded `PHASE_GOAL.md`, not by invoking `RunCreatePhase`. Extra/missing files from phase create logic will not show in the preview. Acceptable for “planned scaffold” if docs do not claim byte-identical output.
3. **No unit test that a failed seed validation path still writes nothing** beyond the mutual-exclusion check — seed failure returns before dry-run; OK.
4. **CodeQL green only** in status rollup — integration proof is in-repo tests, not this Actions run.

## Verdict

**COMMENT** — festival dry-run is actually no-write with tests that hit the prior marker-bearing and marker-free failure shapes. No merge-blocking gap against the stated create-festival dry-run contract on `3216b58`.

PR #330, opened 2026-08-17
fix(errors): tell the operator what actually failed when a remote is unreachable

lancekrogers:
## The bug

`fest init` on a device that could not reach GitHub:

```
[root@kindle kindle-camp]# fest init
Template cache not found, syncing from GitHub...
Error: auto-sync failed: resolving latest dev tag: listing remote tags: I/O operation failed: exit status 128
Hint: Check file/directory permissions and try again
```

Everything after `auto-sync failed` is unhelpful or actively wrong. Reproduced locally by pointing the configured repository at an unreachable host, which produces those two lines verbatim. Three separate defects stack into it.

### 1. git's stderr was discarded

`exec.Cmd.Output` populates `ExitError.Stderr`, but the call sites kept only the exit code. `128` is git's generic fatal: it does not distinguish "no route to host" from "cannot verify certificate" from "access denied". git had already written the answer and camp threw it away.

### 2. Remote failures were reported as I/O failures

There was no network category, so `git ls-remote` failures used `errors.IO`, whose hint is `HintCheckPermissions`. A machine with no route to GitHub was told to check file permissions, which is the single place the problem is not.

### 3. `WithField("hint", ...)` renders nothing

`Error()` renders `e.Hint`. A hint stored as a *field* is invisible in text output and surfaces only under `--json`, in a different key. Verified rather than assumed:

```
WithField("hint") renders as:
    op: I/O operation failed
    Hint: Check file/directory permissions and try again
WithHint renders as:
    op: I/O operation failed
    Hint: AUTHOR_INTENDED_HINT
RESULT: WithField hint is INVISIBLE in text output
```

**49 call sites** across 21 files had an author-written hint that no user has ever seen, while the constructor's generic hint took its place. `init` was one of them: it already had the right guidance, `"check your internet connection or run 'fest sync' manually"`, and it was being dropped in favor of the permissions advice.

## After

```
Error: auto-sync failed: resolving latest dev tag: listing remote tags: could not reach the remote: unable to access 'https://invalid.invalid/nope.git/': Could not resolve host: invalid.invalid: exit status 128
Hint: check your internet connection or run 'fest sync' manually
```

## The change

- `gitStderr` recovers git's own message from `ExitError.Stderr`, strips the `fatal:` ceremony, collapses it to one line, and caps the length so a pathological remote cannot flood a CLI error.
- `errors.Network(op, err, detail)` carries the remote's message and a reachability hint. Both `ls-remote` call sites use it.
- All 49 `WithField("hint", ...)` become `WithHint(...)`, so authored guidance reaches the terminal.

Making those hints visible exposed that a nested error printed **one `Hint:` line per layer**. `Error()` now renders exactly one: the outermost, being closest to what the operator typed, falling back inward so a bare `Wrap()` does not discard the inner guidance.

## A bug this PR introduced and the suite caught

My first version of that renderer recursed into nested `*Error` values using `errors.As`. That looked equivalent and was not: `errors.As` searches the **whole** chain, so a plain `fmt.Errorf` sitting between two `*Error` layers was skipped and its message silently deleted. `approval readiness failed: invalid evidence path "../outside.md": escapes the phase directory` became `approval readiness failed: escapes the phase directory`, losing the sentence that names what the operator got wrong.

`TestCheckApprovalReadiness_InvalidEvidencePathsFailClosed` caught it. The renderer now strips rendered `Hint:` lines from the cause text instead, which cannot drop a layer whatever the chain shape, and that case has its own regression test.

## Tests

- `gitStderr`: recovers the real message from a live failed `ls-remote` (skips if the bogus host resolves), strips `fatal:`, joins multiple lines, stays bounded, and ignores non-`ExitError` failures such as a missing git or a cancelled context.
- `errors.Network`: carries the cause and the network hint, and specifically must **not** carry the permissions hint.
- Nested errors render exactly one hint; a bare `Wrap` keeps the inner one; `errors.Is` still reaches the root cause.
- An intervening `fmt.Errorf` keeps its message (the regression above).
- A tree walk fails on any new `WithField("hint", ...)`, so this cannot silently come back.

## Gates

`go vet ./...`, `golangci-lint run ./...` (0 issues), `just test unit`: **2846/2846 passed**.

## Not addressed here

`fest init` still cannot work at all on a machine with no network, since the templates only exist after a sync. That is a design question (bundled or embedded fallback templates) rather than an error-reporting one, so it is deliberately out of scope. This PR's job is that the operator now learns the real reason and the actionable next step instead of being sent to check file permissions. Happy to open an intent for the offline path if you want it pursued.

### DIFF
--- a/internal/commands/feedback/feedback.go
+++ b/internal/commands/feedback/feedback.go
@@ -77,7 +77,7 @@ func runInit(ctx context.Context, criteria []string, force bool) error {
 	festivalPath, err := shared.ResolveFestivalPath(cwd, "")
 	if err != nil {
 		return errors.Wrap(err, "detecting festival").
-			WithField("hint", "run from within a festival directory")
+			WithHint("run from within a festival directory")
 	}
 
 	store := feedback.NewStore(festivalPath)
@@ -162,7 +162,7 @@ func runCriteriaAdd(ctx context.Context, criteria []string) error {
 	festivalPath, err := shared.ResolveFestivalPath(cwd, "")
 	if err != nil {
 		return errors.Wrap(err, "detecting festival").
-			WithField("hint", "run from within a festival directory")
+			WithHint("run from within a festival directory")
 	}
 
 	store := feedback.NewStore(festivalPath)
@@ -230,7 +230,7 @@ func runAdd(ctx context.Context, criteria, observation, jsonInput, task, severit
 	festivalPath, err := shared.ResolveFestivalPath(cwd, "")
 	if err != nil {
 		return errors.Wrap(err, "detecting festival").
-			WithField("hint", "run from within a festival directory")
+			WithHint("run from within a festival directory")
 	}
 
 	store := feedback.NewStore(festivalPath)
@@ -247,7 +247,7 @@ func runAdd(ctx context.Context, criteria, observation, jsonInput, task, severit
 		// Use flags
 		if criteria == "" || observation == "" {
 			return errors.Validation("criteria and observation are required").
-				WithField("hint", "use --criteria and --observation, or --json")
+				WithHint("use --criteria and --observation, or --json")
 		}
 		obs = &feedback.Observation{
 			Criteria:    criteria,
@@ -311,7 +311,7 @@ func runView(ctx context.Context, criteria, severity string, jsonOutput, summary
 	festivalPath, err := shared.ResolveFestivalPath(cwd, "")
 	if err != nil {
 		return errors.Wrap(err, "detecting festival").
-			WithField("hint", "run from within a festival directory")
+			WithHint("run from within a festival directory")
 	}
 
 	store := feedback.NewStore(festivalPath)
@@ -399,7 +399,7 @@ func runExport(ctx context.Context, format string) error {
 	festivalPath, err := shared.ResolveFestivalPath(cwd, "")
 	if err != nil {
 		return errors.Wrap(err, "detecting festival").
-			WithField("hint", "run from within a festival directory")
+			WithHint("run from within a festival directory")
 	}
 
 	store := feedback.NewStore(festivalPath)
--- a/internal/commands/list/list.go
+++ b/internal/commands/list/list.go
@@ -167,7 +167,7 @@ or lifecycle status changes (similar to fest watch, but without cycling). Ctrl+C
 			}
 			if opts.watch && opts.json {
 				return errors.Validation("--watch cannot be combined with --json").
-					WithField("hint", "use one-shot fest list --json for agents; --watch is human terminal output")
+					WithHint("use one-shot fest list --json for agents; --watch is human terminal output")
 			}
 			if opts.watch && !listStdoutIsTerminal() {
 				return errors.Validation("--watch requires an interactive terminal")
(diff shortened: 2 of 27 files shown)


obey-agent (approved):
## Verdict: APPROVE
**Head:** `1e1e2bc` · **PR goal:** surface git’s real remote failure and authored hints when a remote is unreachable (stop “exit 128” + permissions red herring)

_(none)_

### Non-blocking residuals
1. **[stripHintLines is textual]** Drops cause lines prefixed `Hint: `; a pathological cause body that embeds that exact prefix could lose a non-hint line. Unlikely in practice; better than the errors.As chain-skip bug this replaces.
2. **[offline init still impossible]** Templates still require network/sync — correctly out of scope; only the diagnosis improved.

### What I checked
- `gitStderr` from `ExitError.Stderr`, `errors.Network`, `ls-remote` call sites, `WithField("hint")` → `WithHint`, single-hint `Error()` + intervening `fmt.Errorf` regression test
- Tree-walk test against new `WithField("hint"`)
- Local: `go test ./internal/errors ./internal/github` pass

### What I did not check
- Full `just test unit` 2846 count; every CLI path’s human wording under a real offline Kindle

lancekrogers:
## Update: the Kindle has working network, so the hint was still wrong

`exit 128` with a live network is usually TLS, not DNS. A machine with no CA bundle is the normal state on minimal and embedded systems, and the original PR would have told that operator "check your internet connection" for a network that is already fine.

`networkHint` now classifies from git's own message:

| git says | hint |
| --- | --- |
| certificate / SSL / CAfile / self-signed | install CA certificates, or set `GIT_SSL_CAINFO` |
| could not resolve host / name resolution | check DNS and connection |
| authentication failed / permission denied / terminal prompts disabled | check credentials |
| anything else | generic reachability |

Verified against a real bad-certificate host rather than a mocked string:

```
before: Hint: check your internet connection or run 'fest sync' manually
after:  Hint: The remote was reached but its certificate could not be verified.
        Install CA certificates (e.g. the ca-certificates package) or set
        GIT_SSL_CAINFO to a CA bundle
```

### This disproved the hint-precedence rule in the first commit

That commit made the **outermost** hint win, reasoning that the layer closest to what the operator typed carries the most context. The TLS case shows it backwards: `init`'s outer hint could only guess "check your internet connection", while the inner layer knew it was a certificate.

Precedence is now **innermost wins**, falling back outward when inner layers carry no hint. Outer context is not lost, since the message chain still carries it. The test that encoded the old rule is corrected rather than deleted, and the reason for the new one has its own test.

Gates re-run: `go vet ./...`, `golangci-lint run ./...` (0 issues), `just test unit` **2848/2848**.

obey-agent (approved):
## Verdict: APPROVE
**Head:** `3889f97` · **PR goal:** surface git’s real remote failure and actionable hints when a remote is unreachable

_(none)_

### What changed since prior APPROVE (`1e1e2bc`)
1. **[networkHint]** Classifies TLS / DNS / auth from git detail instead of one generic “check your connection” (Kindle/minimal CA case is the motivating example).
2. **[resolveHint]** Innermost hint wins over outer generic advice; tests cover cert-vs-connection and nested single-hint rendering.

### Non-blocking residuals
1. **[substring classification]** `networkHint` is keyword-based (`ssl`/`403`/…); odd remotes can mis-bin. Still far better than permissions.
2. **[stripHintLines textual]** unchanged residual from prior review.
3. **[offline init]** still needs network/sync by design.

### What I checked
- Delta `1e1e2bc` → `3889f97` only
- `go test ./internal/errors ./internal/github` pass on this head

### What I did not check
- Live Kindle TLS-without-CA and no-DNS end-to-end; full unit suite count

PR #331, opened 2026-08-17
shell: add POSIX sh integration so fgo and fls work outside bash/zsh

lancekrogers:
## The bug

`fest shell-init` emitted bash/zsh and fish, and nothing else. On a rooted Kindle (busybox ash, login shell `-sh`) that means no `fgo`, no `fls`, and no `fest` wrapper at all:

```
### shell is: /bin/sh (busybox ash)

### BEFORE: eval bash script
sh: eval: line 213: syntax error: unexpected "(" (expecting "}")
-> fls defined? NO
```

It dies on the first `COMPREPLY=($(compgen ...))`. A shell that cannot parse an array assignment aborts the entire `eval`, so a script that is 95% portable delivers 0% of its functions. This is the same failure class as camp #570, in the other half of the user's command surface: `csw`/`cgo` come from camp, `fls`/`fgo` come from fest, and fixing camp alone left half the Kindle broken.

## After

```
### AFTER: eval sh script
-> fls defined? yes
-> fgo defined? yes

### fls actually runs:
List festivals filtered by status.
Works from anywhere - finds the festivals workspace automatically.
```

Navigation, including the two-argument status form that used to rely on a bash `=~` match:

```
### cwd before: /root/ws
### cwd after fgo active: /root/ws/festivals/active
### two-arg status form (fgo active demo-fest):
### cwd: /root/ws/festivals/active/demo-fest
### error path is clean (no linked project):
fgo: no project linked (use 'fest link <path>' from a festival)
### exit code: 1
```

## The change

`bash_zsh.sh` is split by what a shell must be able to do, not by shell name:

- **`posix_core.sh`** — the guard and the `fgo` / `fls` / `fest` wrappers, valid POSIX.
- **`bash_zsh_completions.sh`** — `COMPREPLY`, `compgen`, `complete`, `compdef`, `compadd`, process substitution.

bash and zsh compose the core plus completions; `sh` gets the core alone. **The wrappers now have one definition instead of two**, so an improvement to `fgo` cannot land in one shell and be forgotten in the other. POSIX sh gives up exactly one thing — tab completion — because there is no programmable completion to hook.

Syntax converted in the shared wrappers: `[[ ]]` → `[ ]`, the `=~` status match → a `case` statement (POSIX, and no subprocess), and `local exit_code=$?` → a separate assignment so the exit code is not masked by `local`'s own status.

Added `detect` / `InitCommand` / `InitHint` so no surface hardcodes a shell in a hint, and `ValidArgs` on `shell-init` so `fest shell-init <TAB>` follows `SupportedShells` instead of a list that goes stale.

### ksh93 is deliberately excluded

ksh93 **parses** this script fine, so a parse check would call it supported. At runtime it is worse than a syntax error: ksh93 has no `local`, does not abort when it fails, and simply skips the assignment — so every wrapper reads whatever the caller happened to have in that variable name, and `fgo` cds somewhere unrelated with no error.

The script probes for the **effect** of `local`, not merely that the command ran, and refuses:

```
### ksh (no local builtin) must refuse cleanly and define nothing:
fest: this shell has no 'local' builtin, which fest's sh integration requires.
fest: known affected: ksh93. Use bash, zsh, dash, busybox ash, or mksh.
-> fls not defined (correct)
```

## A pre-existing drift this surfaced

Writing the parity test found that `fls` tab completion **already** disagreed across shells: fish offered `active planning completed dungeon` where bash offered `active ready planning ritual completed dungeon all` plus flags, and fish described `--all` as "Include empty status categories" when the flag is "include completed and dungeon festivals". Both corrected, and the test now fails if they diverge again.

## Tests

- A static scanner rejects 14 bash-only constructs in the POSIX core, and a real `dash`/`busybox` parses it. Both matter: the scanner only knows what it was taught, and the parse only catches what the sample exercises.
- bash and zsh must **embed the POSIX core verbatim**, which is what makes the two-copies-drift failure impossible rather than merely unlikely.
- bash and zsh must still register their completions — sharing the core has to cost them nothing.
- The `local` probe must assert `local`'s effect and must run before any wrapper is defined.
- `fls` completion vocabulary must match across bash and fish, comparing whole tokens rather than substrings (`all` is inside `--all`, `ready` inside `already`).
- `detect` mapping, including the `-sh` login form seen on the Kindle, and that every value it returns is one `Generate` accepts.

Both new guards were **verified to fail**, not just observed passing — injecting `local -a items=("a" "b")` and `[[ -n "$1" ]]` into the core trips the scanner at both lines and gets `/bin/dash: 192: Syntax error: "(" unexpected`.

## Verified in containers, with the real binary

Cross-compiled `fest` run under: **busybox ash** (loads, navigates, error paths clean), **dash** (loads), **mksh** (loads), **ksh93** (refuses with the explanation above), **bash** (`complete -p` still shows `_fls_completions` and `_fgo_completions`).

## Gates

`go vet ./...` clean, `golangci-lint run ./...` 0 issues, `just test unit` **2846/2846 passed**.

## Note on the docs diff

`just docs` also wanted to add `docs/cli-reference/fest_tokens.md` and its index entry — pre-existing drift from `fest tokens` landing without a docs regen, unrelated to this change. Left out of this PR so the diff stays reviewable; worth its own regen commit.


obey-agent (approved):
## Verdict: APPROVE
**Head:** `7152d6a` · **PR goal:** Make `fest shell-init` work on POSIX sh (busybox ash/dash/mksh) by splitting wrappers from bash/zsh completion so eval no longer dies on `COMPREPLY=($(compgen...))`.

none

### Non-blocking residuals
1. **[hints]** `InitHint` / `DetectFromEnv` are added and tested but not adopted at existing operator surfaces. `internal/commands/navigation/go.go` still hardcodes `eval "$(fest shell-init zsh)"`. Primary fix (Generate/sh script) does not depend on this; follow-up wiring would close the claim fully.
2. **[ci/env]** `TestShScriptParsesUnderRealPosixShell` skips when dash/busybox/ash are absent; GitHub checks on this head only show CodeQL. Static scanner + local `go test ./internal/shell` pass; real ash/dash verification is claimed in the PR body (containers), not re-proven in CI here.
3. **[pre-existing]** `fest()` go-wrapper path still suppresses stderr and does not propagate `fest go` exit codes on failure (same as main). `fgo` / `_fest_cd_to` path is correct.

### What I checked
- Full diff: `posix_core.sh`, completions split, `shell.go`/`scripts.go`, `detect.go`, fish status vocabulary, docs/shell-init help, ValidArgs from SupportedShells
- Claim vs code: bash-only constructs removed from shared core; `local` effect probe before wrappers; `[[`/`=~` → `[`/`case`; `local x=$?` → separate assignment
- Tests: `go test ./internal/shell` PASS (posix parse test skipped on this host — no dash/busybox)
- `TestEveryBourneShellSharesTheSameWrappers` / completion keep tests — single-definition drift guard is real
- Prior CR: none (first review on this head)
- CI: CodeQL green

### What I did not check
- Live busybox ash / Kindle session, ksh93 refusal under a real ksh, full `just test unit` 2846 suite, fish interactive completion UX

PR #332, opened 2026-08-17
docs(cli): regenerate reference for fest tokens and gate against staleness

lancekrogers:
## The drift

`fest tokens` shipped without a docs regen, so `docs/cli-reference` has been missing the command page and its index entry. Surfaced while regenerating docs for #331, and deliberately left out of that PR so its diff stayed reviewable.

The regen is **purely additive** (116 insertions, 0 deletions), so nothing stale was overwritten — this was a missing page, not a wrong one.

## Why this needs more than a regen

Regenerating fixes today's drift and nothing else. The same gap reopens the next time a command lands, and it is invisible until someone happens to run `just docs` for an unrelated reason — which is exactly how this one was found.

Added a `docs-check` recipe, wired into `just gate`:

```
=== gate: docs ===
docs/cli-reference is up to date
```

It generates into a **temp dir rather than the working tree**, so running the gate never leaves modified files behind — a check that dirties the tree it is checking would be its own problem.

## Verified in both directions

A guard that only ever passes is decoration, so I confirmed it fails on the case it exists for. With the docs reverted to their stale committed state:

```
error: recipe `docs-check` failed with exit code 1
```

and it prints the offending diff so the fix is obvious:

```
+  fest tokens                    # Current festival's planning corpus
+  fest tokens --all              # Entire festivals workspace
```

I also confirmed `gendocs` output is **deterministic** across runs before wiring it into a gate — a check that compares generated output is only safe if that output is stable, otherwise it fails spuriously and gets disabled.

## Gates

`just docs-check` passes. The change is generated docs plus a justfile recipe; no Go code is touched.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -64,6 +64,8 @@ gate:
     go vet -tags=integration ./...
     echo "=== gate: lint ==="
     just lint all
+    echo "=== gate: docs ==="
+    just docs-check
     echo "=== gate: stable tests ==="
     just test all
     echo "=== gate: dev unit tests ==="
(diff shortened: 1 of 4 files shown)


obey-agent (approved):
## Verdict: APPROVE
**Head:** `6f1d040` · **PR goal:** Add the missing `fest tokens` CLI reference pages and gate `docs/cli-reference` against future regen drift.

none

### Non-blocking residuals
1. **[gate cost]** `docs-check` runs `just build quick` even when `gate` already built; pure wall-time, not correctness.
2. **[scope]** Gate is a full-tree `diff -ru` of `docs/cli-reference` vs gendocs output — correct only while that directory is fully generated. Holds today (additive tokens pages only).

### What I checked
- Diff is additive docs (fest_tokens.md, index link, fest-reference section) + Justfile `docs-check` wired into `gate`
- Check generates into a temp dir and traps cleanup — will not dirty the working tree
- No Go code changes; CodeQL green
- Aligns with the #331 note that tokens docs were intentionally deferred

### What I did not check
- Live `just docs-check` binary run on this machine, gendocs determinism across OS/locales

PR #333, opened 2026-08-17
feat(init): bundle the methodology so fest init works without a network

lancekrogers:
## The bug

The methodology scaffold only ever existed after a successful sync, so a machine that could not reach GitHub could not run `fest init` **at all**. From the Kindle report:

```
[root@kindle kindle-camp]# fest init
Template cache not found, syncing from GitHub...
Error: auto-sync failed: ...
```

#330 made that failure legible. It did not make the command work — the templates still only arrive over the network, so an offline machine gets a clear explanation of why it cannot start.

## After

Verified in a container with **`--network none`**:

```
### DEFAULT REPO, no network, clean HOME:
Template cache not found, syncing from GitHub...
⚠ Could not reach the methodology repository: resolving latest dev tag: git command not found
Using the methodology bundled with this fest build.
Run 'fest sync' once the network is available to pick up newer templates.
Initializing festival structure at festivals...
-> workspace: README.md active dungeon planning ready ritual
```

A complete, usable workspace — `fest list` runs against it, and `.festival/` carries the templates, extensions, and agent guides.

## The change

`methodology/festivals` (440K, 60 files) is embedded in the binary and seeded when the sync fails. Sync becomes the way to **update** the methodology rather than the way to obtain it.

### The bundle is not a universal substitute

It is a copy of the *default* repository, so it stands in only for the default source. An operator who pointed fest at their own methodology keeps failing:

```
### CUSTOM REPO, no network, its own clean HOME:
Template cache not found, syncing from GitHub...
Error: auto-sync failed: resolving latest dev tag: git command not found
-> workspace created? 0 entries
```

Seeding ours there would hand them a methodology they never chose, and `init` would report success — the swap would be easy to miss. Failing is the honest outcome: fest cannot produce templates it has never seen.

## A build-hygiene problem this uncovered

`//go:embed` reads the **working tree, not the commit**, so anything sitting in `methodology/festivals` at build time is baked into the binary — including untracked and gitignored files. That makes the shipped scaffold depend on whose machine ran the build.

Not hypothetical. The repo gitignores `CLAUDE.md`, and this checkout has **eight generated ones** under the methodology tree:

```
$ head -1 methodology/festivals/.festival/CLAUDE.md
<claude-mem-context>
```

Those are local session logs. A release built on that machine would have embedded them and written them into **every user's** workspace on `fest init`; a clean-checkout build would not. `TestEmbeddedScaffoldMatchesGitTrackedFiles` compares the embedded set against `git ls-files`, so the binary can only carry what the commit contains, and it names the offending file when it fails.

> Worth knowing separately: those eight files are sitting in the main `projects/fest` checkout right now. They are gitignored and regenerate, so they are harmless until something embeds that tree — which this PR does. The test catches it; they should be deleted before cutting a release from that checkout.

## Tests

- The bundled scaffold matches the on-disk methodology **file for file**, so adding a methodology file cannot quietly fail to ship.
- `.festival` and the status directories are present by name. This is the specific casualty of a missing `all:` prefix — `go:embed` silently skips dot-prefixed entries without it, which would produce a workspace that looks complete and has no templates. **Verified by removing the prefix: 1 file embedded instead of 60.**
- Embedded contents match byte for byte after seeding.
- Context cancellation is honored; an empty target is rejected.
- The bundle-applies rule is table-driven across unconfigured, explicit-default, pinned-channel, someone-else's-repo, and default-repo-at-another-path.
- The git-parity guard was **verified to fail** by planting a stray gitignored file.

## Two adjacent fixes

- `--from` pointing at a missing directory ran a sync that writes to the config dir — somewhere else entirely — and then failed on the original path anyway. It now says what is missing.
- `errors.Message` renders an error without its `Hint:` line. Splicing a cause into a warning was emitting a next-step for something the command was already handling.

## Gates

`go vet` clean, `golangci-lint run ./...` **0 issues**, `just test unit` **2870/2870 passed**.

## Follow-up I am not doing here

While verifying the custom-repo path I found a **pre-existing** defect from #330: `errors.Validation` attaches the generic `HintSeeHelp` *in the constructor*, so innermost-wins hint precedence surfaces the most generic hint in a chain rather than the most specific. The refusal above prints `Hint: Run 'fest help' for more information` instead of the connection advice this code attaches. That is the "check file permissions" problem in a new form, it predates this branch, and it is an errors-package semantics change that deserves its own review. Separate PR.

### DIFF
--- a/internal/bundled/bundled.go
+++ b/internal/bundled/bundled.go
@@ -1,0 +1,69 @@
+// Package bundled materializes the methodology scaffold embedded in the fest
+// binary onto disk, for machines that cannot reach GitHub.
+package bundled
+
+import (
+	"context"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/Obedience-Corp/fest/internal/errors"
+	"github.com/Obedience-Corp/fest/methodology"
+)
+
+const (
+	dirPerm  = 0o755
+	filePerm = 0o644
+)
+
+// Seed writes the embedded methodology scaffold into targetDir, creating it if
+// needed. Existing files are overwritten, so a caller that wants to preserve
+// local edits must check before calling.
+//
+// Embedded files carry no mode bits, so directories are created 0755 and files
+// written 0644 — the scaffold is documents and templates, nothing executable.
+func Seed(ctx context.Context, targetDir string) error {
+	if err := ctx.Err(); err != nil {
+		return errors.Wrap(err, "seeding bundled methodology")
+	}
+	if targetDir == "" {
+		return errors.Validation("seeding bundled methodology: no target directory")
+	}
+
+	return fs.WalkDir(methodology.FS, methodology.Root, func(path string, d fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
+			return walkErr
+		}
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		rel, err := filepath.Rel(methodology.Root, path)
+		if err != nil {
+			return errors.Wrap(err, "resolving bundled path").WithField("path", path)
+		}
+		dest := filepath.Join(targetDir, rel)
+
+		if d.IsDir() {
+			if err := os.MkdirAll(dest, dirPerm); err != nil {
+				return errors.IO("creating directory", err).WithField("path", dest)
+			}
+			return nil
+		}
+
+		data, err := methodology.FS.ReadFile(path)
+		if err != nil {
+			return errors.IO("reading bundled file", err).WithField("path", path)
+		}
+		// A file can precede its parent only if the walk is reordered, but
+		// creating the parent here costs nothing and removes the assumption.
+		if err := os.MkdirAll(filepath.Dir(dest), dirPerm); err != nil {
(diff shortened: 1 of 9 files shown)


obey-agent (approved):
## Verdict: APPROVE
**Head:** `87b93b7` · **PR goal:** Embed the methodology scaffold (`all:festivals`) so offline `fest init` can seed when GitHub sync fails, only for the default repository.

none

### Non-blocking residuals
1. **[merge]** Conflicts likely with #334 on `internal/errors/errors.go` — both additive.
2. **[ops]** PR notes gitignored files under `methodology/festivals` in a dirty checkout would embed; `TestEmbeddedScaffoldMatchesGitTrackedFiles` catches that — keep release builds clean.

### What I checked
- `go:embed all:festivals` + Seed walk; `.festival` required by test
- `bundledMethodologyApplies` refuses foreign URL / non-default Path
- `--from` missing path fails before unrelated sync
- `errors.Message` strips Hint for recovered warnings
- Local: `go test ./internal/bundled/ ./internal/commands/system/ ./internal/errors/` PASS
- CodeQL green

### What I did not check
- Full `just test unit` 2870 suite, binary size impact of embed, live offline Kindle/airgap init

PR #334, opened 2026-08-17
fix(errors): let a specific hint outrank a generic one from any depth

lancekrogers:
## The bug

Innermost-wins hint precedence inverted its own purpose.

Several constructors attach a catch-all hint from the **error class alone**, with no knowledge of what actually failed:

```go
func Validation(message string) *Error {
	return &Error{..., Hint: HintSeeHelp}   // "Run 'fest help' for more information"
}
```

So a chain whose innermost layer happened to be a `Validation` surfaced that generic advice over the hint the call site wrote for the real operation. Innermost-wins was meant to prefer the layer that *knows what failed*; in practice it preferred whichever layer was deepest, and the deepest layer is often the one with the least specific advice.

Real case, `fest init` against an unreachable remote:

```
before: Hint: Run 'fest help' for more information
after:  Hint: check your internet connection or run 'fest sync' manually
```

The call site attached the second one. The first is what users saw. That is the same dead end as being told to check file permissions when a remote is unreachable — the exact failure that making hints visible was supposed to remove.

## The fix

Precedence is now **specificity first, depth second**.

A hint is *specific* when it says something about this failure:
- a caller set it with `WithHint` / `WithHintf`, or
- a constructor derived it from the failure itself — `Network` classifies git's stderr, `NotFound` reads the resource.

A hint is *class-level* when a constructor attached it from the error code alone (`Validation`, `IO`, `Config`, `Template`, `Parse`). Those are marked at construction and used only when no layer in the chain offers something better — still innermost-first among themselves, so nothing regresses for chains that only have class hints.

The marker is an unexported field, so it cannot be set from outside the package: `WithHint` is how a caller says "this one is mine", and it clears the flag even on an error that started with a class hint.

### This deliberately preserves the TLS case

The rule that motivated innermost-wins still holds, and has its own test:

```go
inner := Network("listing remote tags", err, "server certificate verification failed. CAfile: none")
outer := Wrap(inner, "auto-sync failed").WithHint("check your internet connection ...")
// → the CA-certificate hint, not the network one
```

Both hints are specific there, so depth breaks the tie and the inner one wins — the inner layer knows it is a certificate problem, the outer can only guess at the network.

## Verified by breaking it

The tests were confirmed to fail against the unfixed code, reproducing the exact symptom:

```
--- FAIL: TestSpecificHintBeatsAnInnerClassHint
    hint = "Run 'fest help' for more information",
      want "check your internet connection or run 'fest sync' manually"
    the class-level hint still surfaced:
      auto-sync failed: resolving latest dev tag: git command not found
      Hint: Run 'fest help' for more information
```

And end to end, offline in a container:

```
Error: auto-sync failed: resolving latest dev tag: git command not found
Hint: check your internet connection or run 'fest sync' manually
```

## Tests

Specific-beats-inner-class (the defect), derived-inner-still-beats-outer (the TLS rule), class hints used when nothing is better, `WithHint` overriding a class hint on its own error, inner-specific over outer-class, no hint anywhere rendering no `Hint:` line, and the message chain surviving whichever hint wins.

## Also

Corrects the `Error()` doc comment, which still described the **outermost-wins** rule that innermost-wins had already replaced — it had been stale since that change, and directly contradicted the code beneath it.

## Gates

`go vet` clean, `golangci-lint run ./...` **0 issues**, `just test unit` **2867/2867 passed**. Notably no existing test depended on the old ordering, which is some evidence the previous behavior was never deliberate.

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -56,16 +56,21 @@ type Error struct {
 	Err     error                  `json:"-"`
 	Fields  map[string]interface{} `json:"fields,omitempty"`
 	Hint    string                 `json:"hint,omitempty"` // actionable suggestion
+
+	// hintFromClass marks a Hint that a constructor attached from the error
+	// class alone, with no knowledge of what actually failed. Those lose to any
+	// hint chosen for the specific failure, wherever it sits in the chain.
+	// Unexported so it cannot be set from outside: WithHint is how a caller
+	// says "this one is mine".
+	hintFromClass bool
 }
 
 // Error returns the error string with context, followed by exactly one hint.
 //
-// The hint is the outermost one in the chain, because that is the layer closest
-// to what the operator actually typed: "run 'fest sync' manually" beats the
-// generic advice attached wherever the failure originated. Inner hints are
-// deliberately not printed. Rendering a cause with %v would otherwise splice
-// its "Hint:" line into the middle of this one's message, so a two-layer error
-// printed two hints and a three-layer error printed three.
+// Which hint that is, is resolveHint's decision; see it for the precedence and
+// why. Inner hints are deliberately not printed: rendering a cause with %v
+// would splice its "Hint:" line into the middle of this one's message, so a
+// two-layer error printed two hints and a three-layer error printed three.
 func (e *Error) Error() string {
 	msg := e.chainMessage()
 	if hint := e.resolveHint(); hint != "" {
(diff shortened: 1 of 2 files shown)


obey-agent (approved):
## Verdict: APPROVE
**Head:** `ab513b5` · **PR goal:** Specific/WithHint hints outrank constructor class catch-alls (`HintSeeHelp` et al.) at any chain depth.

none

### Non-blocking residuals
1. **[merge]** Also open: fest#333 touches `internal/errors/errors.go` (adds `Message`). Land one before the other to avoid a trivial conflict.

### What I checked
- `hintFromClass` on Validation/IO/Config/Template/Parse; Network/derived stay specific (default false)
- `specificHint` then `fallbackHint` precedence; `WithHint` clears the class flag
- Tests: specific-beats-inner-class, derived-TLS-still-innermost, class-as-fallback — all PASS locally
- CodeQL green

### What I did not check
- Exhaustive audit of every call site that relied on innermost class hints accidentally

